PR #35, opened 2022-05-26
Stack overflow error in AssembleModelGenerator.kt for arrays

### LINKED ISSUE #33: StackOverflowError in AssembleModelGenerator
**Description**

We got `StackOverflowError` in org.utbot.framework.assemble.AssembleModelGenerator#assembleArrayModel during the construction of a const model for an array that contains itself as an element.

**To Reproduce**

1. Create a new project in IntelliJ Idea
2. Use plugin to generate tests
3. Generate tests for method that have an array with itself as an element (as a static value or parameter)

**Expected behavior**

Correctly generated test with a described array as parameter or static

**Actual behavior**

Fail because of StackOverflowError in AssembleModelGenerator

**Example**
Code snippet to reproduce:

~~~kotlin
public Object[] arrayWithItselfAnAsElement(Object[] array) {
    UtMock.assume(array != null && array.length > 0);

    if (array[0] == array) {
        return array;
    }

    return null;
}
~~~

Output (shorten):
~~~
java.lang.StackOverflowError: null
	at kotlin.collections.CollectionsKt___CollectionsKt.toMutableList(_Collections.kt:1303) ~[utbot-instrumentation-1.0-SNAPSHOT.jar:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleModel(AssembleModelGenerator.kt:172) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleArrayModel(AssembleModelGenerator.kt:201) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleModel(AssembleModelGenerator.kt:182) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleArrayModel(AssembleModelGenerator.kt:203) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleModel(AssembleModelGenerator.kt:182) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleArrayModel(AssembleModelGenerator.kt:203) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleModel(AssembleModelGenerator.kt:182) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleArrayModel(AssembleModelGenerator.kt:203) ~[main/:?]
	at org.utbot.framework.assemble.AssembleModelGenerator.assembleModel(AssembleModelGenerator.kt:182) ~[main/:?]
~~~

**Environment**

Test framework: no matter

Java: 1.8

Mockito: no mocks

 - [] Other packages
 - [] Other classes
 - [] Static mocking

Test:

- [ ] Parametrized

CaelmBleidd:
Closes #33.

The problem was inside `assembleArrayModel` function. We used to check whether we have already instantiated a model, then, if we don't, we assembled its const model and stores, and, finally, put assembled model in the memory. Because of the processing order, when we had a source model that has itself as an element, we got a stack overflow error. 

To avoid it, I changed the assembling model. At first, we create a new assemble model with empty stores and the const model taken from the source model. Then we put it into the cache, and only after that we transform the const model and add transformed stores to it.

### DIFF
--- a/utbot-framework/src/test/kotlin/org/utbot/examples/arrays/ArrayOfArraysTest.kt
+++ b/utbot-framework/src/test/kotlin/org/utbot/examples/arrays/ArrayOfArraysTest.kt
@@ -266,4 +266,14 @@ internal class ArrayOfArraysTest : AbstractTestCaseGeneratorTest(testClass = Arr
             { valueBefore, valueAfter -> valueAfter.withIndex().all { it.value == valueBefore[it.index] + it.index } }
         )
     }
+
+    @Test
+    fun testArrayWithItselfAnAsElement() {
+        check(
+            ArrayOfArrays::arrayWithItselfAnAsElement,
+            eq(2),
+            coverage = atLeast(percents = 94)
+            // because of the assumption
+        )
+    }
 }
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -420,7 +420,7 @@ data class UtArrayModel(
     override val id: Int,
     override val classId: ClassId,
     val length: Int = 0,
-    val constModel: UtModel,
+    var constModel: UtModel,
     val stores: MutableMap<Int, UtModel>
 ) : UtReferenceModel(id, classId) {
     override fun toString() = withToStringThreadLocalReentrancyGuard {
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/assemble/AssembleModelGenerator.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/assemble/AssembleModelGenerator.kt
@@ -195,19 +195,24 @@ class AssembleModelGenerator(private val methodUnderTest: UtMethod<*>) {
     /**
      * Assembles internal structure of [UtArrayModel].
      */
-    private fun assembleArrayModel(arrayModel: UtArrayModel): UtModel {
-        instantiatedModels[arrayModel]?.let { return it }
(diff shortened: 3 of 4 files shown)


utbot-framework/src/test/kotlin/org/utbot/examples/arrays/ArrayOfArraysTest.kt
```diff
@@ -266,4 +266,14 @@ internal class ArrayOfArraysTest : AbstractTestCaseGeneratorTest(testClass = Arr
             { valueBefore, valueAfter -> valueAfter.withIndex().all { it.value == valueBefore[it.index] + it.index } }
         )
     }
+
+    @Test
+    fun testArrayWithItselfAnAsElement() {
+        check(
+            ArrayOfArrays::arrayWithItselfAnAsElement,
+            eq(2),
+            coverage = atLeast(percents = 94)
```
EgorkaKulikov, line 275: It feels like a magic to me...

EgorkaKulikov (approved):

utbot-framework/src/test/kotlin/org/utbot/examples/arrays/ArrayOfArraysTest.kt
CaelmBleidd: It happened because we have something like this in the bytecode:
```
label1:
    if (!predicate) goto z0 = 0
    z0 = 1
label2:
    z0 = 0
label3: 
    UtMock.assume(z0)
```
Since we do not cover branch where we have unsatisfied predicate, we won't cover the second label and lose this instructions in the final coverage

PR #50, opened 2022-05-27
Fuzzer improvements

Markoutte:
# Description

List of changes:

* Randomize fuzzer inputs
* Limit fuzzing attempts
* Mutate primitive values according to comparison operations
* Mutate string constants
* Move fuzz-relative code out of UtBotSymbolicEngine.kt
* Test added

## Type of Change

# How Has This Been Tested?

- org.utbot.framework.plugin.api.CombinationsTest
- org.utbot.framework.plugin.api.ModelProviderTest

## Manual Scenario 

1. Disable `engine.traverse()` function by commenting it in UtBotTestCaseGenerator:185
2. Build and run
3. Create a function as listed below
4. Run UtBot and check that it creates tests with all branches covered and the tests have names which end with word "FuzzerN"

```java
    public String foo(int a) {
        if (a < -2.0) {
            return "-1";
        }
        if (a > 5) {
            return "-2";
        }
        if (a == 3) {
            return "-3";
        }
        if (4L < a) {
            return "-4";
        }
        return "0";
    }
```

# Checklist:

- [ ] No new warnings


utbot-framework/src/main/kotlin/org/utbot/fuzzer/FuzzerFunctions.kt
```diff
@@ -0,0 +1,224 @@
...
+            ?.findFirstInstanceOf<Constant>()
+            ?.plainValue
+        if (stringConstantWasPassedAsThis != null) {
+            return listOf(FuzzedConcreteValue(stringClassId, stringConstantWasPassedAsThis, FuzzedOp.CH))
+        }
+    }
+    return emptyList()
+}
+
+@Suppress("UNUSED_PARAMETER")
```
denis-fokin, line 191: Please, add explanation why the parameters are unused, may be it worse to introduce a default value.

utbot-framework/src/main/kotlin/org/utbot/fuzzer/SimpleModelProvider.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.utbot.framework.plugin.api.util.kClass
+import org.utbot.fuzzer.providers.AbstractModelProvider
+import java.util.*
+import java.util.function.IntSupplier
+import kotlin.collections.ArrayDeque
+import kotlin.collections.ArrayList
+import kotlin.collections.HashMap
+import kotlin.reflect.KClass
+
+open class SimpleModelProvider(
```
denis-fokin, line 29: In my opinion simple is too general term. Could you pick a more specific name? And add some documentation to the class.

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/CartesianProduct.kt
```diff
@@ -1,18 +1,29 @@
...
     override fun iterator(): Iterator<List<T>> {
-        return Combinations(*lists.map { it.size }.toIntArray())
-            .asSequence()
-            .map { combination ->
-                combination.mapIndexedTo(mutableListOf()) { element, value -> lists[element][value] }
-            }
-            .iterator()
+        val combinations = Combinations(*lists.map { it.size }.toIntArray())
+        val sequence = if (random != null) {
+            // todo create lazy random algo for this because this method can cause OOME even we take only one value
```
denis-fokin, line 18: even we -> even if we

PR #69, opened 2022-05-30
Assume or execute concretely support

### LINKED ISSUE #64: AssumeOrExecuteConcretely support
**Description**

We have an instruction called `assume` that will remove all the paths from exploration that do not satisfy the given predicate. We use it whenever it is required to create some preconditions to the state. 

For example, we use it in `preconditionCheck` method in the wrappers to make constraints for their properties. Among other things, we limit their size. It means that we cannot find, i.e., a list, with size greater than our artificial limitation. Therefore, we lose correct branches to improve performance.

The idea is to introduce additional instruction called `assumeOrExecuteConcretely`, that will run the MUT concretely from the moment when we encounter contradiction containing the provided assumption. Thus, we'll generate additional branches.

For now, we will generate no tests for the following code:

```Java
int bigListFromParameters(List<Integer> list) {
    UtMock.assume(list != null && list.size() == 11);

    return list.size();
}
```

**Expected behavior**

Successfully generated execution with the list containing 11 elements.

**Environment**

Doesn't matter.

**Potential alternatives**

Not applicable since it is an additional API.

**Context**

There is no specific context.

CaelmBleidd:
Closes #64.

Implemented a new class Assumption that contains required constraints for the symbolic execution, but which not nessesarily should be satisfied in real (concrete) execution.

Now UtSolver has three types of constraints: 
1. Hard constraints -- they must be satisfied at any point of the analysis;
2. Soft constraints -- the solver tries to satisfy them, but they will be dropped if it is impossible or we couldn't determine if it possible during the timeout
3. Assumptions -- they must be satisfied during the symbolic execution, but if at some point of the analysis we will receive unsat status from the solver, one of the assumptions will be in the produced unsat core and without it the request to the solver is SAT, we will run concrete execution from that point.

There were added corresponding tests for it and changes in utbot-api module. 

Note: I've tried to remove all the `assume` instructions from the preconditionChecks of the wrappers, but seems like it is impossible because of untimely contradiction raised in the preconditionCheck analysis of the UtHashMap and UtHashSet (in the cycle). More details you can find in the design doc `docs/AssumptionsMechanism.md`

### DIFF
--- a/docs/AssumptionsMechanism.md
+++ b/docs/AssumptionsMechanism.md
@@ -1,0 +1,216 @@
+# Assumptions mechanism
+We have a public API that might help both us and external users to interact with UTBot. 
+This document contains detailed instructions about how to use `assume` methods of `UtMock` class and description 
+of the problems we encountered during the implementation.
+
+## Brief description
+This section contains short explanations of the meaning for mentioned functions and examples of usage.
+
+### UtMock.assume(predicate)
+
+`assume` is a method that gives the opportunity for users to say to the symbolic virtual machine that
+instructions of the MUT (Method Under Test) encountered after this instruction satisfy the given predicate.
+It is natively understandable concept and the closest analog from Java is the `assert` function. 
+When the virtual machine during the analysis encounters such instruction, it drops all the branches in the 
+control flow graph violating the predicate.
+
+Examples: 
+```java
+int foo(int x) {
+    // Here `x` is unbounded symbolic variable.
+    // It can be any value from [MIN_VALUE..MAX_VALUE] range.
+        
+    UtMock.assume(x > 0);
+    // Now engine will adjust the range to (0..MAX_VALUE].
+        
+    if (x <= 0) {
+        throw new RuntimeException("Unreachable exception");    
+    } else {
+        return 0;
+    }
+}
+
+// A function that removes all the branches with a null, empty or unsorted list.
+public List<Integer> sortedList(List<Integer> a) {
+    // An invariant that the list is non-null, non-empty and sorted
+    UtMock.assume(a != null);
+    UtMock.assume(a.size() > 0);
+    UtMock.assume(a.get(0) != null);
+    
+    for (int i = 0; i < a.size() - 1; i++) {
+        Integer element = a.get(i + 1)
+        UtMock.assume(element != null);
+        UtMock.assume(a.get(i) <= element);
+    }
+    
+    return a;
+}
+```
+
+Thus, `assume` is a mechanism to provide some known properties of the program to the symbolic engine.
+
+### UtMock.assumeOrExecuteConcretely(predicate)
+It is a similar concept to the `assume` with the only difference: it does not drop 
+paths violating the predicate, but execute them concretely from the moment of the 
+first encountered controversy. Let's take a look at the example below:
+
+```java
+int foo(List<Integer> list) {
+    // Let's assume that we have small lists only
+    UtMock.assume(list.size() <= 10);
+    
+    if (list.size() > 10) {
+        throw new RuntimeException("Unreachable branch");
+    } else {
+        return 0;    
+    }
+}
+```
+Here we decided to take into consideration lists with sizes less or equal to 10 to improve performance, and, therefore, lost
+the branch with possible exception. Let's change `assume` to `assumeOrExecuteConcretely`. 
+```java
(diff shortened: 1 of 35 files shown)


docs/AssumptionsMechanism.md
```diff
@@ -0,0 +1,216 @@
...
+    UtMock.assumeOrExecuteConcretely(list.size() <= 10);
+    
+    if (list.size() > 10) {
+        throw new RuntimeException("Now we will cover this branch");
+    } else {
+        return 0;    
+    }
+}
+```
+Now we will cover all two branches of the MUT. How did we do it?
```
denis-fokin, line 83: all two branches -> both branches?

utbot-framework/src/main/kotlin/org/utbot/engine/pc/UtSolver.kt
```diff
@@ -227,19 +243,31 @@ data class UtSolver constructor(
...
+    private fun check(
+        translatedSoft: MutableMap<BoolExpr, UtBoolExpression>,
+        translatedAssumptions: MutableMap<BoolExpr, UtBoolExpression>
+    ): UtSolverStatusKind {
+        val assumptionsInUnsatCore = mutableListOf<UtBoolExpression>()
+
         while (true) {
             val res = logger.trace().bracket("Low level check(): ", { it }) {
-                z3Solver.check(*translatedSoft.keys.toTypedArray())
+                val constraintsToCheck = translatedSoft.keys + translatedAssumptions.keys
```
SBOne-Kenobi, line 254: What happens if assumptions and softs contradict each other? We don't want to prioritize any of them?

utbot-framework/src/main/kotlin/org/utbot/engine/pc/UtSolver.kt
```diff
@@ -184,6 +196,7 @@ data class UtSolver constructor(
             copy(
                 constraints = newConstraints,
```
SBOne-Kenobi, line 197: newConstraints' status is SAT If hard constraints will be satisfied, thus `check(respectSoft = false)` will ignore softs and assumptions until calling `check(respectSoft = true)`.
You may set newConstraints' status to Undefined to fix it or change implementation of `check` method (or something else you prefer).

utbot-framework/src/main/kotlin/org/utbot/engine/pc/UtSolver.kt
CaelmBleidd: We check it later. If we have soft constraints and assumptions in an unsat core, we remove soft constraints at first and re-check satisfiability without them

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -1323,28 +1329,63 @@ class UtBotSymbolicEngine(
...
+                hardConstraints = hardConstraints,
                 softConstraints = setOfNotNull(negativeCaseSoftConstraint).asSoftConstraint(),
+                assumptions = assumption
             ) + resolvedCondition.symbolicStateUpdates.negativeCase
         )
         pathSelector.offer(negativeCaseState)
     }
 
+    /**
+     * Returns true of the next stmt is an [assumeOrExecuteConcretelyMethod] invocation, false otherwise.
```
sergeypospelov, line 1381:
```suggestion
     * Returns true if the next stmt is an [assumeOrExecuteConcretelyMethod] invocation, false otherwise.
```

utbot-framework/src/main/kotlin/org/utbot/engine/pc/UtSolver.kt
```diff
@@ -182,8 +194,12 @@ data class UtSolver constructor(
             New solver hasn't already added constraints thus we must add them again
             */
             copy(
-                constraints = newConstraints,
+                // We pass here undefined status to force calculation
+                // at the next `check` call. Otherwise, we'd ignore
+                // the given assumptions and return already calculated status.
+                constraints = newConstraints.withStatus(UtSolverStatusUNDEFINED),
```
SBOne-Kenobi, line 200: I think to reduce SMT solver calls, setting status Undefined should be done if and only if this.assumtions + assumptions are not empty.

sergeypospelov (approved):

PR #78, opened 2022-05-30
Fix wrong test resources directory in Android Studio

IdeaSeeker:
# Description

1. Fixed searching test resources folder
2. Changed order of suitable test source folders:

image

### DIFF
--- a/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/utils/RootUtils.kt
+++ b/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/utils/RootUtils.kt
@@ -11,14 +11,11 @@ import org.jetbrains.kotlin.config.ResourceKotlinRootType
 import org.jetbrains.kotlin.config.SourceKotlinRootType
 import org.jetbrains.kotlin.config.TestResourceKotlinRootType
 import org.jetbrains.kotlin.config.TestSourceKotlinRootType
-import org.jetbrains.kotlin.idea.configuration.externalProjectPath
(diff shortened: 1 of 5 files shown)


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/utils/RootUtils.kt
```diff
@@ -42,5 +45,15 @@ fun SourceFolder.isForGeneratedSources(): Boolean {
...
     val resourceProperties = jpsElement.getProperties(resourceRootTypes + testResourceRootTypes)
 
-    return properties?.isForGeneratedSources == true && resourceProperties?.isForGeneratedSources == true
+    return isInBuildGeneratedFolder() || properties?.isForGeneratedSources == true && resourceProperties?.isForGeneratedSources == true
+}
+
+/**
+ * Returns true if the [SourceFolder] is located in the `build/generated` directory.
+ */
+fun SourceFolder.isInBuildGeneratedFolder(): Boolean {
```
EgorkaKulikov, line 54: I suggest to create extension method `JavaSourceRootProperties.isAndroidStudioGeneratedSources` instead of your method in lines 54-58 and check not only that the path contains "build/generated", but also that your current IDE is Android Studio. And could you mention in method documentation that standart property has false value in Android Studio, so we need to create such hack.

EgorkaKulikov (changes requested):

denis-fokin:
Please, provide a PR description.

EgorkaKulikov (changes requested):

nikitavlaev (approved):

PR #144, opened 2022-06-03
Register throw statements of the wrapper's methods in the global graph

### LINKED ISSUE #36: Attach wrappers' methods graphs to the global graph
**Description**

For now, we treat wrappers (e.g. `UtArrayList`) as library classes. It means that we do not register edges from their methods graphs into the global graph, and, therefore, we do not explore them entirely. It might lead to missing branches because we could explore some paths that did not encounter any exception, and if this path will entirely cover the method under test, we will drop states that might find missing paths to exceptions.

**Solution**

Change the way we decide if we should attach some graphs to the global one. For now, we check if the given method belongs to a library class, and we could change it to a more specific choice.

**Comments**

As a side effect, after this change, we will have wrappers' graphs in the visualization.

CaelmBleidd:
# Description

The issue was about missed exceptions that should be found in the wrapper's methods. It happened because we treated wrappers as library classes, and, therefore, did not try to cover all the edges in their graphs.

First of all, I tried to register their graph as if they were non-library classes, but it does not work because of the Strings -- if we try to cover all of the instructions in their internal methods, we lose meaningful branches since we exceeded step limit trying to cover branches we don't really need.

So, I've decided to change a tactic. We do not register them into a global graph but remember to throw statements they contain. They, when we check whether some edge is covered, we should verify that we have covered all such statements in the method the edge belongs to.

Fixes #36

## Type of Change

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Unfortunately, it is quite hard to test, because it depends on the order of the path exploration. I've changed an instruction inside the `preconditionCheck` method in `UtArrayList` and `UtLinkedList` that caused several tests to fail. 

## Automated Testing

One of such tests is `org.utbot.examples.collections.LinkedListsTest#testElement`

## Manual Scenario 

It is quite hard to check manually.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/java/org/utbot/engine/overrides/collections/UtArrayList.java
+++ b/utbot-framework/src/main/java/org/utbot/engine/overrides/collections/UtArrayList.java
@@ -82,7 +82,8 @@ public class UtArrayList<E> extends AbstractList<E>
 
         int size = elementData.end;
         assume(elementData.begin == 0);
-        assume(size >= 0 & size <= MAX_LIST_SIZE);
+        assume(size >= 0);
+        assume(size <= MAX_LIST_SIZE);
 
         visit(this);
     }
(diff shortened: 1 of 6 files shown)


korifey (approved):
LGTM

SBOne-Kenobi (approved):

PR #209, opened 2022-06-14
Temporary directories should be removed on exit #167

### LINKED ISSUE #167: Temporary directories should be removed on exit
**Description**

When UTBot adds temporary directories (named with pattern \d+) to unpack Z3 DLLs and then never removes it.
(See _https://github.com/UnitTestBot/UTBotJava/blob/main/utbot-framework/src/main/kotlin/org/utbot/engine/z3/Z3initializer.kt#L43_)
I believe the problem is that File.deleteOnExit() doesn't work with non empty folder.
Guava misbehave in the same way when adds its "\d+-0" names folders with JavaProbe.class inside (_See org.gradle.internal.jvm.inspection.DefaultJvmMetadataDetector.writeProbeClass()_)
But Guava eats 733 bytes while we consume 14M on every "start IDE" + "generate tests with UTBot" cycle.

**To Reproduce**

Start IDE, generate some tests with UTBot.

**Expected behavior**

We have to clean our temporary files

**Actual behavior**

Garbage eats user's disk.

**Environment**

**Additional context**

Vassiliy-Kudryashov:
# Description

We don't delete some of temporary directories because deleteOnExit doesn't work for them.
The solution is to use existing FileUtil.clearTempDirectory and delete outdated files later at the next startup.

Fixes #167

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Start IntelliJ with plugin, generate some tests. All temporary directories and files should be inside $tmpDir/UTBot.
Default 'expiration period' is 3 days by default. For testing purposes create $homeDir/.utbot/settings.properties with the following line inside: `daysLimitForTempFiles=-1`.
After restart all 'old' files should be deleted automatically.

# Checklist:

- [ ] No new warnings

### DIFF
--- a/utbot-core/src/main/kotlin/org/utbot/common/FileUtil.kt
+++ b/utbot-core/src/main/kotlin/org/utbot/common/FileUtil.kt
@@ -75,12 +75,8 @@ object FileUtil {
         }.forEach { it.deleteRecursively() }
     }
 
-    private fun createTempDirectory(): Path {
-        val tempFolderPath = Paths.get(tempDirectoryPath + File.separator + utBotTempFolderPrefix)
-
-        tempFolderPath.toFile().mkdirs()
-
-        return createTempDirectory(tempFolderPath, "generated-")
+    fun createTempDirectory(prefix: String): Path {
+        return createTempDirectory(utBotTempDirectory, prefix)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


SBOne-Kenobi (approved):
LGTM

PR #218, opened 2022-06-15
RunTimeException Error in process of test generation #182

### LINKED ISSUE #182: RunTimeException Error in process of test generation
**Description**

While choosing several classes randomly for test generation, if at least one Interface has been chosen along with classes, tests are not generated at all. Besides, sometimes two RunTimeException errors appear while trying to generate tests.

**To Reproduce**

_Example:_

1. Run IDEA Commit https://github.com/UnitTestBot/UTBotJava/commit/cab4799cb7ff61fbd00571125d6e35a1c23857d5
2. Use org.utbot.examples.mock to generate tests
3. Choose a couple classes and at least one Interface
4. Check the result

**Expected behavior**

Tests are supposed to be generated for classes anyway, ignoring chosen interface.

**Actual behavior**

 If at least one Interface has been chosen along with classes, tests are not generated neither for Interface nor for classes, both are ignored.
If we choose classes only, everything goes fine, tests are generated.

**Environment**

IDEA Commit https://github.com/UnitTestBot/UTBotJava/commit/cab4799cb7ff61fbd00571125d6e35a1c23857d5
Test framework: JUnit4
Mock strategy: other packages: Mockito
Mock static: No static mocking
Timeout: 60 sec

Vassiliy-Kudryashov:
# Description

From plugin side we check and skip interfaces, thus we don't try to process them at all
From engine side we throw IllegalStateException with helpful message instead of RuntimeException

Fixes #182

## Type of Change

# How Has This Been Tested?
## Manual Scenario 
"Generate tests..." action skips interfaces, so it's disabled if there is nothing except interfaces in the context.

- [ ] No new warnings

### DIFF
--- a/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/actions/GenerateTestsAction.kt
+++ b/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/actions/GenerateTestsAction.kt
@@ -58,6 +58,7 @@ class GenerateTestsAction : AnAction() {
             e.getData(CommonDataKeys.VIRTUAL_FILE_ARRAY)?.let {
                 srcClasses += getAllClasses(project, it)
             }
+            srcClasses.removeIf { it.isInterface }
             var commonSourceRoot = null as VirtualFile?
             for (srcClass in srcClasses) {
                 if (commonSourceRoot == null) {
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/plugin/api/UtBotTestCaseGenerator.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/plugin/api/UtBotTestCaseGenerator.kt
@@ -48,6 +48,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.yield
 import mu.KotlinLogging
+import org.utbot.engine.*
 import soot.Scene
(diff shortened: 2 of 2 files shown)


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/actions/GenerateTestsAction.kt
```diff
@@ -58,6 +58,7 @@ class GenerateTestsAction : AnAction() {
             e.getData(CommonDataKeys.VIRTUAL_FILE_ARRAY)?.let {
                 srcClasses += getAllClasses(project, it)
             }
+            srcClasses.removeIf { it.isInterface }
```
Damtev, line 61: It is a bit unclear what should we do with default methods in interfaces and abstract methods in abstract classes. Could you please create an issue for investigating such cases and link it here?

Damtev (changes requested):

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/actions/GenerateTestsAction.kt
Vassiliy-Kudryashov: The issue: https://github.com/UnitTestBot/UTBotJava/issues/221

Damtev (approved):

PR #224, opened 2022-06-16
Fix bug in FeatureProcessor

### LINKED ISSUE #223: FeatureProcessor dumps empty test cases
**Description**

Feature `.csv` files contains only header line.

**To Reproduce**

1. `UtSettings.enableFeatureProcess=true`
2. Run ContestEstimator
3. Look at files in `UtSettings.featurePath`

**Expected behavior**

Each file should contains feature rows for each state in test case.

**Actual behavior**

In `UtSettings.featurePath` all files contains only header line.

Atos1337:
# Description

Add `forEach` after `generateSequence`.

Fixes #223

## Type of Change

# How Has This Been Tested?

## Automated Testing

org.utbot.features.FeatureProcessorWithRepetitionsTest

## Manual Scenario 

1. `UtSettings.enableFeatureProcess = true`
2. Run ContestEstimator
3. `UtSettings.featurePath` contains files with feature rows

### DIFF
--- a/utbot-analytics/src/main/kotlin/org/utbot/features/FeatureProcessorWithStatesRepetition.kt
+++ b/utbot-analytics/src/main/kotlin/org/utbot/features/FeatureProcessorWithStatesRepetition.kt
@@ -52,7 +52,7 @@ class FeatureProcessorWithStatesRepetition(
         val states = mutableListOf<Pair<Int, Long>>()
         var newCoverage = 0
 
-        generateSequence(executionState) { currentState ->
+        generateSequence(executionState) { it.parent }.forEach { currentState ->
             val stateHashCode = currentState.hashCode()
 
             if (currentState.features.isEmpty()) {
@@ -68,8 +68,6 @@ class FeatureProcessorWithStatesRepetition(
                     newCoverage++
                 }
             }
-
-            currentState.parent
         }
 
         generatedTestCases++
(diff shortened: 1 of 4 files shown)


utbot-analytics/src/test/java/org/utbot/features/OnePath.java
```diff
@@ -0,0 +1,7 @@
+package org.utbot.features;
+
+public class OnePath {
```
amandelpie, line 3: Please add some comments for this test feature

utbot-analytics/src/main/kotlin/org/utbot/features/FeatureProcessorWithStatesRepetition.kt
```diff
@@ -52,7 +52,7 @@ class FeatureProcessorWithStatesRepetition(
         val states = mutableListOf<Pair<Int, Long>>()
         var newCoverage = 0
 
-        generateSequence(executionState) { currentState ->
+        generateSequence(executionState) { it.parent }.forEach { currentState ->
```
amandelpie, line 55: could it.parent be an empty or null?

amandelpie (approved):
It could merged after answering on minor comments from my side

utbot-analytics/src/main/kotlin/org/utbot/features/FeatureProcessorWithStatesRepetition.kt
Atos1337: Yes, it can be null, if state is root, so we stop generation or if `UtSettings.enableFeatureProcess=false`, but in this case, we don't use `FeatureProcessor`

PR #249, opened 2022-06-20
Fix incorrect analysis steps in the SARIF report

### LINKED ISSUE #247: Incorrect analysis steps in the SARIF report for private functions
**Description**

When we generate a SARIF report for the class with a private function, there is no step (in the analysis steps) for this function from the test file.

**To Reproduce**

1. Run test generation (using any plugin) for the class with private method. For example:
```Java
public class Example {
    private int invert(int x) {
        return 1 / x;
    }
}
```
2. Open the SARIF report using VS Code and click to the "Analysis steps"

**Expected behavior**

There are two steps:
image

**Actual behavior**

There is only one step (only a line with an exception in our function)
image

**Additional context**

We are trying to find the line in the generated tests with something like `<objectName>.<methodName>(<arguments>)`, but in the case of private functions we have `<methodName>Method.invoke(<objectName>, <arguments>)`

IdeaSeeker:
# Description

Fix incorrect analysis steps in the SARIF report generated for the class with a private function.

Fixes #247

## Type of Change

# How Has This Been Tested?

## Automated Testing

`org.utbot.sarif.SarifReportTest`

## Manual Scenario 

Repeat the steps from the issue #247

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/sarif/SarifReport.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/sarif/SarifReport.kt
@@ -254,10 +254,12 @@ class SarifReport(
             return null
 
         // searching needed method call
+        val publicMethodCall = "$methodName("
+        val privateMethodCall = Regex("""$methodName.*\.invoke\(""") // using reflection
         val methodCallLineNumber = testsBodyLines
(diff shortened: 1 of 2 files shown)


SBOne-Kenobi (commented):
Where are any tests that represent this bug?

PR #288, opened 2022-06-23
Better naming for functions and comments could be deduced from fuzzer

### LINKED ISSUE #180: Better naming for functions and comments could be deduced from fuzzer routines
**Description**

Fuzzer can provide more information about values to produce a better names for functions, such as `testWithIntegerMinValue`, `testWithStringConstant`, etc. We can combine names for several parameters as well using 'and' like this: `testWithCharDefaultValueAndIntegerConstant`.

Also, fuzzer models can provide this info within the values to generate names by another methods.

**Expected behavior**

Fuzzer generated methods should have more meaningful names instead of `testMethodByFuzzer0`. For example, if there's a method `void foo(int value)` next test can be generated:

* `testFooWithIntegerDefaultValue`
* `testFooWithIntegerMinValue`
* `testFooWithIntegerConstantValue`

Tests with same group but different values can have number in the name.

Markoutte:
# Description

Generates more reasonable names of tests which were generated by fuzzer.

Fixes #180

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Create method like this:

```java
public boolean isStringShorterThan(String string, int size) {
    return string.length() < size;
}
```

Generate test with fuzzing. There's 2 test should be generated (can be a little different):

```java
@Test
@DisplayName("isStringShorterThan: string has special characters, size = 0 -> return false")
public void testIsStringShorterThanReturnsFalseWithBlankStringAndCornerCase() {
    Test180 test180 = new Test180();

    boolean actual = test180.isStringShorterThan("\n\t\r", 0);

    assertFalse(actual);
}

@Test
@DisplayName("isStringShorterThan: string = blank string, size = Int.MAX_VALUE -> return true")
public void testIsStringShorterThanReturnsTrueWithBlankStringAndCornerCase() {
    Test180 test180 = new Test180();

    boolean actual = test180.isStringShorterThan("   ", Integer.MAX_VALUE);

    assertTrue(actual);
}
```

- [ ] No new warnings


utbot-summary/src/main/kotlin/org/utbot/summary/Summarization.kt
```diff
@@ -113,14 +113,18 @@ class Summarization(val sourceFile: File?, val invokeDescriptions: List<InvokeDe
...
                             if (!GENERATE_DISPLAYNAME_FROM_TO_STYLE) {
                                 traceTags.execution.displayName = displayName
                             } else {
                                 traceTags.execution.displayName = fromToName
                             }
                         }
-                        if (GENERATE_NAMES) {
+                        if (GENERATE_NAMES
+                            // do not rewrite display name if already set
+                            && traceTags.execution.testMethodName.isNullOrBlank()) {
```
amandelpie, line 127: Probably we need to move this behaviour to settings? Like GENERATE_DISPLAY_NAMES
Because it's very direct logic in the internals of summarization module

I suggest settings like OVERRIDE_EXISITING_SUMMURY = false

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/NameSuggester.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.utbot.framework.plugin.api.UtExecutionResult
+import org.utbot.fuzzer.FuzzedMethodDescription
+import org.utbot.fuzzer.FuzzedValue
+
+/**
+ * Name suggester generates a sequence of suggested test information such as:
+ * - test name
+ * - display test name.
+ */
+interface NameSuggester {
```
amandelpie, line 12: Does it generates method names and display names only or it generates JavaDocs too? I'm a little bit confusing because in the name of proposed PR we could see `comments`

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/SingleModelNameSuggester.kt
```diff
@@ -0,0 +1,52 @@
...
+object PrimitiveModelNameSuggester : SingleModelNameSuggester {
+    override fun suggest(description: FuzzedMethodDescription, value: FuzzedValue): String? {
+        val model = value.model
+        if (model is UtPrimitiveModel) {
+            val v = model.value
+            return when {
+                v == "" -> "EmptyString"
+                v is String && v.isBlank() -> "BlankString"
+                v is String && v.isNotEmpty() -> "NonEmptyString"
+                v is Char && (v == Char.MIN_VALUE || v == Char.MAX_VALUE) -> "CornerCase"
```
amandelpie, line 22: Probably better to add some type information instead of CornerCase only?

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/TestSuggestedInfo.kt
```diff
@@ -0,0 +1,9 @@
+package org.utbot.fuzzer.names
+
+/**
+ * Information that can be used to generate tests.
```
amandelpie, line 4: To generate tests or test names?

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/TestSuggestedInfo.kt
```diff
@@ -0,0 +1,9 @@
...
+ */
+class TestSuggestedInfo(
+    val testName: String,
```
amandelpie, line 7: How do you think, could it be extended in the future with the information about main constants and mutated variables in the test body?

utbot-summary/src/main/kotlin/org/utbot/summary/Summarization.kt
```diff
@@ -113,14 +113,18 @@ class Summarization(val sourceFile: File?, val invokeDescriptions: List<InvokeDe
                         val nameIndex = namesCounter.getOrPut(name) { 0 }
                         namesCounter[name] = nameIndex + 1
                         updatedExecutions += traceTags.execution
-                        if (GENERATE_DISPLAY_NAMES) {
+                        if (GENERATE_DISPLAY_NAMES
+                            // do not rewrite display name if already set
+                            && traceTags.execution.displayName.isNullOrBlank()) {
```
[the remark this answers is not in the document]
Markoutte: I don't think it is good idea to keep global settings at all. Also, I don't like to add new settings until it really necessary. I think it is better to make this settings be customizable through the Summarization instance. But at the moment I find Summarization code is a bit messy.

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/NameSuggester.kt
Markoutte: Right now it provides only test name and test display name (by annotation). It can be extended later by adding all necessary information into TestSuggestedInfo object. There's no exact specification about comments, so, I just omitted this part and will do it later after we can understand what comments we want to see there.

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/SingleModelNameSuggester.kt
Markoutte: I played a bit with it, and it is very hard to read names like `testFooWithIntCornerCaseAndDoubleCornerCaseAndFloatCornerCase` for only 3 parameters (int, double, float).

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/names/TestSuggestedInfo.kt
Markoutte: Yes, this class is supposed to be extended if needed to add any of useful information.

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ArrayModelProvider.kt
```diff
@@ -25,7 +25,9 @@ class ArrayModelProvider(
                         length = arraySize,
                         arrayClassId.elementClassId!!.defaultValueModel(),
                         mutableMapOf()
-                    )
+                    ).fuzzed {
+                        this.summary = "%var% = ${arrayClassId.elementClassId!!.simpleName}[$arraySize]"
```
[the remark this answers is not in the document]
Markoutte: It is described as a kdoc in FuzzedValue class for summary field. The test is follow:

```
 /**
     * Summary is a piece of useful information that clarify why this value has a concrete value.
     *
     * It supports a special character `%var%` that is used as a placeholder for parameter name.
     *
     * For example:
     * 1. `%var% = 2` for a value that have value 2
     * 2. `%var% >= 4` for a value that shouldn't be less than 4
     * 3. `foo(%var%) returns true` for values that should be passed as a function parameter
     * 4. `%var% has special characters` to describe content
     */
```

utbot-summary/src/main/kotlin/org/utbot/summary/Summarization.kt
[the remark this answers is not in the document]
amandelpie: Waiting TODO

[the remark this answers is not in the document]
amandelpie: I suggest to use alternative format with todos like TODO: link to the issue or discussion on the Github. If you disagree with the settings model lets discuss it, don't hide the thoughts

amandelpie (approved):
LGTM for me

Issue #248, opened 2022-06-20
Incorrect parametrized test method on function returns boolean


EgorkaKulikov:
**Description**

Parametrized test method is created with incorrect assert statement

**To Reproduce**

Create parametrized tests with JUnit5 and NoMocks options for `IntExamples.orderCheck`

**Expected behavior**

Generated tests code contains assert that actual return values is equal to expected

**Actual behavior**

Test method is generated is follows

```java
@ParameterizedTest
    @MethodSource("provideDataForOrderCheck")
    public void parameterizedTestsForOrderCheck(IntExamples intExamples,
                                                int int1, int int2, int int3, Boolean expectedResult) {
        boolean actual = intExamples.orderCheck(int1, int2, int3);
        assertFalse(Boolean.valueOf(actual));
    }
```
It always fails when actual values is true.

PR #303, opened 2022-06-24
Correct asserts on booleans in parametrized tests

EgorkaKulikov:
# Description

Assert booleans equality with assertEquals in parametrized tests

Fixes # ([248](https://github.com/UnitTestBot/UTBotJava/issues/248))

## Type of Change

# How Has This Been Tested?

## Automated Testing

`IntExamples.orderCheck` from our utbot-samples project

## Manual Scenario 

Parametrized tests on function that return boolean. Not required if pipline passes.

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/visitor/CgJavaRenderer.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/visitor/CgJavaRenderer.kt
@@ -213,7 +213,7 @@ internal class CgJavaRenderer(context: CgContext, printer: CgPrinter = CgPrinter
         //we do not have a good string representation for two-dimensional array, so this strange if-else is required
         val returnType =
             if (element.returnType.simpleName == "Object[][]") "java.lang.Object[][]" else "${element.returnType}"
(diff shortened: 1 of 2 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/visitor/CgJavaRenderer.kt
```diff
@@ -213,7 +213,7 @@ internal class CgJavaRenderer(context: CgContext, printer: CgPrinter = CgPrinter
         //we do not have a good string representation for two-dimensional array, so this strange if-else is required
         val returnType =
             if (element.returnType.simpleName == "Object[][]") "java.lang.Object[][]" else "${element.returnType}"
-        print("public static $returnType ${element.name}() throws Exception")
+        print("public static $returnType ${element.name}()")
```
Damtev, line 216: This change is a bit unobvious to me. Do we always call in `CgParameterizedTestDataProviderMethod` body only methods that do not throw exceptions?

Damtev (commented):
For now, it looks like a mandatory fix, but I suggest creating a specific version of `deepEquals` for parameterized tests in the future.

PR #305, opened 2022-06-27
Test generation for Optional<T> in the plugin (#226)

### LINKED ISSUE #226: Test generation fails for Optional<T> class in the plugin
**Description**

No test cases are generated for methods involving Java `Optional<T>` class. `UtError`s are generated by the engine during `traverse` due to `ClassNotFound` exception: `UtOptional` class can't be loaded. This behavior may be reproduced using the Idea plugin, but models are successfully generated for the same code in the engine unit tests.

**To Reproduce**

Generate unit tests for the sample method:

```
import java.util.Optional;

public class OptionalExamples {
    public Optional<Integer> nonEmptyIfPositive(int n) {
        if (n > 0) {
            return Optional.of(n);
        } else {
            return Optional.empty();
        }
    }
}
```

**Expected behavior**

Two tests are generated, one for each branch in the code (see *Additional context* for more details).

**Actual behavior**

Errors are reported:

```
public class OptionalExamplesTest {
    ///region Test suites for executable collections.OptionalExamples.nonEmptyIfPositive


    ///region Errors report for nonEmptyIfPositive

    public void testNonEmptyIfPositive_errors() {
        // 
        // 2 occurrences of:
        // org.utbot.engine.overrides.collections.UtOptional

    }
    ///endregion

    ///endregion
}
```

**Environment**

Mockito: any configuration. Test framework: JUnit5.

**Additional context**

The primary error reason: `UtOptional` wrapper is used instead of `Optional`, but the class loader used by the test generator in the plugin can't find it. It's OK (engine-specific wrappers should never leak to the user), but `shouldMock` function checks the class and fails with `ClassNotFound` exception.

Adding a check to `shouldMock` fixes the exception, but the execution path corresponding to `Optional.empty()` return value is lost unless `checkNpeForFinalFields` setting is set to `true` (it is `false` by default). See `SootField.shouldBeNotNull()` declared in `Extensions.kt`.

A similar code sample that uses `OptionalInt` instead of `Optional<Integer>` is processed without any problems, because primitive field types are never mocked so no class check is performed.

Code sample:
```
public OptionalInt nonEmptyOptionalIntIfPositive(int n) {
    if (n > 0) {
        return OptionalInt.of(n);
    } else {
        return OptionalInt.empty();
    }
}
```

Generated tests:
```
///region SUCCESSFUL EXECUTIONS for method nonEmptyOptionalIntIfPositive(int)

/**
 * <pre>
 * Test executes conditions:
 *     {@code (n > 0): True }
 * invokes:
 *     OptionalInt::of once
 * returns from: {@code return OptionalInt.of(n); }
 * </pre>
 */
@Test
@DisplayName("nonEmptyOptionalIntIfPositive: n > 0 : True -> return OptionalInt.of(n)")
public void testNonEmptyOptionalIntIfPositive_NGreaterThanZero() {
    OptionalExamples optionalExamples = new OptionalExamples();

    OptionalInt actual = optionalExamples.nonEmptyOptionalIntIfPositive(1);

    OptionalInt expected = of(1);

    assertTrue(deepEquals(expected, actual));
}

/**
 * <pre>
 * Test executes conditions:
 *     {@code (n > 0): False }
 * </pre>
 */
@Test
@DisplayName("nonEmptyOptionalIntIfPositive: -> n > 0 : False")
public void testNonEmptyOptionalIntIfPositive_NLessOrEqualZero() throws ClassNotFoundException, Exception {
    Class optionalIntClazz = Class.forName("java.util.OptionalInt");
    OptionalInt prevEMPTY = ((OptionalInt) getStaticFieldValue(optionalIntClazz, "EMPTY"));
    try {
        OptionalInt empty = empty();
        setStaticField(optionalIntClazz, "EMPTY", empty);
        OptionalExamples optionalExamples = new OptionalExamples();

        OptionalInt actual = optionalExamples.nonEmptyOptionalIntIfPositive(0);


        assertTrue(deepEquals(empty, actual));
    } finally {
        setStaticField(OptionalInt.class, "EMPTY", prevEMPTY);
    }
}
///endregion
```

dtim:
# Description

Added a check for overridden classes in `shouldMock` to avoid access to engine classes that are not available in the plugin.

Implemented more accurate speculative marking of final fields as not null to avoid losing paths involving `Optional.empty()`,
to enable NPE checks for final fields in user code, and to avoid generating non-informative NPE tests for final fields in system
classes.

`UtSettings.checkNpeForFinalFields` is now set default (`false`) in `AbstractTestCaseGEneratorTest` and  `SummaryTestCaseGeneratorTest`.

Fixes #226 

## Details

There were two issues with test generation for `Optional<T>`.

1. The `shouldMock` function tried to load overridden classes (in particular, `UtOptional`) and failed as the entire
   `org.utbot.engine.overrides` package is not included in the `UtContext` class loader created by the plugin. 

   This commit fixes the problem an explicit check for overridden classes in `shouldMock`.

2. Execution paths involving `Optional.empty()` were lost. This was caused by the assumption that all final fields
   can be considered non-nullable to reduce the number of irrelevant NPE execution paths, especially in system classes,
   that can usually be reproduced only by reflection.

   The `Optional` class is different from most other system classes as  it uses `null` as the marker of the empty value in its final field. The "non-nullable final" optimization leads to discarding execution paths with `Optional.empty()` as UNSAT, as the final
   `value` field should be both `null` (explicit assignment) and `not null` (assumption for final fields).

   This commit introduces a new memory array to track final fields. For each NPE branch, the "value is null" condition is augmented by the "field is not final" condition. As a result, NPE branches for final fields are discarded with UNSAT, while non-final fields are not affected. The finality check is enabled for system classes only, and may be turned off by setting `UtSettings.checkNpeForFinalFields` to `true`.

## Type of Change

This commit changes the default behavior of the engine with respect to final fields nullability assumptions.

Previous behavior:
  - `null` values for final fields are not allowed.
  - Branches where the final field gets `null` value are discarded even if they are legitimate. No NPE tests are generated.

New behavior:
  - `null` values for final fields are allowed.
  - NPE branches for final fields of system classes are discarded (the only expected change for system classes is fixing the `Optional.empty()` bug and any similar bugs if they exist), but can be allowed by setting  `UtSettings.checkNpeForFinalFields` to `true`.
  - NPE tests for application classes are now generated (they may be legit or redundant, but the best way to avoid redundant tests here and not to lose legit branches is to create objects using the public API instead of reflection).

# How Has This Been Tested?

## Automated Testing

The `UtSettings.checkNpeForFinalFields = true` setting was previously used to allow NPE branch generation for final fields in `utbot-framework` unit tests. Setting it to `false` would break some tests.

This commit turns off this special mode, so `utbot-framework` use the same settings w.r.t. final fields nullability as the plugin. No new test failures have been detected in this mode.

A new test case was added: `org.utbot.examples.collections.OptionalsTest#testOptionalOfPositive`.

## Manual Scenario 

Generate tests for the following examples using the plugin.

### `Optional.empty()`

```
public class OptionalExamples {
    public Optional<Integer> nonEmptyIfPositive(int n) {
        if (n > 0) {
            return Optional.of(n);
        } else {
            return Optional.empty();
        }
    }

    public Optional<Boolean> nonEmptyIfTrue(boolean condition) {
        return condition ? Optional.of(true) : Optional.empty();
    }
}
```

For both methods, two tests should be generated, one expecting a non-empty value, and one expecting the `Optional.empty()` result. No NPE tests should be generated.

### NPE checks for application classes

```
public class FinalNulls {
    private final Integer shift;

    public FinalNulls() {
        this.shift = null;
    }

    public FinalNulls(int shift) {
        this.shift = shift;
    }

    public int plus(int a){
        return a + shift;
    }
}
```

Two tests should be generated: a test for `shift != null` and a NPE test (`shift == null`).

- [X] No new warnings
- [X] Tests that prove my change is effective

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
@@ -77,6 +77,7 @@ import org.utbot.engine.pc.mkBVConst
 import org.utbot.engine.pc.mkBoolConst
 import org.utbot.engine.pc.mkChar
 import org.utbot.engine.pc.mkEq
+import org.utbot.engine.pc.mkFalse
 import org.utbot.engine.pc.mkFpConst
 import org.utbot.engine.pc.mkInt
 import org.utbot.engine.pc.mkNot
@@ -115,6 +116,7 @@ import org.utbot.engine.util.statics.concrete.makeEnumStaticFieldsUpdates
 import org.utbot.engine.util.statics.concrete.makeSymbolicValuesFromEnumConcreteValues
 import org.utbot.framework.PathSelectorType
 import org.utbot.framework.UtSettings
+import org.utbot.framework.UtSettings.checkNpeForFinalFields
 import org.utbot.framework.UtSettings.checkSolverTimeoutMillis
 import org.utbot.framework.UtSettings.enableFeatureProcess
 import org.utbot.framework.UtSettings.pathSelectorStepsLimit
@@ -2199,8 +2201,15 @@ class UtBotSymbolicEngine(
         val chunkId = hierarchy.chunkIdForField(objectType, field)
         val createdField = createField(objectType, addr, field.type, chunkId, mockInfoGenerator)
 
-        if (field.type is RefLikeType && field.shouldBeNotNull()) {
-            queuedSymbolicStateUpdates += mkNot(mkEq(createdField.addr, nullObjectAddr)).asHardConstraint()
+        if (field.type is RefLikeType) {
+            if (field.shouldBeNotNull()) {
+                queuedSymbolicStateUpdates += mkNot(mkEq(createdField.addr, nullObjectAddr)).asHardConstraint()
+            }
+
+            // See docs/SpeculativeFieldNonNullability.md for details
+            if (field.isFinal && field.declaringClass.isLibraryClass && !checkNpeForFinalFields) {
+                markAsSpeculativelyNotNull(createdField.addr)
+            }
         }
 
         return createdField
@@ -2368,6 +2377,10 @@ class UtBotSymbolicEngine(
 
     private fun touchAddress(addr: UtAddrExpression) {
         queuedSymbolicStateUpdates += MemoryUpdate(touchedAddresses = persistentListOf(addr))
+    }
+
+    private fun markAsSpeculativelyNotNull(addr: UtAddrExpression) {
+        queuedSymbolicStateUpdates += MemoryUpdate(speculativelyNotNullAddresses = persistentListOf(addr))
     }
 
     /**
@@ -3290,9 +3303,11 @@ class UtBotSymbolicEngine(
     private fun nullPointerExceptionCheck(addr: UtAddrExpression) {
         val canBeNull = addrEq(addr, nullObjectAddr)
         val canNotBeNull = mkNot(canBeNull)
+        val notMarked = mkEq(memory.isSpeculativelyNotNull(addr), mkFalse())
+        val notMarkedAndNull = mkAnd(notMarked, canBeNull)
 
         if (environment.method.checkForNPE(environment.state.executionStack.size)) {
-            implicitlyThrowException(NullPointerException(), setOf(canBeNull))
+            implicitlyThrowException(NullPointerException(), setOf(notMarkedAndNull))
         }
 
         queuedSymbolicStateUpdates += canNotBeNull.asHardConstraint()
--- a/docs/SpeculativeFieldNonNullability.md
+++ b/docs/SpeculativeFieldNonNullability.md
@@ -1,0 +1,55 @@
+# Speculative field non-nullability assumptions
+
+## The problem
+
+When a field is used as a base for a dot call (i.e., a method call or a field access),
+the symbolic engine creates a branch corresponding to the potential `NullPointerException`
+that can occur if the value of the field is `null`. For this path, the engine generates
+the hard constraint `addr(field) == null`.
+
+If the field is marked as `@NotNull`, a hard constraint `addr(field) != null` is generated
+for it. If both constraints have been generated simultaneously, the `NPE` branch is discarded
+as the constraint set is unsatisfiable.
+
+If a field does not have `@NotNull` annotation, the `NPE` branch will be kept. This behavior
+is desirable, as it increases the coverage, but it has a downside. It is possible that
+most of generated branches would be `NPE` branches, while useful paths could be lost due to timeout.
+
+Beyond that, in many cases the `null` value of a field can't be generated using the public API
+of the class. This is particularly true for final fields, especially in system classes.
+Automatically generated tests assign `null` values to fields in questions using reflection,
+but these tests may be uninformative as the corresponding `NPE` branches would never occur
+in the real code that limits itself to the public API.
+
+## The solution
+
+To discard irrelevant `NPE` branches, we can speculatively mark fields we as non-nullable even they
+do not have an explicit `@NotNull` annotation. In particular, we can use this approach to final
+fields of system classes, as they are usually correctly initialized and are not equal `null`.
+
+At the same time, we can't always add the "not null" hard constraint for the field: it would break
+some special cases like `Optional<T>` class, which uses the `null` value of its final field
+as a marker of an empty value.
+
+The engine checks for NPE and creates an NPE branch every time the address is used
+as a base of a dot call (i.e., a method call or a field access);
+see `UtBotSymbolicEngine.nullPointerExceptionCheck`). The problem is what at that moment, we would have
+no way to check whether the address corresponds to a final field, as the corresponding node
+of the global graph would refer to a local variable. The only place where we have the complete
+information about the field is this method.
+
+We use the following approach. If the field is final and belongs to a system class,
+we mark it as a speculatively non-nullable in the memory
+(see `org.utbot.engine.Memory.speculativelyNotNullAddresses`). During the NPE check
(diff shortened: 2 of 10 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -3280,9 +3274,10 @@ class UtBotSymbolicEngine(
     private fun nullPointerExceptionCheck(addr: UtAddrExpression) {
         val canBeNull = addrEq(addr, nullObjectAddr)
         val canNotBeNull = mkNot(canBeNull)
+        val notFinalAndNull = mkAnd(mkEq(memory.isSpeculativelyNotNull(addr), mkFalse()), canBeNull)
```
Damtev, line 3277: Could you please introduce a variable for the inner expression? It is a bit difficult to read.

Damtev (changes requested):

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -32,55 +32,7 @@ import org.utbot.engine.MockStrategy.NO_MOCKS
...
-import org.utbot.engine.pc.mkBoolConst
-import org.utbot.engine.pc.mkChar
-import org.utbot.engine.pc.mkEq
-import org.utbot.engine.pc.mkFpConst
-import org.utbot.engine.pc.mkInt
-import org.utbot.engine.pc.mkNot
-import org.utbot.engine.pc.mkOr
-import org.utbot.engine.pc.select
-import org.utbot.engine.pc.store
+import org.utbot.engine.pc.*
```
CaelmBleidd, line 35: Please, check your settings for the asterisk operator. We do not use it anywhere, so this fix can cause formatting problems in the future

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -1985,6 +1938,10 @@ class UtBotSymbolicEngine(
         is JInstanceFieldRef -> {
             val instance = (base.resolve() as ObjectValue)
             recordInstanceFieldRead(instance.addr, field)
+
+            // We know that [base] is not null as we are accessing its field (dot access).
+            // At the same time, we don't want to check for NPE if [base] is a final field
+            // (or if it is a non-nullable field).
```
CaelmBleidd, line 1944: Seems like redundant comment

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -2191,8 +2148,41 @@ class UtBotSymbolicEngine(
...
+            // as non-nullable", and "address is null".
+            //
+            // For final fields, this condition can't be satisfied, as we speculatively mark final fields
+            // as non-nullable here. As a result, the NPE branch would be discarded. If a field is not final,
+            // the condition is satisfiable, so the NPE branch would stay alive.
+            //
+            // We limit this approach to the system classes only, because it is hard to speculatively assume
+            // something about non-nullability of final fields in the user code.
+
+            if (field.isFinal && !field.declaringClass.isApplicationClass && !checkNpeForFinalFields) {
```
CaelmBleidd, line 2183: `!field.declaringClass.isApplicationClass` looks like `field.declaringClass.isLibraryClass`

CaelmBleidd (approved):
Overall, LGTM

Damtev (approved):

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -2191,8 +2148,41 @@ class UtBotSymbolicEngine(
...
+            // we mark it as a speculatively non-nullable in the memory. During the NPE check
+            // we will add two constraints to the NPE branch: "address has not been speculatively marked
+            // as non-nullable", and "address is null".
+            //
+            // For final fields, this condition can't be satisfied, as we speculatively mark final fields
+            // as non-nullable here. As a result, the NPE branch would be discarded. If a field is not final,
+            // the condition is satisfiable, so the NPE branch would stay alive.
+            //
+            // We limit this approach to the system classes only, because it is hard to speculatively assume
+            // something about non-nullability of final fields in the user code.
```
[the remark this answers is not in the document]
dtim: Updated the text to add a bit of context and moved it from the comment to a separate design document: `/docs/SpeculativeFieldNonNullability.md`.

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
dtim: Thank you, updated my settings in the IDE and replaced asterisk package import with individual class imports in the file.

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
dtim: Agree, replaced negated `isApplicationClass` property with `isLibraryClass`.

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
dtim: Yes, thank you. I forgot to remove the obsolete/temporary comment, removed it now.

PR #335, opened 2022-06-28
Fuzzing consume much memory for a method with 10 and more parameters

### LINKED ISSUE #293: Fuzzing consume much memory for a method with 10 and more parameters
**Description**

Fuzzer creates an array for all combinations to randomize input combinations. It may lead to high memory consumption.

**To Reproduce**

Try to run fuzzing on method with 10 and more parameters.

**Expected behavior**

Fuzzing runs without a high memory consumption.

**Actual behavior**

There's an array that takes much memory. For example, this one was found for a method with 10 different parameters:

```
int[6553600], GC Root: Java Frame: org.utbot.fuzzer.CartesianProduct.iterator(CartesianProduct.kt:19) 
```

Markoutte:
# Description

Provides another way to randomize fuzzer generated values. This approach helps to create a sequence of shuffled unique integers  without huge memory consumption for corner cases. Previously, it was creating an array for storing as much integers as fuzzer had possible values combinations thus it leads to OOME.

Fixes #293

## Type of Change

# How Has This Been Tested?

## Automated Testing

Run PseudoShuffledIntProgressionTest to validate that an implementation is correct.

## Manual Scenario 

Create a test with 10 parameters:

```java
    public int test(int x, boolean y, String s, Integer z, float f, double g,
                    int x2, boolean y3, String s4, Integer z5, float f6, double g2) {
        return -1;
    }
```

Run test generation with fuzzer and capture memory snapshot. Find `org.utbot.fuzzer.PseudoShuffledIntProgression` object and check that retained size isn't big (in my case for 6 553 600 values it takes only 20 596 bytes).

### DIFF
--- a/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
+++ b/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
@@ -1,0 +1,139 @@
+package org.utbot.fuzzer
+
+import kotlin.math.sqrt
+import kotlin.random.Random
+
+/**
+ * Generates pseudo random values from 0 to size exclusive.
+ *
+ * In general there are 2 ways to get random order for a given range:
+ * 1. Create an array of target size and shuffle values.
+ * 2. Create a set of generated values and generate new values until they're unique.
+ *
+ * Both cases cause high memory usage when target number of values is big.
+ *
+ * The memory usage can be reduced by using a pseudo-random sequence.
+ *
+ * Algorithm to create pseudo-random sequence:
+ * 1. Separate a sequence into 2 parts: matrix-part and tail
+ * 2. Transform linear representation of the matrix into 2-array view with cols × rows arrays
+ * 3. The tail array stores values that are greater than the biggest value of the matrix (cols × rows)
+ * 4. Shuffle all arrays
+ * 5. For a given index the target value is taken from the matrix or from the tail array.
+ *
+ * Example, input size = 23
+ * ```
+ * matrix             tail
+ * -----------------------
+ *  0  5 10 15        20
+ *  1  6 11 16        21
+ *  2  7 12 17        22
+ *  3  8 13 18
+ *  4  9 14 19
+ * ```
+ * Columns are shuffled:
+ *
+ * ```
+ * matrix             tail
+ * -----------------------
+ * 10  5 15  0        20
+ * 11  6 16  1        21
+ * 12  7 17  2        22
+ * 13  8 18  3
+ * 14  9 19  4
+ * ```
+ * Rows and tail are shuffled
+ * ```
+ * matrix             tail
+ * -----------------------
+ * 12  7 17  2        22
+ * 13  8 18  3        20
+ * 14  9 19  4        21
+ * 10  5 15  0
+ * 11  6 16  1
+ * ```
+ * Merge matrix and tail:
+ * ```
+ * 12  7 17  2 22
+ * 13  8 18  3 20
+ * 14  9 19  4 21
+ * 10  5 15  0 ×
+ * 11  6 16  1 ×
+ * ```
+ *
+ * Write down the sequence: `[12, 7, 17, 2, 22, 13, 8, 18, 3, 20, 14, 9, 19, 4, 21, 10, 5, 15, 0, 11, 6, 16, 1]`.
+ *
+ * This algorithm requires only `537 552 bytes (~ 550 KB)` instead of `Int.MAX_VALUE * 4 = 8 589 934 588 bytes (~ 8 GB)`
+ * in the simple array-shuffle algorithm.
+ *
+ */
+class PseudoShuffledIntProgression : Iterable<Int> {
+    private val top: IntArray
+    private val left: IntArray
+    private val tail: IntArray
+
+    val size: Int
+
+    constructor(size: Int, random: Random = Random) : this (size, random, { sqrt(it.toDouble()).toInt() })
+
+    /**
+     * Test only constructor
+     */
+    internal constructor(size: Int, random: Random, side: (Int) -> Int) {
+        check(size >= 0) { "Size must be positive or 0 but current value is $size" }
+        this.size = size
+        var topSize = side(size)
+        if (topSize > 0 && topSize > size / topSize) {
+            topSize = size / topSize
+        }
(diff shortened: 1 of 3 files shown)


utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
```diff
@@ -0,0 +1,139 @@
...
+ * In general there are 2 ways to get random order for a given range:
+ * 1. Create an array of target size and shuffle values.
+ * 2. Create a set of generated values and generate new values until they're unique.
+ *
+ * Both cases cause high memory usage when target number of values is big.
+ *
+ * The memory usage can be reduced by using a pseudo-random sequence.
+ *
+ * Algorithm to create pseudo-random sequence:
+ * 1. Separate a sequence into 2 parts: matrix-part and tail
```
CaelmBleidd, line 18: How do we separate them? From the example we can see, that we transform the sequence into `n * n` matrix with a `size - n * n` tail. Is it always true?

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
```diff
@@ -0,0 +1,139 @@
...
+ *
+ */
+class PseudoShuffledIntProgression : Iterable<Int> {
+    private val top: IntArray
+    private val left: IntArray
+    private val tail: IntArray
+
+    val size: Int
+
+    constructor(size: Int, random: Random = Random) : this (size, random, { sqrt(it.toDouble()).toInt() })
```
CaelmBleidd, line 77: extra space

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
```diff
@@ -0,0 +1,139 @@
...
+        check(size >= 0) { "Size must be positive or 0 but current value is $size" }
+        this.size = size
+        var topSize = side(size)
+        if (topSize > 0 && topSize > size / topSize) {
+            topSize = size / topSize
+        }
+        check(topSize > 0 || size == 0) { "Side of matrix must be greater than 0 but $topSize <= 0" }
+
+        top = IntArray(size = topSize) { it }.apply { shuffle(random) }
+        left = IntArray(size = if (topSize == 0) 0 else size / topSize) { it }.apply { shuffle(random) }
```
CaelmBleidd, line 92: How `topSize` can be equal to zero here? We checked that it must be positive

CaelmBleidd: Can you add a description for the fields, please? Or, probably, it is better to change names. I believe, `top` is the number of the columns and `left` is the number of the rows?

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
```diff
@@ -0,0 +1,139 @@
...
+            }
+            top[i] * l + left[j]
+        }
+    }
+
+    fun toArray(): IntArray = IntArray(size, this::get)
+
+    override fun iterator(): IntIterator = object : IntIterator() {
+        var current = 0
+        override fun hasNext() = current < size
```
CaelmBleidd, line 136: Is it right? We stated that it `Generates pseudo random values from 0 to size exclusive.`

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
```diff
@@ -0,0 +1,139 @@
...
+ * ```
+ * matrix             tail
+ * -----------------------
+ *  0  5 10 15        20
+ *  1  6 11 16        21
+ *  2  7 12 17        22
+ *  3  8 13 18
+ *  4  9 14 19
+ * ```
+ * Columns are shuffled:
```
CaelmBleidd, line 34: Can you add an extended explanation, please? Not only with the matrix but with `top` and `left` arrays as well

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
```diff
@@ -0,0 +1,139 @@
...
+        var j = index / e
+        return if (i == t && j < tail.size) {
+            tail[j]
+        } else {
+            val o = ((index - tail.size * e) / t).toLong()
+            if (o > 0) {
+                i = ((index + o) % e).toInt()
+                j = ((index + o) / e).toInt()
+            }
+            top[i] * l + left[j]
```
CaelmBleidd, line 128: Can you add an example, just like you added for the whole algorithm, please?  For me, it is not obvious work with `tail.size * e`

CaelmBleidd (changes requested):
Some places are a bit confusing at first sight, but everything seems fine

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
Markoutte: > How topSize can be equal to zero here? We checked that it must be positive

Top size can be 0 in case when size is 0. Both arrays should be 0 in this case

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
Markoutte: Looks like yes, because for size 10 we want to generate sequence like [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]. Added a test on this

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/PseudoShuffledIntProgression.kt
Markoutte: Rewrite this algorithm to clarify this place

CaelmBleidd (approved):
LGTM

Issue #346, opened 2022-06-29
Use constructor call for package-private class


alisevych:
**Description**

If analyzed class has package-private (default) access - and there is a public or a package-private constructor - need to use the applicable constructor. Currently reflection is used and createInstance (String className) is generated.

**To Reproduce**

Preconditions:

1. Open/create a project in IntelliJ Idea
2. UTBotJava plugin is installed

1. Add the following java class:

```
 class A {

    public int a = 0;
    public int b = 7;

    public A() {
    }

    A(int a, int b) {
        this.a = a;
        this.b = b;
    }

    int sum() {
        return a + b;
    }
}
```
2. Generate Test with UTBot for this class
3. Open the generated test

**Expected behavior**

Constructor call must be used to instantiate A object.
```
        A a = new A(-255, -255);
```

**Actual behavior**

There is createInstance(String className) method generated. 
It is used to instantiate A object in the test.

```
        A a = ((A) createInstance("A"));
        a.b = -255;
        a.a = -255;
```

**Environment**

IDEA 2022.1.3
JDK 8
[UTBotJava plugin build](https://github.com/UnitTestBot/UTBotJava/actions/runs/2578156782)

alisevych:
**Additional context**

Reflection method is used instead of constructor when
- class and constructors have all public access modifiers
- fields are private
```
 public class A {

    private int a = 0;
    private int b = 7;

    public A() {
    }

    public A(int a, int b) {
        this.a = a;
        this.b = b;
    }

    int sum() {
        return a + b;
    }
}
```

**Expected behavior**
```
        A a = new A(-255, -255);
```

**Actual behavior**
```
        A a = ((A) createInstance("A"));
        setField(a, "b", -255);
        setField(a, "a", -255);
```

PR #356, opened 2022-06-30
Support package-private classes and constructors in AssembleModelGenerator

EgorkaKulikov:
# Description

Classes have package-private modifier by default, it is good enough to construct assemble models.
Java reflection does not allow to call newInstance method for a non-public class, but it may be fixed with `withAccessibility` modifier

Fixes # ([346](https://github.com/UnitTestBot/UTBotJava/issues/346))

## Type of Change

# How Has This Been Tested?

## Automated Testing

Run `AssembleModelGeneratorTests` from standard pipeline

## Manual Scenario 

Could be tested on classes with default modifier of class or it constructors.

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/assemble/AssembleModelGenerator.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/assemble/AssembleModelGenerator.kt
@@ -38,6 +38,7 @@ import org.utbot.framework.plugin.api.util.defaultValueModel
 import org.utbot.framework.plugin.api.util.executableId
 import org.utbot.framework.plugin.api.util.jClass
 import org.utbot.framework.util.nextModelName
+import java.lang.reflect.Constructor
 import java.util.IdentityHashMap
 
 /**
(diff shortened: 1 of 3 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/assemble/AssembleModelGenerator.kt
```diff
@@ -357,15 +358,21 @@ class AssembleModelGenerator(private val methodUnderTest: UtMethod<*>) {
...
         return classId.jClass.declaredConstructors
-            .filter { it.isPublic || !it.isPrivate && it.declaringClass.packageName.startsWith(methodPackageName) }
+            .filter { it.isVisible }
             .sortedByDescending { it.parameterCount }
             .map { it.executableId }
             .firstOrNull { constructorAnalyzer.isAppropriate(it) }
     }
 
+    private val ClassId.isVisible : Boolean
+            get() = this.isPublic || !this.isPrivate && this.packageName.startsWith(methodPackageName)
```
sergeypospelov, line 371: Tabs
```suggestion
        get() = this.isPublic || !this.isPrivate && this.packageName.startsWith(methodPackageName)
```

PR #357, opened 2022-06-30
utbot-instrumentation kotlinx.serialization

Domonion:
@dtim review


utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/ConcreteExecutor.kt
```diff
@@ -202,6 +203,8 @@ class ConcreteExecutor<TIResult, TInstrumentation : Instrumentation<TIResult>> p
     }
 
     fun warmup() {
+        val kek = Lifetime.Eternal
```
amandelpie, line 206: Rlly?

Issue #373, opened 2022-07-01
CLI with JDK 17 throws Unsupported class file major version 61


alisevych:
**Description**

CLI with JDK 17
Soot throws "IllegalArgumentException: Unsupported class file major version 61" when called from CLI with JDK 17.

**To Reproduce**

1. Define JDK-17 as primary (Path, JAVA_HOME)
2. Download one of the latest [CLI from master](https://github.com/UnitTestBot/UTBotJava/actions/runs/2596727243)
3. For simplicity in the folder where utbot-cli*.jar is located :
Create ExampleString.java using var (feature introduced in Java 9)

```
public class ExampleString {

    public void stringNewMethod() {
        var text = "Hello! This is String defined with var";

        text = text.toLowerCase();
        System.out.println(text);

        text = text.toUpperCase();
        System.out.println(text);
    }
}
```

4. Create ExampleRecord.java using record (feature introduced in Java 16)

```
public class ExampleRecord {

    public record Person(String name, int age) {
        public Person {
            if(age < 0) {
                throw new IllegalArgumentException("Age cannot be negative");
            }
        }
    }
}

```

5. Compile your files with JDK-17

`javac ExampleRecord.java`
`javac ExampleString.class`

6. Run utbot-cli to generate tests for it, like that:

`java -jar utbot-cli-2022.7.jar generate --source ExampleString.java --classpath "D:\Current\Java" -o ExampleStringTest.java ExampleString
`

`java -jar utbot-cli-2022.7.jar generate --source ExampleRecord.java --classpath "D:\Current\Java" -o ExampleRecordTest.java ExampleRecord`

**Expected behavior**

Tests are supposed to be generated.

**Actual behavior**

There is the following output in console:

```
WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
16:14:54.331 | INFO  | AllInMemoryClassProvider | Loaded: 2 path entries, 48 classes, 196740 bytes
16:14:54.378 | ERROR | GenerateTestsCommand | An error has occurred while generating test for snippet ExampleRecord : java.lang.IllegalArgumentException: Unsupported class file major version 61
java.lang.IllegalArgumentException: Unsupported class file major version 61
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:196)
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:177)
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:163)
[...]

PR #380, opened 2022-07-01
Add title pane with unsupported jdk notification (#373)

nikitavlaev:
# Description

Bring back notification pane with unsupported jdks.

Fixes # (373)

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Checked that 8 and 11 does not show notification, 13 and 17 -- show.

- [x] No new warnings

### DIFF
--- a/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/GenerateTestsDialogWindow.kt
+++ b/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/GenerateTestsDialogWindow.kt
@@ -2,6 +2,68 @@
 
 package org.utbot.intellij.plugin.ui
 
+import com.intellij.codeInsight.hint.HintUtil
+import com.intellij.icons.AllIcons
+import com.intellij.ide.impl.ProjectNewWindowDoNotAskOption
+import com.intellij.openapi.application.runWriteAction
+import com.intellij.openapi.command.WriteCommandAction
+import com.intellij.openapi.components.service
+import com.intellij.openapi.editor.colors.EditorColorsManager
+import com.intellij.openapi.options.ShowSettingsUtil
+import com.intellij.openapi.projectRoots.JavaSdkVersion
+import com.intellij.openapi.roots.ContentEntry
+import com.intellij.openapi.roots.ModuleRootManager
+import com.intellij.openapi.roots.ui.configuration.ClasspathEditor
+import com.intellij.openapi.roots.ui.configuration.ProjectStructureConfigurable
+import com.intellij.openapi.ui.ComboBox
+import com.intellij.openapi.ui.DialogPanel
+import com.intellij.openapi.ui.DialogWrapper
+import com.intellij.openapi.ui.Messages
+import com.intellij.openapi.ui.ValidationInfo
+import com.intellij.openapi.ui.popup.IconButton
+import com.intellij.openapi.util.Computable
+import com.intellij.openapi.vfs.StandardFileSystems
+import com.intellij.openapi.vfs.VfsUtil
+import com.intellij.openapi.vfs.VfsUtilCore.urlToPath
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.openapi.vfs.newvfs.impl.FakeVirtualFile
+import com.intellij.psi.PsiClass
+import com.intellij.psi.PsiManager
+import com.intellij.psi.PsiMethod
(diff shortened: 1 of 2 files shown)


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/GenerateTestsDialogWindow.kt
```diff
@@ -219,18 +242,77 @@ class GenerateTestsDialogWindow(val model: GenerateTestsModel) : DialogWrapper(m
             contextHelpLabel?.let { add(it, BorderLayout.LINE_END) }
         })
 
-    private fun findSdkVersion(): Int {
+    private fun findSdkVersion(): JavaVersion {
         val projectSdk = ModuleRootManager.getInstance(model.testModule).sdk
```
dtim, line 246: Although the variable name is `projectSdk`, actually the SDK for the test module is checked. I think it's a rare and possibly an artificial case, but it's possible that SDK versions do not match for the test module and the corresponding source module. For example, I can set up a project so the test module uses Java 11 while the source module uses Java 13. We will not show the "unsupported JDK" notification, as Java 11 is supported. Nevertheless, the plugin will fail with an exception due to incompatible bytecode version.

I'd suggest to check the SDK version not for the test module but for other modules as well. It seems easy to do conservatively (show the error message if at least one JDK is not supported), but may be more hard to do correctly with respect to module dependencies.

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/GenerateTestsDialogWindow.kt
```diff
@@ -219,18 +242,77 @@ class GenerateTestsDialogWindow(val model: GenerateTestsModel) : DialogWrapper(m
...
-        val sdkVersion = JavaVersion.tryParse(projectSdk?.versionString)
+        return JavaVersion.tryParse(projectSdk?.versionString)
             ?: error("No sdk found in ${model.testModule}")
```
dtim, line 248: Do you think it's OK to throw an exception if the module has no configured JDK? This exception will be presented to users as a plugin error that should be reported. I think it would be better to show an informative error message instead of just failing.

dtim (changes requested):

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/GenerateTestsDialogWindow.kt
```diff
@@ -219,18 +242,76 @@ class GenerateTestsDialogWindow(val model: GenerateTestsModel) : DialogWrapper(m
             contextHelpLabel?.let { add(it, BorderLayout.LINE_END) }
         })
 
-    private fun findSdkVersion(): Int {
-        val projectSdk = ModuleRootManager.getInstance(model.testModule).sdk
-        val sdkVersion = JavaVersion.tryParse(projectSdk?.versionString)
-            ?: error("No sdk found in ${model.testModule}")
-        return sdkVersion.feature
+    private fun findSdkVersion(): JavaVersion? {
+        val projectSdk = ModuleRootManager.getInstance(model.srcModule).sdk
```
dtim, line 246: As a different issue, we'll have to research how to compile the project with different SDK versions for different modules.

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/GenerateTestsDialogWindow.kt
```diff
@@ -219,18 +242,76 @@ class GenerateTestsDialogWindow(val model: GenerateTestsModel) : DialogWrapper(m
...
+        private val model: GenerateTestsModel,
+        private val sdkVersion: JavaVersion?,
+    ) : BorderLayoutPanel() {
+        init {
+            border = merge(empty(10), createBorder(JBColor.border(), SideBorder.BOTTOM), true)
+
+            addToLeft(JBLabel().apply {
+                icon = AllIcons.Ide.FatalError
+                text = if (sdkVersion != null) {
+                    "SDK version $sdkVersion is not supported, use ${JavaSdkVersion.JDK_1_8} or ${JavaSdkVersion.JDK_11}."
```
dtim, line 278: It would be better to use user-readable version names.

dtim (approved):
The code seems OK. Remaining questions can be addressed later.

PR #402, opened 2022-07-04
ToString support for abstract collection using makeSymbolic #391

### LINKED ISSUE #391: ArrayList.toString() - fails to generate tests
**Description**

Presence of ArrayList.toString() method leads to test generation failure.

**To Reproduce**

1. Open IntelliJ IDEA
2. Plugin is installed
3. A project with JDK 8/11 is opened
4. Add a class using ArrayList.toString() method, like:

```
public class ListToStringExample {

    public void toStringList() {
        Arrays.asList("a", "b", "c").toString();
    }
}
```

**Expected behavior**

Tests should be successfully generated

**Actual behavior**

Failed to generate unit tests... popup is shown.
Timeout increasing is suggested, but that doesn't help.
**Failure is displayed in 2-3 seconds** , timeout is not reached.

ListToString failed

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.1.3
JDK 8, 11

**Additional context**

Same AbstractCollection.toString() implementation is called for this code - same error is displayed:

```
public class ListToStringExample {

    public void toStringList() {
        List<String> list = new ArrayList<>();
        list.add("a");
        list.add("b");
        list.add("c");
        list.toString();
    }
}
```

ListToString failed 2

CaelmBleidd:
# Description

ToString for lists (as well as for other inheritors of `AbstractCollection` didn't work since it's too complicated for the engine because of StringBuilder usage somewhere inside of it. I've decided to make a temporary solution, implementing this method with `makeSymbolic` mechanism -- the method should return an unbounded symbolic variable, that will allow us to continue the analysis (in contrast to the `executeConcretely`). Later, the concrete execution will find the right result for such execution, if it's required. 

My suggestion is based on thoughts that people usually use `toString` on collections just for logging, showing or something like that, and the resulting line won't present in the path constraints. It's not 100% true, but in many cases, it will work.

Fixes #391

## Type of Change
- Breaking change. `toString` without concrete executor works differently now for the inheritors of the `AbstractCollection` and `HashSet`

# How Has This Been Tested?

## Automated Testing

Added test in the corresponding package.

## Manual Scenario 
Same test, since it is only about `toString` method on any collection.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/java/org/utbot/engine/overrides/collections/AbstractCollection.java
+++ b/utbot-framework/src/main/java/org/utbot/engine/overrides/collections/AbstractCollection.java
@@ -1,0 +1,13 @@
+package org.utbot.engine.overrides.collections;
+
+import org.utbot.api.annotation.UtClassMock;
+
+import static org.utbot.api.mock.UtMock.makeSymbolic;
+
+@UtClassMock(target = java.util.AbstractCollection.class, internalUsage = true)
+public abstract class AbstractCollection<E> implements java.util.Collection<E> {
+    @Override
+    public String toString() {
+        return makeSymbolic();
+    }
+}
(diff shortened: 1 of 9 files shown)


Damtev (approved):

dtim (approved):
LGTM

Issue #344, opened 2022-06-29
Minimize exceptions in test methods signature


EgorkaKulikov:
**Description**

The generated tests code looks as follows:

```java
 public void testSetImage2() throws Throwable, Exception  { ... }
```

Throwing both `Exception` and `Throwable` is odd. 
Can be reproduced with `ContestEstimator` on `PDVisibleSignDesigner` ,  method `setImage`

**Expected behavior**

The signature looks like

1) if Throwable is really required

```java
 public void testSetImage2() throws Throwable { ... } 
```

2) if throwable is really not required

```java
 public void testSetImage2() throws Exception  { ... }
```

**Context**

Possibly the reason of this problem is in 

```kotlin
//If [InvocationTargetException] is thrown manually in test, we 
// to add "throws Throwable" and other exceptions are not required so on.
if (methodId == getTargetException) {
        collectedExceptions.clear()
        addException(Throwable::class.id)
        return
}
```

EgorkaKulikov:
```java
public void testPDVisibleSignDesigner2() throws Exception, IOException
```

Another example on the same class

EgorkaKulikov:
```
public void testPageHeightByFuzzer7() throws Exception, ClassNotFoundException, NoSuchMethodException, SecurityException, IllegalAccessException, IllegalArgumentException, InvocationTargetException
```

PR #418, opened 2022-07-05
Minimize exceptions in signature

sofurihafe:
# Description

Filters exceptions in test method signature.

Fixes # ([344](https://github.com/UnitTestBot/UTBotJava/issues/344))
Fixes # ([250](https://github.com/UnitTestBot/UTBotJava/issues/344))

## Type of Change

Minor bug fix (non-breaking small changes)

# How Has This Been Tested?

## Automated Testing

Run utbot-samples pipeline.

## Manual Scenario 

Create a method throwing many exceptions inheriting each other. Verify that only required ones are represented in test signature.

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/context/CgContext.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/context/CgContext.kt
@@ -54,11 +54,11 @@ import kotlinx.collections.immutable.persistentListOf
 import kotlinx.collections.immutable.persistentMapOf
 import kotlinx.collections.immutable.persistentSetOf
 import org.utbot.framework.codegen.model.constructor.builtin.streamsDeepEqualsMethodId
-import org.utbot.framework.plugin.api.util.isSubtypeOf
-import org.utbot.framework.plugin.api.util.isNotSubtypeOf
-import org.utbot.framework.plugin.api.util.isCheckedException
-import org.utbot.framework.plugin.api.util.id
 import org.utbot.framework.plugin.api.util.executableId
+import org.utbot.framework.plugin.api.util.id
+import org.utbot.framework.plugin.api.util.isCheckedException
+import org.utbot.framework.plugin.api.util.isSubtypeOf
+import org.utbot.framework.plugin.api.util.jClass
 
 /**
  * Interface for all code generation context aware entities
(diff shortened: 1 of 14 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/context/CgContext.kt
```diff
@@ -232,7 +233,14 @@ internal interface CgContextOwner {
         currentExecutable = method.callable.executableId
     }
 
-    fun addException(exception: ClassId) {
+    fun addExceptionIfNeeded(exception: ClassId) {
```
Damtev, line 236: Passing `BuiltinClassId`  to this method possibly leads to `ClassNotFoundException` because of a missing class in the classpath.  Perhaps we should check it and early return in such a case.

Also, I suggest checking if passed `ClassId` is actually a checked exception - we do not need to collect an unchecked exception.

Damtev (changes requested):
Overall, LGTM

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/context/CgContext.kt
```diff
@@ -232,10 +233,18 @@ internal interface CgContextOwner {
...
-            importIfNeeded(exception)
+    fun addExceptionIfNeeded(exception: ClassId) {
+        for (existingException in collectedExceptions) {
+            if (exception isSubtypeOf existingException) {
+                return
+            }
         }
+
+        collectedExceptions
+            .removeIf { existingException -> existingException isSubtypeOf exception }
```
Damtev, line 244: Looks like you inverted the logic here - we are going to remove a super exception in case the passed exception is an inheritor of it, but here you are removing all inheritors of the passed exception.

Damtev (changes requested):

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/context/CgContext.kt
EgorkaKulikov: It is the filtering of "small" exceptions if we add "huge" ones, seems to be correct to me.

EgorkaKulikov (approved):

Damtev (changes requested):
The big problem found, need to discuss

Damtev (approved):
Discussed, everything is OK

PR #466, opened 2022-07-08
Fix the way of setting static final fields

### LINKED ISSUE #458: Cannot set static final java.io.InputStream field java.lang.System
**Description**

If we mock `java.util.Scanner` class, we get exceptions in the concrete executor -- cannot set field (see screenshot below)

**To Reproduce**

Steps to reproduce the behaviour:

1. Run any project in IntelliJ Idea
2. Create a method with scanner usage
3. Use the plugin to generate tests
4. Open the generated tests

**Expected behaviour**

Correctly generated tests for the method

**Actual behaviour**

Tests containing only exceptions

**Visual proofs**
image
image

**Environment**

You need to add `java.util.Scanner` into the set of classes to mock always (in the settings of UTBot). 

**Additional context**
There is no additional context.

sergeypospelov:
# Description

Merged `withAccessibility` and `withRemovedFinalModifer` function in `ReflectionUtil.kt`, so it won't be possible to get the wrong cached FieldAccessor field, if we use our function.

Fixes #458

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Repeated **To Reproduce** section from #458. Tests are successfully generated.

- [x] No new warnings


CaelmBleidd (approved):

Damtev (changes requested):
Do we really need this dependency on the `utbot-sample`? Please consider the idea of moving the test class to the "utbot-core" module.

Damtev (approved):

Issue #470, opened 2022-07-11
Incomplete analysis steps in SARIF report


IdeaSeeker:
**Description**

Sometimes there is only one analysis step in the SARIF report

![image](https://user-images.githubusercontent.com/54814796/178209887-4e7905b1-fcc7-4095-bd68-4a3071be0075.png)

**To Reproduce**

1. Choose the default test directory:
![image](https://user-images.githubusercontent.com/54814796/178210215-94537685-43d4-4308-89da-e508ee69e196.png)
2. Run test generation on 
```Java
public class Util {
    public static int multiply(int x, int y) {
        return Math.multiplyExact(x, y);
    }
}
```
3. Check the _Analysis steps_

**Expected behavior**

There are 2 steps: from _UtilTest.java_ and from _Util.java_.

**Actual behavior**

Only one step from the source code.

PR #471, opened 2022-07-11
Fix incomplete results in the SARIF report

### LINKED ISSUE #460: Incorrect region in SARIF report
**Description**

Sometimes we have such region in SARIF report:
```JSON
"region" : {
    "startLine" : 1,
    "startColumn" : 1
}
```

But the first line of the source code contains the package name or class name. So there can't be an unchecked exception there.

**To Reproduce**

1. Run test generation using IDEA plugin:
```Java
public class Util {
    public static int multiply(int x, int y) {
        return Math.multiplyExact(x, y);
    }
}
```
2. Open generated SARIF report
3. Find the incorrect region

**Expected behavior**

The SARIF report contains only correct regions.

**Visual proofs**

image

IdeaSeeker:
# Description

- Fixed incorrect highlighted region
- Fixed incomplete analysis steps

Fixes #460, #470

## Type of Change

# How Has This Been Tested?

## Manual Scenario

1. Run test generation using IDEA plugin
2. Check the created SARIF report. All results should:
  - refer to the correct line in the source code:
  image
  - have the correct `Generated test for this case` link:
  image
  - have at least 2 (correct) _Analysis steps_: from the test method and from the source method:
  image

# Checklist:

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/sarif/SarifReport.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/sarif/SarifReport.kt
@@ -66,16 +66,13 @@ class SarifReport(
      */
     private val relatedLocationId = 1 // for attaching link to generated test in related locations
 
-    private fun shouldProcessUncheckedException(result: UtExecutionResult) = (result is UtImplicitlyThrownException)
-            || ((result is UtOverflowFailure) && UtSettings.treatOverflowAsError)
-
     private fun constructSarif(): Sarif {
         val sarifResults = mutableListOf<SarifResult>()
         val sarifRules = mutableSetOf<SarifRule>()
 
         for (testCase in testCases) {
             for (execution in testCase.executions) {
-                if (shouldProcessUncheckedException(execution.result)) {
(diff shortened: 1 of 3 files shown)


utbot-framework/src/main/kotlin/org/utbot/sarif/SarifReport.kt
```diff
@@ -301,10 +298,24 @@ class SarifReport(
         return "..."
     }
 
-    private fun extractLineNumber(utExecution: UtExecution): Int? =
-        try {
+    /**
+     * Returns the number of the last line in the execution path.
+     */
+    private fun getLastLineNumber(utExecution: UtExecution): Int? {
```
nikitavlaev, line 304: maybe like 
```
        try {
            utExecution.path.lastOrNull()?.stmt?.javaSourceStartLineNumber
        } catch (t: Throwable) {
            // if for some reason we can't extract the last line from the path
            // return last covered instruction
            utExecution.coverage?.coveredInstructions?.lastOrNull()?.lineNumber
        }
```

nikitavlaev (approved):

PR #484, opened 2022-07-11
Use ClassId name in prettifiedName if canonical name is null

### LINKED ISSUE #459: Failed to generate tests because canonical name was not found
**Description**

Tests generation failed because `ClassId.canonicalName` has thrown an exception

**To Reproduce**

1. Run Contest Estimator with method com.google.common.base.CaseFormat.* and 300s time limit.

**Actual behavior**
```java
    public void testTo_errors()
     {
        // 
        // 1 occurrences of:
        // ClassId com.google.common.base.CaseFormat$1 does not have canonical name
    }
    
```

dtim:
# Description

Anonymous classes do not have canonical names, It makes `ClassId::canonicalName` property fail with an exception. `ClassId::prettifiedName` fails for anonymous classes as well.

This fix changes the call to exception-throwing `ClassId::canonicalName` in `prettifiedName` with an explicit call to `jClass::canonicalName` which returns a nullable value. If the class has no canonical name, `ClassId::name` is used instead.

Fixes #459 

## Type of Change

# How Has This Been Tested?

## Automated Testing

Existing tests should not fail.

## Manual Scenario 

Run contest estimator with `methodFilter = "com.google.common.base.CaseFormat.*"` and `projectFilter = null`.

Without this fix, errors like this will be presented:
```
 public void testValues_errors()
     {
        // 
        // 1 occurrences of:
        // ClassId com.google.common.base.CaseFormat$1 does not have canonical name
        
    }
```

With the fix, no these errors will be produced (many other errors will probably be generated instead, as there are related bugs and limitations in enum processing, both in engine and codegen).

- [X] No new warnings

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -642,8 +642,12 @@ open class ClassId(
     open val simpleName: String get() = jClass.simpleName
 
     /**
-     * For regular classes this is just a simple name
-     * For anonymous classes this includes the containing class and numeric indices of the anonymous class
+     * For regular classes this is just a simple name.
+     * For anonymous classes this includes the containing class and numeric indices of the anonymous class.
+     *
+     * Note: according to [java.lang.Class.getCanonicalName] documentation, local and anonymous classes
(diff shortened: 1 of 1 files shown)


Damtev (changes requested):
LGTM, but could you please extend the existing documentation (in what cases do we return just a `ClassId.name`)?

Damtev (approved):

PR #486, opened 2022-07-11
Better PR description

denis-fokin:
# Description

There was a couple of issues with the request: we needed another option for the PR type and better wording for the self-check list.


dtim (approved):
Looks good.

PR #498, opened 2022-07-12
Add sarif results minimization #490

### LINKED ISSUE #490: Minimize results in SARIF report
**Description**

Sometimes there are duplicates or similar results in the SARIF report.

For example, the user runs test generation for the class:

```Java
public class Main {
    public int functionA(int a) {
        return functionB(a);
    }
    public int functionB(int b) {
        return functionC(b);
    }
    public int functionC(int c) {
        return 1 / c;
    }
}
```

And the created SARIF report contains 3 results:

image

But only only one of them is relevant to the user, because an unchecked exception is occurred in the `functionC` in all results.

**Expected behavior**

SARIF report does not contain duplicates and similar results.

IdeaSeeker:
# Description

Minimize the displayed SARIF results.

Fixes #490 

## Type of Change

# How Has This Been Tested?

## Automated Testing

`org.utbot.sarif.SarifReportTest`

## Manual Scenario 

See details in the issue #490

# Checklist:


sergeypospelov (changes requested):
Everything is OK, but I'd like to see a little bit more tests. Also maybe add some simple clarifying examples of minimization to the commentaries or even to the documentation? (it's just a suggestion)

PR #515, opened 2022-07-13
Generate test by fuzzing for methods with no parameters #511

### LINKED ISSUE #511: Generate test by fuzzing for methods with no parameters
**Description**

Fuzzing should generate tests for methods without parameters. Usually, these methods have logic using internal object fields. For example:

```java
public class NestedClasses {
    public int field;

    public boolean testObject() {
        if (field > 10) {
            return true;
        }
        return false;
    }
}
```

**Expected behavior**

Fuzzing should provide several test with values for a field (for example, 10 and 11).

**Environment**

This problem is bound to the fact that fuzzer uses simple creation of _this object_. For example, fuzzer uses reflection to create _this instance_ when method's class doesn't have public empty constructor.

**Potential alternatives**

There's no alternatives

Markoutte:
# Description

This PR resolves several issues:
* _This instance_ now created with any public constructor. Reflection is used only if there's no another ways to create _this instance_
* ModelProvider changed its signature to provide a sequence. This is needed to supply values lazily (for example, generate first 10 values and take random)
* For methods without parameters fuzzer will generate models to change _this instance_
* No more fallback for a regular value generating. Thus, fuzzer doesn't generate test using reflection anymore (except case with _this instance_)

Fixes #511

## Type of Change

# How Has This Been Tested?

## Automated Testing

`org.utbot.framework.plugin.api.ModelProviderTest`

## Manual Scenario 

Test method `foo` from this example:

```java
package com.company.objects;

public class TestThisInstance {

    MyObject value;

    TestThisInstance() {}

    TestThisInstance(int value) {
        this.value = new MyObject();
        this.value.setField(value);
    }

    public boolean foo() {
        if (value.getField() < 100 || value.getField() == 102) {
            return true;
        }
        return value.getField() % 2 == 0;
    }

}

class MyObject {
    private int field;

    public int getField() {
        return field;
    }

    public void setField(int field) {
        this.field = field;
    }
}
```

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
@@ -64,6 +64,7 @@ import org.utbot.framework.plugin.api.EnvironmentModels
 import org.utbot.framework.plugin.api.Instruction
 import org.utbot.framework.plugin.api.MissingState
 import org.utbot.framework.plugin.api.Step
+import org.utbot.framework.plugin.api.UtAssembleModel
 import org.utbot.framework.plugin.api.UtConcreteExecutionFailure
 import org.utbot.framework.plugin.api.UtError
 import org.utbot.framework.plugin.api.UtExecution
@@ -79,6 +80,7 @@ import org.utbot.framework.plugin.api.util.id
 import org.utbot.framework.plugin.api.util.utContext
 import org.utbot.framework.plugin.api.util.description
 import org.utbot.framework.util.jimpleBody
+import org.utbot.framework.plugin.api.util.voidClassId
 import org.utbot.fuzzer.FallbackModelProvider
 import org.utbot.fuzzer.FuzzedMethodDescription
 import org.utbot.fuzzer.FuzzedValue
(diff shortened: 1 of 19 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -462,7 +486,7 @@ class UtBotSymbolicEngine(
                         fullPath = emptyList(),
                         coverage = concreteExecutionResult.coverage,
                         testMethodName = testMethodName?.testName,
-                        displayName = testMethodName?.displayName
+                        displayName = testMethodName?.takeIf { hasMethodUnderTestParametersToFuzz }?.displayName
```
dtim, line 489: If I understand this code right, we will not have display name for tests without parameters at all to avoid explicit "this" argument in the name?

dtim (approved):
There are a couple of minor comments, but generally looks OK.

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -396,7 +401,9 @@ class UtBotSymbolicEngine(
                 null
             }
             else -> {
-                fallbackModelProvider.toModel(methodUnderTest.clazz).apply {
+                ObjectModelProvider { nextDefaultModelId++ }.withFallback(fallbackModelProvider).generate(
+                    FuzzedMethodDescription("thisInstance", voidClassId, listOf(methodUnderTest.clazz.id), constantValues)
+                ).take(10).shuffled(Random(0)).map { it.value.model }.first().apply {
```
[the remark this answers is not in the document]
Markoutte: Yes, fuzzer should randomize but be reproducable for same method

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
Markoutte: Yes, it's true

PR #544, opened 2022-07-18
Fix possible NPE for package name in rarely cases

Markoutte:
# Description

In some cases `Class#getPackage` can return null. This should fix these cases.

## Type of Change

- [x] No new warnings

### DIFF
--- a/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ObjectModelProvider.kt
+++ b/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ObjectModelProvider.kt
@@ -153,7 +153,7 @@ class ObjectModelProvider : ModelProvider {
 
         private fun isAccessible(member: Member, packageName: String?): Boolean {
             return isPublic(member.modifiers) ||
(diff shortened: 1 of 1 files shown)


utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ObjectModelProvider.kt
```diff
@@ -153,7 +153,7 @@ class ObjectModelProvider : ModelProvider {
 
         private fun isAccessible(member: Member, packageName: String?): Boolean {
             return isPublic(member.modifiers) ||
-                    (isPackagePrivate(member.modifiers) && member.declaringClass.`package`.name == packageName)
+                    (packageName != null && isPackagePrivate(member.modifiers) && member.declaringClass.`package`?.name == packageName)
```
dtim, line 156: It would be nice to have a comment to explain why the check for `packageName != null` is here.

dtim (approved):
LGTM, but it would be nice to have some comments as the logic becomes not so clear as before.

PR #547, opened 2022-07-18
Implement new approach to create variables in parametrized tests

EgorkaKulikov:
# Description

Currently parametrized test variables contains a lot of hacks in variable creation process.
For example, `expectedResult` in method parameters and asserts is represented by to different `CgVariable` instances
Also there is a strange hack with wrappers for `actual` variable.

We need to remove all this ugly hacks.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

## Automated Testing

`utbot-samples` and `ContestEstimator`

## Manual Scenario 

Smoke regression testing of Codegen


utbot-framework/src/main/kotlin/org/utbot/framework/assemble/AssemblePrimitiveWrapper.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Creates [UtAssembleModel] of the wrapper for a given [UtPrimitiveModel].
+ */
+fun assemble(model: UtPrimitiveModel): UtAssembleModel {
+    val modelType = model.classId
+    val assembledModelType = wrapIfPrimitive(modelType)
+
+    val constructorCall = when (modelType) {
+        shortClassId -> java.lang.Short::class.java.getConstructor(Short::class.java)
```
ArsenHD, line 24: It seems like `byte` is missing here

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgMethodConstructor.kt
```diff
@@ -1059,13 +1045,19 @@ internal class CgMethodConstructor(val context: CgContext) : CgContextOwner by c
...
+                    when (expected) {
+                        is CgLiteral -> {
+                            // Literal can only be Primitive or String, can use equals here
+                            testFrameworkManager.assertEquals(expected, actual)
+                        }
+                        is CgNotNullAssertion -> {
+                            require(expected.expression is CgVariable) {
+                                "Only Variable wrapped in CgNotNullAssertion is supported in deep equals"
+                            }
+                            currentBlock = currentBlock.addAll(generateDeepEqualsAssertion(expected.expression, actual))
```
ArsenHD, line 1057: Maybe statements of `generateDeepEqualsAssertion()` should be added to the `currentBlock` inside of this method, just like it is done in `generateDeepEqualsOrNullAssertion()` (a couple lines below). So, `generateDeepEqualsAssertion()` will be a void method and will add statements to the current block itself.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/tree/CgElement.kt
```diff
@@ -580,7 +582,17 @@ open class CgVariable(
  * - in Java it is an equivalent of [CgVariable]
  * - in Kotlin the difference is in addition of "!!" to the name
  */
-class CgNotNullVariable(name: String, type: ClassId) : CgVariable(name, type)
+class CgNotNullAssertion(val expression: CgExpression) : CgValue {
+    override val type: ClassId
+        get() {
+            val expressionType = expression.type
+            return ClassId(
```
ArsenHD, line 589: It just occurred to me that we should also consider a case when `expressionType` is a `BuiltinClassId`. If that is the case, then here we are creating a `ClassId` from a `BuiltinClassId`, which may lead to errors if we try to access `jClass`. I suggest to write it like that:
```
return when (val expressionType = expression.type) {
    is BuiltinClassId -> BuiltinClassId(
        name = expressionType.name,
        canonicalName = expressionType.canonicalName,
        simpleName = expressionType.simpleName,
        isNullable = false
    )
    else -> ClassId(
        expressionType.name,
        expressionType.elementClassId,
        isNullable = false
    )
}
```

But that will also require adding a `isNullable: Boolean` parameter to the `BuiltinClassId` constructor. Just add this parameter and pass it to the `ClassId` constructor.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/tree/CgElement.kt
```diff
@@ -604,6 +616,21 @@ data class CgParameterDeclaration(
...
 }
 
+/**
+ * Test methd parameter can be one of the following types:
+ * - this instance for method under test (MUT)
+ * - argument of MUT with a certain index
+ * - result expected from MUT with the given arguments
+ * - exception expected from MUT with the given arguments
+ */
+sealed class CgParameterType {
```
ArsenHD, line 626: I suggest we change this name to `CgParameterKind`. The word `type` will probably confuse people.

ArsenHD (approved):

PR #583, opened 2022-07-25
Added an approach to differ tests produced by Fuzzer and SymbolicExecutor #581

### LINKED ISSUE #581: There is no approach to differ tests produced by Fuzzer and SymbolicExecutor
**Description**

At this moment there is no way to differ tests produced by Fuzzer and SymbolicExecutor

When we use Trace Summarization framework to generate cluster descriptions, we need a stable and good approach to differ these groups of tests.

The existing solution uses the heuristic with the empty path in UtExecution, but it's known that paths could be empty in tests produced by SymbolicExecutor

amandelpie:
# Description

This PR adds the ability to  differ tests produced by Fuzzer and SymbolicExecutor.
It also improves filtration of executions to be used for summaries generation.

Fixes #581

## Type of Change

# How Has This Been Tested?

No new tests and no new behavior to be tested.

- [x] No new warnings


utbot-framework/src/main/kotlin/org/utbot/external/api/UtBotJavaApi.kt
```diff
@@ -240,12 +241,13 @@ object UtBotJavaApi {
...
-            emptyList(),
-            mutableListOf(),
-            listOf()
+            stateBefore = testInfo.initialState,
+            stateAfter = testInfo.initialState, // it seems ok for concrete execution
+            result = utExecutionResult,
+            instrumentation = emptyList(),
+            path = mutableListOf(),
+            fullPath = listOf(),
+            createdBy = UtExecutionCreator.SYMBOLIC_ENGINE
```
Markoutte, line 250: Looks like this execution doesn't belong nor fuzzer nor symbolic engine, because it can run concreate executor to get a value. But at this point we cannot say if the origin testInfo was generate by fuzzer or SE, or it is syntetic execution. Looks like this is good example where `null` can be used

amandelpie: Looks like you are right

amandelpie:
@Markoutte please have a look, you suggestion was fixed

Markoutte (approved):

PR #592, opened 2022-07-26
Add support for classes with hidden fields #563

### LINKED ISSUE #563: Errors in test generation when there is variable hiding in src code
**Description**

When we generate tests for a class with a field that hides field from superclass, incorrect tests are generated.

**To Reproduce**

Launch the plugin on `class C` in the following example:

```Java
class A {
    public int f;
    void setF(int val) { this.f = val; }
    int getF() { return this.f; }
}

class B extends A {
    public int f;
}

public class C {
    boolean foo(B b) {
        return b.getF() == b.f + 1;
    }
    boolean bar(B b) {     // replace body with return b.getF() == b.f and test generation will work ok
        if (b.getF() == b.f)  
            return true;
        else
            return false;
    }
}

```

**Expected behavior**

Correct tests with complete coverage are generated.

**Actual behavior**

For function `bar` plugin generates incorrect test `testBar_BGetFEqualsBF` (assertEquals in this test fails when launching).
For function `foo` there is no testcase in which return value would be true.

Both problems disappear if we rename field in `class A` to any other name.

Incorrect test generated:
```Java
    @Test
    @DisplayName("bar: BGetF -> return true")
    public void testBar_BGetFEqualsBF() {
        C c = new C();
        B b = new B();
        b.f = 1;

        boolean actual = c.bar(b);

        assertFalse(actual);

        int finalBF = b.f;

        assertEquals(0, finalBF);
    }
```

**Environment**

plugin, Java11

**Additional context**

The problem also happens if we mark the field in `class B` as private: in some of the generated tests there will be accesses to this private member, so generated tests are not even compilable. Try launching test generation on `C.baz()` in this example:
```Java
package inheritanceIssues;

class A {
    public int f;
    void setF(int val) { this.f = val; }
    int getF() { return this.f; }
}

class B extends A {
    private int f;
}

public class C {
    boolean baz(B b) {
        return b.getF() == 1;
    }
}
```

volivan239:
# Description

Now if class has hidden fields, they can be recognized by engine and correct code for instantiation of such classes will be generated.

Fixes #563

## Type of Change

# How Has This Been Tested?

## Automated Testing

Switched `org.utbot.examples.objects.HiddenFieldExampleTest#testCheckSuccField` to enabled, added `org.utbot.examples.objects.HiddenFieldAccessModifiersTest`

## Manual Scenario 

Tested plugin on examples from #563 -- generates tests the way as expected.

- [ ] No new warnings

### DIFF
--- a/utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
+++ b/utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
@@ -7,22 +7,9 @@ import java.lang.reflect.Method
 
 val Class<*>.packageName: String get() = `package`?.name?:""
 
-fun Class<*>.findField(name: String): Field =
-    findFieldOrNull(name) ?: error("Can't find field $name in $this")
-
-fun Class<*>.findFieldOrNull(name: String): Field? = generateSequence(this) { it.superclass }
-    .mapNotNull {
-        try {
-            it.getField(name)
-        } catch (e: NoSuchFieldException) {
-            try {
-                it.getDeclaredField(name)
-            } catch (e: NoSuchFieldException) {
-                null
-            }
-        }
-    }
-    .firstOrNull()
+fun Class<*>.findDirectAccessedField(fieldName: String): Field? = generateSequence(this) { it.superclass }
+    .flatMap { it.declaredFields.asSequence() }
+    .firstOrNull { it.name == fieldName }
 
 fun Method.invokeCatching(obj: Any?, args: List<Any?>) = try {
     Result.success(invoke(obj, *args.toTypedArray()))
--- a/utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
+++ b/utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
@@ -7,7 +7,12 @@ import java.lang.reflect.Method
 
 val Class<*>.packageName: String get() = `package`?.name?:""
 
-fun Class<*>.findDirectAccessedField(fieldName: String): Field? = generateSequence(this) { it.superclass }
+/**
+ * Returns the Field that would be used by compiler when you try to direct access [fieldName] from code, i.e.
+ * when you write `${this.name}.$fieldName`.
+ * If there is no field named [fieldName] in the class, null is returned
+ */
+fun Class<*>.findDirectAccessedFieldOrNull(fieldName: String): Field? = generateSequence(this) { it.superclass }
     .flatMap { it.declaredFields.asSequence() }
     .firstOrNull { it.name == fieldName }
 
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -519,20 +519,13 @@ data class UtDirectSetFieldModel(
     val fieldId: FieldId,
     val fieldModel: UtModel,
 ) : UtStatementModel(instance) {
-    private val fieldName
-        get() =
-            if (instance.classId == fieldId.declaringClass)
-                "${instance.modelName}.${fieldId.name}"
-            else
-                "${instance.modelName}.(${fieldId.declaringClass.name})${fieldId.name}"
-
 
     override fun toString(): String = withToStringThreadLocalReentrancyGuard {
             val modelRepresentation = when (fieldModel) {
                 is UtAssembleModel -> fieldModel.modelName
                 else -> fieldModel.toString()
             }
-            "$fieldName = $modelRepresentation"
+            "${instance.modelName}.${fieldId.name} = $modelRepresentation"
         }
 
 }
@@ -813,6 +806,9 @@ enum class FieldIdStrategyValues {
  */
 open class FieldId(val declaringClass: ClassId, val name: String) {
 
+    init {
+        //assert(declaringClass.jClass.declaredFields.any { it.name == name })
+    }
     object Strategy {
         var value: FieldIdStrategyValues = FieldIdStrategyValues.Soot
     }
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -20,7 +20,7 @@ import org.utbot.framework.plugin.api.util.charClassId
 import org.utbot.framework.plugin.api.util.constructor
 import org.utbot.framework.plugin.api.util.doubleClassId
 import org.utbot.framework.plugin.api.util.executableId
-import org.utbot.framework.plugin.api.util.findFieldOrNull
+import org.utbot.framework.plugin.api.util.fieldOrNull
 import org.utbot.framework.plugin.api.util.floatClassId
 import org.utbot.framework.plugin.api.util.id
 import org.utbot.framework.plugin.api.util.intClassId
@@ -347,7 +347,7 @@ data class UtCompositeModel(
             if (fields.isNotEmpty()) {
                 append(" ")
                 append(fields.entries.joinToString(", ", "{", "}") { (field, value) ->
-                    if (value.classId != classId || value.isNull()) "${field.name}: $value" else "${field.name}: not evaluated"
+                    if (value.classId != classId || value.isNull()) "(${field.declaringClass}) ${field.name}: $value" else "${field.name}: not evaluated"
                 }) // TODO: here we can get an infinite recursion if we have cyclic dependencies.
             }
             if (mocks.isNotEmpty()) {
@@ -863,7 +863,7 @@ open class FieldId(val declaringClass: ClassId, val name: String) {
         return result
     }
 
-    override fun toString() = declaringClass.findFieldOrNull(name).toString()
+    override fun toString() = declaringClass.fieldOrNull(this).toString()
 }
 
 inline fun <T> withReflection(block: () -> T): T {
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -519,12 +519,20 @@ data class UtDirectSetFieldModel(
     val fieldId: FieldId,
     val fieldModel: UtModel,
 ) : UtStatementModel(instance) {
+    private val fieldName
+        get() =
+            if (instance.classId == fieldId.declaringClass)
+                "${instance.modelName}.${fieldId.name}"
+            else
+                "${instance.modelName}.(${fieldId.declaringClass.name})${fieldId.name}"
+
+
     override fun toString(): String = withToStringThreadLocalReentrancyGuard {
             val modelRepresentation = when (fieldModel) {
                 is UtAssembleModel -> fieldModel.modelName
                 else -> fieldModel.toString()
             }
-            "${instance.modelName}.${fieldId.name} = $modelRepresentation"
+            "$fieldName = $modelRepresentation"
         }
 
 }
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -20,7 +20,7 @@ import org.utbot.framework.plugin.api.util.charClassId
 import org.utbot.framework.plugin.api.util.constructor
 import org.utbot.framework.plugin.api.util.doubleClassId
 import org.utbot.framework.plugin.api.util.executableId
-import org.utbot.framework.plugin.api.util.fieldByIdOrNull
+import org.utbot.framework.plugin.api.util.findFieldByIdOrNull
 import org.utbot.framework.plugin.api.util.floatClassId
 import org.utbot.framework.plugin.api.util.id
 import org.utbot.framework.plugin.api.util.intClassId
@@ -856,7 +856,7 @@ open class FieldId(val declaringClass: ClassId, val name: String) {
         return result
     }
 
-    override fun toString() = declaringClass.fieldByIdOrNull(this).toString()
+    override fun toString() = declaringClass.findFieldByIdOrNull(this).toString()
 }
 
 inline fun <T> withReflection(block: () -> T): T {
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -20,7 +20,7 @@ import org.utbot.framework.plugin.api.util.charClassId
 import org.utbot.framework.plugin.api.util.constructor
 import org.utbot.framework.plugin.api.util.doubleClassId
 import org.utbot.framework.plugin.api.util.executableId
-import org.utbot.framework.plugin.api.util.findFieldByIdOrNull
+import org.utbot.framework.plugin.api.util.field
 import org.utbot.framework.plugin.api.util.floatClassId
 import org.utbot.framework.plugin.api.util.id
 import org.utbot.framework.plugin.api.util.intClassId
@@ -30,6 +30,7 @@ import org.utbot.framework.plugin.api.util.jClass
 import org.utbot.framework.plugin.api.util.longClassId
 import org.utbot.framework.plugin.api.util.method
 import org.utbot.framework.plugin.api.util.primitiveTypeJvmNameOrNull
+import org.utbot.framework.plugin.api.util.safeField
 import org.utbot.framework.plugin.api.util.shortClassId
 import org.utbot.framework.plugin.api.util.toReferenceTypeBytecodeSignature
 import org.utbot.framework.plugin.api.util.voidClassId
@@ -856,7 +857,7 @@ open class FieldId(val declaringClass: ClassId, val name: String) {
         return result
     }
 
-    override fun toString() = declaringClass.findFieldByIdOrNull(this).toString()
+    override fun toString() = safeField.toString()
 }
 
 inline fun <T> withReflection(block: () -> T): T {
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -806,9 +806,6 @@ enum class FieldIdStrategyValues {
  */
 open class FieldId(val declaringClass: ClassId, val name: String) {
 
-    init {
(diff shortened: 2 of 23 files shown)


utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
```diff
@@ -7,22 +7,9 @@ import java.lang.reflect.Method
...
-        } catch (e: NoSuchFieldException) {
-            try {
-                it.getDeclaredField(name)
-            } catch (e: NoSuchFieldException) {
-                null
-            }
-        }
-    }
-    .firstOrNull()
+fun Class<*>.findDirectAccessedField(fieldName: String): Field? = generateSequence(this) { it.superclass }
```
CaelmBleidd, line 10: Please, add documentation here. Especially about what 'directAccessed' means

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -855,7 +856,7 @@ open class FieldId(val declaringClass: ClassId, val name: String) {
         return result
     }
 
-    override fun toString() = declaringClass.findFieldOrNull(name).toString()
+    override fun toString() = declaringClass.fieldOrNull(this).toString()
```
CaelmBleidd, line 859: Why not `findField` anymore?

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/util/IdUtil.kt
```diff
@@ -286,9 +285,17 @@ val ClassId.isMap: Boolean
 val ClassId.isIterableOrMap: Boolean
     get() = isIterable || isMap
 
-fun ClassId.findFieldOrNull(fieldName: String): Field? = jClass.findFieldOrNull(fieldName)
+fun ClassId.fieldOrNull(fieldId: FieldId): Field? {
+    if (this.isNotSubtypeOf(fieldId.declaringClass))
+        return null
```
CaelmBleidd, line 290: Please, do not omit brackets for multiline constructions. Minor: redundant `this`

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/util/IdUtil.kt
```diff
@@ -286,9 +285,17 @@ val ClassId.isMap: Boolean
...
+    return try {
+        fieldId.field
+    } catch (e: IllegalStateException) {
+        null
```
CaelmBleidd, line 294: I believe, it is exception-dependent logic. Why you can't do it without throwing an exception? It is not a cheap operation after all. Moreover, please, add an empty line before the return

utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
CaelmBleidd, line 10: Why it is not `orNull` anymore?

utbot-framework/src/main/kotlin/org/utbot/engine/Hierarchy.kt
```diff
@@ -28,10 +28,11 @@ class Hierarchy(private val typeRegistry: TypeRegistry) {
         type as? RefType ?: error("$type is not a refType")
 
         val realType = typeRegistry.findRealType(type) as RefType
+        val realFieldDeclaringClass = typeRegistry.findRealType(field.declaringClass.type) as RefType
```
CaelmBleidd, line 31: It is a declaringType

utbot-framework/src/main/kotlin/org/utbot/engine/Hierarchy.kt
```diff
@@ -28,10 +28,11 @@ class Hierarchy(private val typeRegistry: TypeRegistry) {
...
 
-        val ancestorType = ancestors(realType.sootClass.id).lastOrNull { it.declaresField(field.subSignature) }?.type
-            ?: error("No such field ${field.subSignature} found in ${realType.sootClass.name}")
-        return ChunkId("$ancestorType", field.name)
+        if (realFieldDeclaringClass.sootClass !in ancestors(realType.sootClass.id))
```
CaelmBleidd, line 33: Please, do not omit brackets

CaelmBleidd, line 33: Why do we always have such a field in ancestors?

utbot-framework/src/main/kotlin/org/utbot/engine/Hierarchy.kt
CaelmBleidd, line 31: Why did it work earlier? Please, add corresponding comment here with explanation and example

utbot-framework/src/main/kotlin/org/utbot/engine/Strings.kt
```diff
@@ -313,7 +313,7 @@ sealed class UtAbstractStringBuilderWrapper(className: String) : BaseOverriddenW
 
         val arrayValuesChunkId = typeRegistry.arrayChunkId(charArrayType)
 
-        val valuesFieldChunkId = hierarchy.chunkIdForField(utStringClass.type, overriddenClass.valueField)
+        val valuesFieldChunkId = hierarchy.chunkIdForField(utStringClass.type, utStringClass.valueField)
```
CaelmBleidd, line 316: Are you sure? Why? 

Moreover, if it is true, do we write in this field in all other places in the wrapper? Not only in the resolver

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -639,7 +639,7 @@ class Traverser(
             SECURITY_FIELD_SIGNATURE -> SecurityManager()
             FIELD_FILTER_MAP_FIELD_SIGNATURE -> mapOf(Reflection::class to arrayOf("fieldFilterMap", "methodFilterMap"))
             METHOD_FILTER_MAP_FIELD_SIGNATURE -> emptyMap<Class<*>, Array<String>>()
-            else -> declaringClass.id.jClass.findField(field.name).let { it.withAccessibility { it.get(null) } }
+            else -> field.fieldId.field.let { it.withAccessibility { it.get(null) } }
```
CaelmBleidd, line 642: Obviously, a problem with naming: `field.fieldId.field`

Moreover, now we have unused `declaringClass` variable

utbot-framework/src/main/kotlin/org/utbot/engine/ValueConstructor.kt
```diff
@@ -214,7 +214,7 @@ class ValueConstructor {
 
         model.fields.forEach { (field, fieldModel) ->
             val declaredField =
-                javaClass.findFieldOrNull(field.name) ?: error("Can't find field: $field for $javaClass")
+                model.classId.fieldOrNull(field) ?: error("Can't find field: $field for $javaClass")
```
CaelmBleidd, line 217: Returning back to the `fieldOrNull` implementation, here it looks like: 
`if we don't have a field, throw an exception. I'll catch it, return null and throw an exception once again`

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
```diff
@@ -128,21 +127,17 @@ internal class CgVariableConstructor(val context: CgContext) :
...
             // Example:
             // Object heapByteBuffer = createInstance("java.nio.HeapByteBuffer");
             // branchRegisterRequest.byteBuffer = heapByteBuffer;
             // byteBuffer is field of type ByteBuffer and upper line is incorrect
-            val canFieldBeDirectlySetByVariableAndFieldTypeRestrictions =
-                fieldFromVariableSpecifiedType != null && fieldFromVariableSpecifiedType.type.id == variableForField.type
-            if (canFieldBeDirectlySetByVariableAndFieldTypeRestrictions && fieldId.canBeSetIn(testClassPackageName)) {
-                // TODO: check if it is correct to use declaringClass of a field here
-                val fieldAccess = if (field.isStatic) CgStaticFieldAccess(fieldId) else CgFieldAccess(obj, fieldId)
+            if (field != null && field.type.id == variableForField.type && fieldId.canBeSetIn(testClassPackageName)) {
```
CaelmBleidd, line 139: @Damtev or @EgorkaKulikov, please, look over these changes with code generation

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/DslUtil.kt
```diff
@@ -72,7 +76,10 @@ fun stringLiteral(string: String) = CgLiteral(stringClassId, string)
 
 // non-static fields
 operator fun CgExpression.get(fieldId: FieldId): CgFieldAccess =
-    CgFieldAccess(this, fieldId)
+    if (type.jClass.findDirectAccessedField(fieldId.name) != fieldId.field)
+        CgFieldAccess(CgTypeCast(fieldId.declaringClass, this), fieldId)
+    else
+        CgFieldAccess(this, fieldId)
```
CaelmBleidd, line 82: brackets

utbot-framework/src/test/kotlin/org/utbot/examples/UtModelTestCaseChecker.kt
```diff
@@ -167,8 +165,8 @@ internal abstract class UtModelTestCaseChecker(
     /**
      * Finds field model in [UtCompositeModel] and [UtAssembleModel]. For assemble model supports direct field access only.
      */
-    protected fun UtModel.findField(fieldName: String): UtModel =
-        findField(this.classId.jClass.findField(fieldName).fieldId)
+    protected fun UtModel.findField(fieldName: String, declarationClass: ClassId = this.classId): UtModel =
```
CaelmBleidd, line 168: is `declaration` the right word?

utbot-sample/src/main/java/org/utbot/examples/objects/HiddenFieldSuccClass.java
```diff
@@ -2,4 +2,9 @@
 
 public class HiddenFieldSuccClass extends HiddenFieldSuperClass {
     public double b;
+
+    @Override
+    public String toString() {
+        return b + ". Super b: " + super.b;
+    }
```
CaelmBleidd, line 9: Isn't it a forgotten method implementation?

utbot-framework/src/test/kotlin/org/utbot/examples/objects/HiddenFieldAccessModifiersTest.kt
```diff
@@ -0,0 +1,20 @@
...
+internal class HiddenFieldAccessModifiersTest : UtValueTestCaseChecker(testClass = HiddenFieldAccessModifiersExample::class) {
+    @Test
+    fun testCheckSuperFieldEqualsOne() {
+        check(
+            HiddenFieldAccessModifiersExample::checkSuperFieldEqualsOne,
+            eq(3),
+            { o, _ -> o == null },
+            { _, r -> r == true },
+            { _, r -> r == false},
+            coverage = DoNotCalculate
```
CaelmBleidd, line 17: Why?

CaelmBleidd (changes requested):

utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
volivan239: Added documentation for this method

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
volivan239: Previously it was a method that was finding `Field` by its name. Now, when the `FieldId` is passed as argument, it is not really finding -- it's more like checking that this field belongs to this class and returning `Field`.

Renamed to `fieldByIdOrNull` for greater clarity

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/util/IdUtil.kt
volivan239: fixed

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/util/IdUtil.kt
volivan239: fixed, although it's a little bit copy-pasted now

utbot-core/src/main/kotlin/org/utbot/common/KClassUtil.kt
volivan239: Renamed to `findDirectAccessedFieldOrNull`

utbot-framework/src/main/kotlin/org/utbot/engine/Hierarchy.kt
volivan239: Renamed to `realFieldDeclraingType`. Is "declaring type" a correct term?

utbot-framework/src/main/kotlin/org/utbot/engine/Hierarchy.kt
volivan239: fixed here

volivan239: Are there other ways how field can belong to the class? At least, similar check was for `ancestorType` before these changes

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
volivan239: Just rewrote using `declaringClass`, seems better now

utbot-framework/src/main/kotlin/org/utbot/engine/ValueConstructor.kt
volivan239: Added same method without `orNull`, fixed

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/DslUtil.kt
volivan239: fixed here

utbot-framework/src/test/kotlin/org/utbot/examples/UtModelTestCaseChecker.kt
volivan239: replaced with `declaring`

utbot-framework/src/test/kotlin/org/utbot/examples/objects/HiddenFieldAccessModifiersTest.kt
volivan239: added coverage here and to `org.utbot.examples.objects.HiddenFieldExampleTest#testCheckSuccField`

utbot-sample/src/main/java/org/utbot/examples/objects/HiddenFieldSuccClass.java
volivan239: This override really seems useless, removed it

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
Damtev: There are no functional changes, but I am a bit confused with this `TODO` removing. Are you sure it is resolved?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/DslUtil.kt
```diff
@@ -72,7 +76,11 @@ fun stringLiteral(string: String) = CgLiteral(stringClassId, string)
 
 // non-static fields
 operator fun CgExpression.get(fieldId: FieldId): CgFieldAccess =
-    CgFieldAccess(this, fieldId)
+    if (type.jClass.findDirectAccessedFieldOrNull(fieldId.name) != fieldId.field) {
+        CgFieldAccess(CgTypeCast(fieldId.declaringClass, this), fieldId)
+    } else {
+        CgFieldAccess(this, fieldId)
+    }
```
Damtev, line 83: This functionality was just a shortcut for a `CgFieldAccess` constructor. I don't think we should use more complex logic with additional checks here.

Also, this logic does not always work - there is no check that `fieldId.declaringClass` is accessible, so there is possible situation with cast to a private class.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/visitor/CgKotlinRenderer.kt
```diff
@@ -121,7 +121,9 @@ internal class CgKotlinRenderer(context: CgContext, printer: CgPrinter = CgPrint
     // Property access
 
     override fun visit(element: CgFieldAccess) {
+        if (element.caller is CgTypeCast) print("(")
         element.caller.accept(this)
+        if (element.caller is CgTypeCast) print(")")
```
Damtev, line 126: Why?

Damtev (changes requested):

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/visitor/CgKotlinRenderer.kt
volivan239: If lhs is typecast, we need brackets for correct priority. Without this, the generated code will be like `Succ as Super.f` and not `(Succ as Super).f`.

This is a little bit of a hack, but I didn't find more general solution.

Damtev: This is definitely a hack, which is also not total - it does not cover similar cases for a method invocation and an array access. We need to find another solution.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/DslUtil.kt
volivan239: What is the best place to move this logic to? Before calling this there are checks that `fieldId` is accessible, aren't they enough?

Also, I believe that this logic should solve the removed TODO: it checks if we should use obj or declaring class as a caller. But maybe I misunderstood the TODO

volivan239:
After discussion with @Damtev and @ArsenHD it was decided to merge into main only changes concerning engine (#627) so far. It was also decided to leave codegen untouched until @ArsenHD finishes his changes in codegen so we would have more rational ways of dealing with field hiding there.

volivan239:
Corresponding issue was finally solved in #763, closing this one therefore.

PR #595, opened 2022-07-26
Added a few tests to check generated clustering comments

### LINKED ISSUE #384: No tests for clusterization of tests and checking produced comments
We need to add the minimal set of tests to cover the functionality of tests clusterization with DBSCAN algorithm.

It groups tests in sets based on the JIMPLE instructions and produces comments with `//` symbols

Currently this behaviour is not fixed and described in the tests

amandelpie:
# Description

This PR proposes a checker for generated clustering comments and adds a few tests to check them on a few non-trivial situations

Fixes #384 

## Type of Change

# How Has This Been Tested?

## Automated Testing

Just run summary-tests

- [ ] No new warnings

### DIFF
--- a/utbot-summary-tests/src/test/kotlin/examples/SummaryTestCaseGeneratorTest.kt
+++ b/utbot-summary-tests/src/test/kotlin/examples/SummaryTestCaseGeneratorTest.kt
@@ -10,10 +10,7 @@ import org.utbot.framework.UtSettings.checkNpeInNestedMethods
 import org.utbot.framework.UtSettings.checkNpeInNestedNotPrivateMethods
 import org.utbot.framework.UtSettings.checkSolverTimeoutMillis
 import org.utbot.framework.codegen.TestExecution
-import org.utbot.framework.plugin.api.CodegenLanguage
-import org.utbot.framework.plugin.api.MockStrategyApi
-import org.utbot.framework.plugin.api.UtExecution
-import org.utbot.framework.plugin.api.UtMethod
+import org.utbot.framework.plugin.api.*
 import org.utbot.framework.plugin.api.util.UtContext
 import org.utbot.summary.comment.nextSynonyms
 import org.utbot.summary.summarize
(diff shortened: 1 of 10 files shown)


utbot-summary-tests/src/test/kotlin/examples/SummaryTestCaseGeneratorTest.kt
```diff
@@ -124,6 +87,32 @@ open class SummaryTestCaseGeneratorTest(
...
+    // TODO: if next synonyms and normalize function will be removed, this method could be moved as overridden equals to the dataClass [UtClusterInfo]
+    private fun UtClusterInfo.normalizedAndEquals(other: UtClusterInfo): Boolean {
+        if (header != other.header) return false
+
+        return if (content == null) {
+            other.content == null
+        } else {
+            if (other.content == null) false
+            else {
+                content!!.normalize() == other.content!!.normalize()
```
onewhl, line 99: do we need null safety operator  `!!` here? it seems both `content` and `other.content` couldn't be null here

utbot-summary-tests/src/test/kotlin/examples/SummaryTestCaseGeneratorTest.kt
```diff
@@ -124,6 +87,32 @@ open class SummaryTestCaseGeneratorTest(
...
+            }
+        }
+    }
+
```
onewhl, line 103: could you please add a comment to clarify what exactly this method checks?

utbot-summary-tests/src/test/kotlin/math/SummaryIntMathTest.kt
```diff
@@ -135,10 +136,14 @@ class SummaryIntMathTest : SummaryTestCaseGeneratorTest(
             methodName14
         )
 
+        val clusterInfo = listOf(
+            Pair(UtClusterInfo("SUCCESSFUL EXECUTIONS for method pow(int, int)", null), IntRange(0, 13))
```
onewhl, line 140: if a cluster content is a range of numbers, why don't we check just it's amount, not a range of numbers? 
i understand that exact numbers might be important, but it seems in tests we check only its amount, not exact values

utbot-summary-tests/src/test/kotlin/examples/SummaryTestCaseGeneratorTest.kt
```diff
@@ -124,6 +87,32 @@ open class SummaryTestCaseGeneratorTest(
...
+    fun UtMethodTestSet.checkClusterInfo(clusterInfo: List<Pair<UtClusterInfo, IntRange>>) {
```
onewhl, line 104: could you please add a comment to clarify what exactly this method checks?

utbot-summary-tests/src/test/kotlin/math/SummaryIntMathTest.kt
amandelpie: Ok, let's check just numbers

utbot-summary-tests/src/test/kotlin/examples/SummaryTestCaseGeneratorTest.kt
amandelpie: This is a message from IDEA: "Smart cast to 'String' is impossible, because 'other.content' is a public API property declared in different module"

amandelpie: But we could do !! especially because we could guarantee that they are not nulls:)

amandelpie:
@onewhl please have a look

onewhl (approved):
Cool, thank you!

PR #600, opened 2022-07-27
Make the `peek` method behaviour in the path selector deterministic #598

### LINKED ISSUE #598: Flaky behavior of the tests depending on visualization option
**Description**

Several tests for wrappers fail (or pass) the pipeline depending on the visualisation option value in UtSettings file. If we have this option enabled, everything works fine. If we disable it -- several tests fail.

**To Reproduce**

Steps to reproduce the behaviour:

It is flaky behaviour and there is no common way to reproduce it. Probably, it depends on the state of the system.

There are no specific visual proofs. 
For example, `org.utbot.examples.collections.MapEntrySetTest#testIteratorNext` fails with absent second branch that leads to the `NoSuchElementException`.

**Environment**

No specific environment is required.

**Additional context**

I've investigated the problem recently. It happens because of the non-deterministic behaviour of the method `peekImpl` in `NonUniformRandomSearch` method. 

When we have enabled visualization, it calls `peek` several times before each `poll` call to visualize the graph, which changes the next chosen state (because of the random usage in the method). At the same time, if we disable visualization, there will be no `peek` calls, and, therefore, we might have a different sequence of the chosen states from the execution queue.

CaelmBleidd:
# Description

Make a behaviour of the `peek` method of the path selector determenistic. To do it we fix taken random value between `poll` calls, so any sequence of the `peek` methods (and one following `poll` call) will take the same element of the queue.

Fixes #598

## Type of Change

# How Has This Been Tested?

## Automated Testing

Any test for collections that try to find `NoSuchElementException`.

## Manual Scenario 

You can set a debug point at any place where there are several elements in the path selector's queue, and call `peek` method multiple times. Check that it doesn't calculate a random value every time and returns the same value all the time.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/selectors/nurs/NonUniformRandomSearch.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/selectors/nurs/NonUniformRandomSearch.kt
@@ -58,6 +58,13 @@ abstract class NonUniformRandomSearch(
 
     private val randomGen: Random? = seed?.let { Random(seed) }
 
+    // We use this value to avoid non-deterministic behaviour of the
+    // peek method. Without it, we might have different states for
+    // a sequence of the `peek` calls.
+    // Now we remember it at the first `peek` call and use it
+    // until a first `poll` call. The first call should reset it back to null.
+    private var lastTakenRandomValue: Double? = null
+
     override fun update() {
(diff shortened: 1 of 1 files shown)


volivan239 (approved):
Fixes #598 for me.

dtim (approved):
LGTM

PR #602, opened 2022-07-28
An attempt to avoid UtMethod in codegen

EgorkaKulikov:
# Description

Using `UtMethod<*>` makes codegen support for Javascript, Python and Go much more difficult.
We try to create a facade `CgMethodTestSet` for `UtMethodTestSet` to use it in all codegen API. This approach allows to reuse our codegen implementation.

Fixes # ([621](https://github.com/UnitTestBot/UTBotJava/issues/621)) as a part of refactoring.

## Type of Change

# How Has This Been Tested?

## Automated Testing

`utbot-samples` pipleline

## Manual Scenario 

regression checks on parametrized tests, including static method under test


ArsenHD (approved):
I added a couple minor changes. Everything is looking good.

PR #625, opened 2022-08-01
Add SecurityManager support to block suspicious code #622

### LINKED ISSUE #622: Add SecurityManager support to block suspicious code
**Description**

Fuzzer doesn't analyze the code it runs for instrumentation. It is needed to sandbox this code to block suspicious code from running.

**Expected behavior**

Fuzzer shouldn't be able to run code without the appropriate permissions.

**Environment**

This restrictions should work for every supported environment.

**Potential alternatives**

The running can be instrumented or to be contained as well.

Markoutte:
# Description

This commit introduces support of SecurityManager in instrumentation.jar to restrict all suspicious code from running (File IO, Socket IO, Property reading and etc.). Security is enabled by default and limits all permissions to a small set that needed for execution. Settings can be extended by editing `.utbot/sandbox.policy` ([more about policy file and syntax](https://docs.oracle.com/javase/7/docs/technotes/guides/security/PolicyFiles.html#Examples)).

Fixes #622

## Type of Change

# How Has This Been Tested?

## Automated Testing

All existing tests should pass

## Manual Scenario 

Disable symbolic execution test generation and leave only fuzzing (in the settings by choosing "Simpler" for "Code analysis"). Run test generations for these methods:

```java
public class SecurityCheck {

    public int normalTest(int value) {
        if (value < 0) {
            return -value;
        }
        return value;
    }

    public int read(File path) throws IOException {
        byte[] bytes = Files.readAllBytes(path.toPath());
        return bytes.length;
    }

    public int connect(Socket socket) throws IOException {
        socket.connect(new InetSocketAddress("0.0.0.0", 22));
        return 0;
    }

    public String property(String key) {
        return System.getProperty(key);
    }

    public String systemExit() {
        System.exit(0);
        return "bad";
    }

}
```

For every method except `normatTest()` there are should be several methods that look like:

```java
    public void testPropertyWithBlankString() {
        SecurityCheck securityCheck = new SecurityCheck();
        
        /* This test fails because method [com.company.security.SecurityCheck.property] produces [java.security.AccessControlException: access denied ("java.util.PropertyPermission" "   " "read")]
            java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
            java.security.AccessController.checkPermission(AccessController.java:886)
            java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
            java.lang.SecurityManager.checkPropertyAccess(SecurityManager.java:1294)
            java.lang.System.getProperty(System.java:719)
            com.company.security.SecurityCheck.property(SecurityCheck.java:32) */
    }
```

Try to edit `{user.home}/.utbot/sandbox.policy`:

```
grant {
    permission java.util.PropertyPermission "*", "read";
};
```

and generate test for `String property(String key)`. Tests should be generated without access exception.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/UtExecutionResult.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/UtExecutionResult.kt
@@ -17,6 +17,10 @@ data class UtOverflowFailure(
     override val exception: Throwable,
 ) : UtExecutionFailure()
 
+data class UtSandboxFailure(
+    override val exception: Throwable
+) : UtExecutionFailure()
+
 /**
  * unexpectedFail (when exceptions such as NPE, IOBE, etc. appear, but not thrown by a user, applies both for function under test and nested calls )
  * expectedCheckedThrow (when function under test or nested call explicitly says that checked exception could be thrown and throws it)
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/Resolver.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/Resolver.kt
@@ -72,6 +72,7 @@ import kotlin.math.min
 import kotlinx.collections.immutable.persistentListOf
 import kotlinx.collections.immutable.persistentSetOf
 import org.utbot.framework.plugin.api.SYMBOLIC_NULL_ADDR
+import org.utbot.framework.plugin.api.UtSandboxFailure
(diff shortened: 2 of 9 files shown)


denis-fokin:
Could we add some tests?

EgorkaKulikov:
I suggest to improve text message "Disabled due to sandbox" and have no other complaints from the point of view of Codegen

denis-fokin (approved):

PR #640, opened 2022-08-01
Replace currentExecutableId with currentTestSet

EgorkaKulikov:
# Description

An entity `currentExecutable` is removed from `CgContext` and is replaced with `currentTestSet`.
We use this test set everywhere and avoid passing it as a parameter. All null-safety checks are provided explicitly.

## Type of Change

# How Has This Been Tested?

Full automatic and manual codegen regression

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/name/CgNameGenerator.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/name/CgNameGenerator.kt
@@ -44,7 +44,7 @@ internal interface CgNameGenerator {
     /**
      * Generate a new test method name.
      */
-    fun testMethodNameFor(executableId: ExecutableId, customName: String? = null): String
+    fun testMethodNameFor(customName: String? = null): String
 
     /**
      * Generates a new parameterized test method name by data provider method name.
@@ -54,7 +54,7 @@ internal interface CgNameGenerator {
     /**
      * Generates a new data for parameterized test provider method name
      */
-    fun dataProviderMethodNameFor(executableId: ExecutableId): String
+    fun dataProviderMethodNameFor(): String
 
     /**
      * Generate a new error method name
@@ -85,8 +85,10 @@ internal class CgNameGeneratorImpl(private val context: CgContext)
         return variableName(baseName.decapitalize(), isMock)
     }
 
-    override fun testMethodNameFor(executableId: ExecutableId, customName: String?): String {
-        val executableName = createExecutableName(executableId)
+    override fun testMethodNameFor(customName: String?): String {
+        val executable = currentTestSet?.executableId
+            ?: error("CurrentTestSet must be initialized to create test method name")
+        val executableName = createExecutableName(executable)
 
         // no index suffix allowed only when there's a vacant custom name
         val name = if (customName != null && customName !in existingMethodNames) {
@@ -104,8 +106,10 @@ internal class CgNameGeneratorImpl(private val context: CgContext)
     override fun parameterizedTestMethodName(dataProviderMethodName: String) =
         dataProviderMethodName.replace(dataProviderMethodPrefix, "parameterizedTestsFor")
 
-    override fun dataProviderMethodNameFor(executableId: ExecutableId): String {
-        val executableName = createExecutableName(executableId)
+    override fun dataProviderMethodNameFor(): String {
+        val executable = currentTestSet?.executableId
+            ?: error("CurrentTestSet must be initialized to create name for data provider method")
+        val executableName = createExecutableName(executable)
         val indexedName = nextIndexedMethodName(executableName.capitalize(), skipOne = true)
 
         existingMethodNames += indexedName
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgMethodConstructor.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgMethodConstructor.kt
@@ -1,7 +1,6 @@
 package org.utbot.framework.codegen.model.constructor.tree
 
 import org.utbot.common.PathUtil
-import org.utbot.common.packageName
 import org.utbot.framework.assemble.assemble
 import org.utbot.framework.codegen.ForceStaticMocking
 import org.utbot.framework.codegen.Junit4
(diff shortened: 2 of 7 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/name/CgNameGenerator.kt
```diff
@@ -44,7 +44,7 @@ internal interface CgNameGenerator {
     /**
      * Generate a new test method name.
      */
-    fun testMethodNameFor(executableId: ExecutableId, customName: String? = null): String
+    fun testMethodNameFor(customName: String? = null): String
```
ArsenHD, line 47: I think we shouldn't remove arguments for current executable. For example, here we hava a `CgNameGenerator`, which is a separate component that could potentially generate names for different executions (even though now we don't do such thing). So I think we shouldn't omit an argument here. Moreover, the name of the method now makes less sense (`testMethodNameFor` - for what?).

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/name/CgNameGenerator.kt
```diff
@@ -104,8 +106,10 @@ internal class CgNameGeneratorImpl(private val context: CgContext)
     override fun parameterizedTestMethodName(dataProviderMethodName: String) =
         dataProviderMethodName.replace(dataProviderMethodPrefix, "parameterizedTestsFor")
 
-    override fun dataProviderMethodNameFor(executableId: ExecutableId): String {
-        val executableName = createExecutableName(executableId)
+    override fun dataProviderMethodNameFor(): String {
```
ArsenHD, line 109: Here I think the same as what I pointed out about `testMethodNameFor()`:
- `CgNameGenerator` should be just a service that is able to generate names for anything, not just current execution
- `For` in the name now has no meaning

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgMethodConstructor.kt
```diff
@@ -1088,11 +1093,11 @@ internal class CgMethodConstructor(val context: CgContext) : CgContextOwner by c
...
     }
 
-    fun createTestMethod(executableId: ExecutableId, execution: UtExecution): CgTestMethod =
+    fun createTestMethod(execution: UtExecution): CgTestMethod =
         withTestMethodScope(execution) {
-            val testMethodName = nameGenerator.testMethodNameFor(executableId, execution.testMethodName)
+            val testMethodName = nameGenerator.testMethodNameFor(execution.testMethodName)
             // TODO: remove this line when SAT-1273 is completed
-            execution.displayName = execution.displayName?.let { "${executableId.name}: $it" }
+            execution.displayName = execution.displayName?.let { "${currentTestSet?.executableId?.name}: $it" }
```
ArsenHD, line 1100: Here in case if `currentTestSet == null` instead of an executable's name there will be just `null`. I advise storing this `executableId` in a variable with `?: error()` check like you did in other places.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgMethodConstructor.kt
```diff
@@ -1567,8 +1572,8 @@ internal class CgMethodConstructor(val context: CgContext) : CgContextOwner by c
             arguments = arrayOf(objectArrayClassId)
         )
 
-    private fun containsFailureExecution(testSet: CgMethodTestSet) =
-        testSet.executions.any { it.result is UtExecutionFailure }
+    private fun containsFailureExecution(): Boolean =
```
ArsenHD, line 1575: I don't think we should drop an argument here. It makes this function less common. Before we could pass any `CgMethodTestSet` here and now it only works with the current one. Also, now the name of the method is harder to understand. Perhaps it would even be better to make this an extension? Like `fun CgMethodTestSet.containsFailureExecution()`?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgTestClassConstructor.kt
```diff
@@ -42,6 +42,25 @@ internal class CgTestClassConstructor(val context: CgContext) :
 
     private val testsGenerationReport: TestsGenerationReport = TestsGenerationReport()
 
+    private fun <R> withTestSetScope(executableTestSet: CgMethodTestSet, block: () -> R): R {
+        /**
+         * This function makes sure that **test set** currently being generated is reset.
+         * It is used at the start of test set generation and right after it.
+         */
+        fun clearTestSetScope() {
```
ArsenHD, line 50: I would recommend making it a private function of `CgTestClassConstructor`, or even setting `currentTestSet = null` twice manually (without this function). It's ok, because we only need this here at `clearTestSetScope()`.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgTestClassConstructor.kt
```diff
@@ -78,64 +101,61 @@ internal class CgTestClassConstructor(val context: CgContext) :
         }
     }
 
-    private fun construct(testSet: CgMethodTestSet): List<CgRegion<CgMethod>>? {
-        if (testSet.executions.isEmpty()) {
-            return null
-        }
+    private fun construct(): List<CgRegion<CgMethod>> {
```
ArsenHD, line 104: It is now more difficult to understand what to construct due to the absence of arguments. I suggest returning an argument and maybe even renaming the `construct` method to something like `constructTestMethods`.

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/generator/UtTestsDialogProcessor.kt
```diff
@@ -203,7 +203,7 @@ object UtTestsDialogProcessor {
                                                 generate = testFlow {
                                                     generationTimeout = model.timeout
                                                     isSymbolicEngineEnabled = true
-                                                    isFuzzingEnabled = UtSettings.useFuzzing
+                                                    isFuzzingEnabled = false
```
ArsenHD, line 206: Why?

EgorkaKulikov:
It was an unsuccessful attempt to suggest a refactoring. Closed as useless.

PR #670, opened 2022-08-04
Ability to create and run tests at once #156

### LINKED ISSUE #156: Ability to create and run tests at once
**Description**

At the moment tests generation and tests running are separate activities and user should define the scope of newly creates tests manually. It's not user-friendly.

**Expected behavior**

There should be an option to Run generated tests after generation (e.g. Run button or 'Run generated tests' checkbox in the dialog)

**Environment**

No specific environment

**Potential alternatives**

Create and select corresponding run configuration for generated test scope but not run.

**Context**

'Generate tests with UtBot' dialog window

Vassiliy-Kudryashov:
# Description

- We create run configuration (type = PATTERN) that runs with coverage all generated tests at once
- Option button replaces standard OK button in the dialog, user can choose what to do
- In case we cannot generate any test for some class in batch mode we won't show tons of "Failed to generate..." message dialogs.
- Useful run{TARGET, runnable} API was moved to IntelliJApiHelper that replaced small existing AndroidApiHelper
- New field named "runGeneratedTestsWithCoverage" was introduced for both Settings and GeneratetestsModel to persist user's choice

Fixes #156

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Choose some example project, open generation dialog. Now it's possible to choose "Generate & Run" option and then call it.
As a result there should appear new (temporary) Run Configuration for the generated test scope, it starts automatically with coverage (IDEA built-in feature)

- [x] No new warnings

### DIFF
--- a/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/RunConfigurationHelper.kt
+++ b/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/RunConfigurationHelper.kt
@@ -8,6 +8,7 @@ import com.intellij.execution.RunManagerEx
 import com.intellij.execution.actions.ConfigurationContext
 import com.intellij.execution.executors.DefaultRunExecutor
 import com.intellij.execution.runners.ExecutionUtil
+import com.intellij.execution.runners.ProgramRunner
 import com.intellij.openapi.actionSystem.DataContext
 import com.intellij.openapi.actionSystem.LangDataKeys
 import com.intellij.openapi.application.ApplicationManager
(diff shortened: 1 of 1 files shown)


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/RunConfigurationHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+                    val mapDataContext = MyMapDataContext().also {
+                        it.put(LangDataKeys.PSI_ELEMENT_ARRAY, testClasses.toTypedArray())
+                        it.put(LangDataKeys.MODULE, model.testModule)
+                        it.put(Location.DATA_KEYS, locations)
+                        it.put(Location.DATA_KEY, locations[0])
+                    }
+                    val myConfigurationContext = try {
+                        MyConfigurationContext(mapDataContext, testClasses[0])
+                    } catch (e: Exception) {
+                        e.printStackTrace()
```
Markoutte, line 62: It is more preferable to use logger here instead of printStackTrace

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/RunConfigurationHelper.kt
```diff
@@ -0,0 +1,95 @@
...
+                                settings,
+                                ExecutorRegistry.getInstance().getExecutorById(CoverageExecutor.EXECUTOR_ID)
+                                    ?: DefaultRunExecutor.getRunExecutorInstance()
+                            )
+                            with(RunManagerEx.getInstanceEx(model.project)) {
+                                if (findSettings(settings.configuration) == null) {
+                                    settings.isTemporary = true
+                                    addConfiguration(settings)
+                                }
+                                selectedConfiguration = settings
```
Markoutte, line 86: Selected configuration is a global state and I am not sure it is good idea to change it directly. Are there any options?

PR #678, opened 2022-08-06
Enlarge the set of cases when we try to create parametrized tests

EgorkaKulikov:
# Description

`ForceMockListener` should not trigger on the force mocking related to `UtMock` or another `utbotSuperClasses` classes.
They are not mocks really, considering them as mocks is just a specific hack for symbolic engine with no impact on the code generation process.

Setting to avoid statics substitution in parametrized tests removed because they are now modified to deal with statics.

## Type of Change

# How Has This Been Tested?

## Automated Testing

Running the pipeline on `utbot-samples`


sofurihafe (approved):
LGTM.

PR #695, opened 2022-08-09
Explicitly set the child process working directory if we run the plugin

### LINKED ISSUE #694: The child process working directory differs from the test runner working directory
**Description**

When we refer to some file relatively, we strictly rely on the current working dir. The problem is that the child process working directory differs from the test runner working directory.

It's the reason of SAT-1710 (False ExceptionInInitializerError in Spring project).

**To Reproduce**

1. Open any project
2. Add an empty file with the name "definitely_exists" to the root project directory
3. Leave only fuzzing turned on
4. Generate tests for this method:
```java
    boolean checkFileExists() {
        File file = new File("./definitely_exists");
        if (file.exists()) {
            return true;
        } else {
            return false;
        }
    }
```

**Expected behavior**

```java
  public void testCheckFileExistsReturnsTrue() {
      CMCSyncService cMCSyncService = new CMCSyncService();

      boolean actual = cMCSyncService.checkFileExists();

      assertTrue(actual);
  }
```
And test passes.

**Actual behavior**

```java
  public void testCheckFileExistsReturnsTrue() {
      CMCSyncService cMCSyncService = new CMCSyncService();

      boolean actual = cMCSyncService.checkFileExists();

      assertFalse(actual);
  }
```
And test fails.

sergeypospelov:
- Now we explicitly sets the child process working directory as the project root.
- Also refactored services a bit

# Description

Substitute this text with a concise description of the proposed change. Emphasize, why particular solution was chosen.

Fixes #694

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

**To reproduce** section from #694 succeeds.

- [x] No new warnings


utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/services/JdkInfoService.kt
```diff
@@ -0,0 +1,31 @@
+package org.utbot.framework.plugin.services
+
+import java.nio.file.Path
+import java.nio.file.Paths
+
+/**
+ * Singleton to enable abstract access to path to JDK.
```
nikitavlaev, line 7: Comment is out of place now and looks a bit scuffed.
Bring back usage or delete `and in the test runs`

nikitavlaev: This comment is for the JdkInfoService

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/services/JdkInfoService.kt
```diff
@@ -0,0 +1,31 @@
...
+
+ * and in the test runs.
```
sergeypospelov, line 9: Broken comment

nikitavlaev (approved):

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/services/JdkInfoService.kt
sergeypospelov: Yeah, thank you.

PR #735, opened 2022-08-17
Fix direct assigning of fields with private type #731

### LINKED ISSUE #731: Class cast is missing for the instance of private nested class
**Description**

Class cast is missing for the Object of private nested class.

**To Reproduce**

1. IntelliJ IDEA 2022.1.4 is running with UTBot plugin installed
2. Open a project with JDK 8/11
3. Add the following class to it:

```java
public class ABC {

    private class NML {
        public NML[] array = new NML[2];

        NML() {
            array[0] = null;
            array[1] = null;
        }
    }

    NML obj;

    ABC() {
        obj = new NML();
    }

    public int size() {
        return obj.array.length;
    }

}
```

4. Open the generated test

**Expected behavior**

Generated tests are supposed to be compilable.
Class cast must be added to the instance of the nested class, like 

```java
ABC.NML nml = ((ABC.NML) createInstance("ABC$NML"));
```

**Actual behavior**

Class cast is missing for the NML Object created by means of reflection.

The following test is present:

~~~java
public void testSize_ThrowNullPointerException_1() throws Exception  {
        ABC abc = new ABC();
        Object nml = createInstance("ABC$NML");
        setField(nml, "array", null);
        abc.obj = nml;
        
        /* This test fails because method [ABC.size] produces [java.lang.NullPointerException]
            ABC.size(ABC.java:19) */
        abc.size();
    }
~~~

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.1.4
JDK 8

**Additional context**

If private modifier is removed  - nested class is package-private or public - the class cast is made correctly.

volivan239:
# Description

If we have a public field with type that is not accessible in testclass, then we should use reflection but not direct assignments to set it (because rhs will likely be declared as `Object`).

Fixes #731 

## Type of Change

# How Has This Been Tested?

## Automated Testing

Added `org.utbot.examples.codegen.ClassWithStaticAndInnerClassesTest#testGetValueFromPublicFieldWithPrivateType`

## Manual Scenario 

Launch plugin on example from #731 -- works as expected. 

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/test/kotlin/org/utbot/examples/codegen/ClassWithStaticAndInnerClassesTest.kt
+++ b/utbot-framework/src/test/kotlin/org/utbot/examples/codegen/ClassWithStaticAndInnerClassesTest.kt
@@ -98,6 +98,15 @@ internal class ClassWithStaticAndInnerClassesTest : UtValueTestCaseChecker(testC
     }
 
     @Test
+    fun testGetValueFromPublicFieldWithPrivateType() {
+        check(
+            ClassWithStaticAndInnerClasses::getValueFromPublicFieldWithPrivateType,
+            eq(2),
+            coverage = DoNotCalculate
+        )
+    }
+
+    @Test
     fun testPublicStaticClassWithPrivateField_DeepNestedStatic_g() {
         checkAllCombinations(
             ClassWithStaticAndInnerClasses.PublicStaticClassWithPrivateField.DeepNestedStatic::g,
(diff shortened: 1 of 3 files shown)


utbot-framework/src/test/kotlin/org/utbot/examples/codegen/ClassWithStaticAndInnerClassesTest.kt
```diff
@@ -97,6 +97,15 @@ internal class ClassWithStaticAndInnerClassesTest : UtValueTestCaseChecker(testC
         )
     }
 
+    @Test
+    fun testGetValueFromPublicFieldWithPrivateType() {
+        check(
+            ClassWithStaticAndInnerClasses::getValueFromPublicFieldWithPrivateType,
+            eq(2),
+            coverage = DoNotCalculate
```
dtim, line 105: JFYI: it would be better to use `coverage = DoNotCalculate` only to avoid false positive test failure when coverage computation returns wrong results (it seems that it is the case here) although the set of executions is correct. It would also be useful to add matchers to be sure that executions of interest have been really found.

I think this test can be left as is, as it only provides data to the integration test (the only way to check the codegen) and follows the style of existing examples. Nevertheless, adding matchers could help to detect engine bugs as well.

dtim (approved):
LGTM

PR #755, opened 2022-08-22
Enhance fuzzing for arrays #738

### LINKED ISSUE #738: Enhance fuzzer's ArrayModelProvider
**Description**

Currently, `ArrayModelProvider` generates only empty array and an array with 10 default elements as models. There are many possibilities of generating more meaningful and interesting models.

**Expected behavior**

We can construct small arrays and fill them with fuzzed values. We can obtain their sizes from `ConstantsFinder`.

**Context**

Same ideas could be used in different collections, but currently it is difficult to support them due to type erasing.

volivan239:
# Description

Now, `ArrayModelProvider` generates more interesting tests:

* each element of array is generated recursively by fuzzing synthetic method.
* arrays with "interesting" small sizes are added ("interesting" means that this number is obtained from source code)

Fixes #738

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Generate tests for methods from `ArrayExample` class (using only fuzzer):

```Java
class Point {
    int x, y;

    public Point(int x, int y) {
        this.x = x;
        this.y = y;
    }

    boolean equals(Point other) {
        return (other.x == this.x && other.y == this.y);
    }
}

public class ArrayExample {
    boolean checkAllSame(Integer[] a) {    // Fuzzer should find parameters giving true as well as parameters giving false
        Set<Integer> s = new HashSet<>(Arrays.asList(a));
        return (s.size() <= 1);
    }

    public boolean checkAllSamePoints(Point[] a) { // Also works for classes
        if (a.length == 4) {
            return false; // Test with array of size 4 should be generated by fuzzer
        }
        for (int i = 1; i < a.length; i++) {
            if (!a[i].equals(a[i - 1]))
                return false;
        }
        return true;
    }

    public boolean checkRowsWithAllSame2D(int[][] a, int y) {
        int cntSame = 0;
        for (int i = 0; i < a.length; i++) {
            boolean same = true;
            for (int j = 1; j < a[i].length; j++) {
                if (a[i][j] != a[i][j - 1]) {
                    same = false;
                    break;
                }
            }
            if (same)
                cntSame++;
        }
        return (cntSame == y && y > 0 && y < a.length);
    }
}
```

For each of them, meaningful tests are generated that cover both true/false outputs.

- [ ] No new warnings

### DIFF
--- a/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/Fuzzer.kt
+++ b/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/Fuzzer.kt
@@ -12,6 +12,7 @@ import org.utbot.fuzzer.providers.EnumModelProvider
 import org.utbot.fuzzer.providers.ObjectModelProvider
 import org.utbot.fuzzer.providers.PrimitiveDefaultsModelProvider
 import org.utbot.fuzzer.providers.PrimitiveWrapperModelProvider
+import org.utbot.fuzzer.providers.PrimitivesModelProvider
 import org.utbot.fuzzer.providers.RegexModelProvider
 import org.utbot.fuzzer.providers.StringConstantModelProvider
 import java.util.*
@@ -145,39 +146,44 @@ fun <T> Sequence<List<FuzzedValue>>.withMutations(statistics: FuzzerStatistics<T
 }
 
 /**
- * Creates a model provider consisting of providers that do not make recursive calls inside them
+ * Creates a model provider from a list of default providers.
  */
-private fun nonRecursiveProviders(idGenerator: IdentityPreservingIdGenerator<Int>): ModelProvider {
+fun defaultModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>): ModelProvider {
     return ModelProvider.of(
+        ObjectModelProvider(idGenerator),
         CollectionModelProvider(idGenerator),
+        ArrayModelProvider(idGenerator),
         EnumModelProvider(idGenerator),
+        ConstantsModelProvider,
         StringConstantModelProvider,
         RegexModelProvider,
         CharToStringModelProvider,
-        ConstantsModelProvider,
-        PrimitiveDefaultsModelProvider,
+        PrimitivesModelProvider,
         PrimitiveWrapperModelProvider,
     )
 }
 
 /**
- * TODO: write doc here
+ * Creates a model provider from a list of providers that we want to use by default in [RecursiveModelProvider]
  */
-private fun recursiveModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>, recursionDepth: Int): ModelProvider {
-    return ModelProvider.of(
-        ObjectModelProvider(idGenerator, recursionDepth),
-        ArrayModelProvider(idGenerator, recursionDepth)
+internal fun modelProviderForRecursiveCalls(idGenerator: IdentityPreservingIdGenerator<Int>, recursionDepth: Int): ModelProvider {
+    val nonRecursiveProviders = ModelProvider.of(
+        CollectionModelProvider(idGenerator),
+        EnumModelProvider(idGenerator),
+        StringConstantModelProvider,
+        CharToStringModelProvider,
+        ConstantsModelProvider,
+        PrimitiveDefaultsModelProvider,
+        PrimitiveWrapperModelProvider,
     )
-}
 
-/**
- * Creates a model provider from a list of default providers.
- */
-fun defaultModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>, recursionDepth: Int = 1): ModelProvider =
-    if (recursionDepth >= 0)
-        nonRecursiveProviders(idGenerator).with(recursiveModelProviders(idGenerator, recursionDepth))
+    return if (recursionDepth >= 0)
+        nonRecursiveProviders
+            .with(ObjectModelProvider(idGenerator, recursionDepth))
+            .with(ArrayModelProvider(idGenerator, recursionDepth))
     else
-        nonRecursiveProviders(idGenerator)
+        nonRecursiveProviders
+}
 
 
 fun defaultModelMutators(): List<ModelMutator> = listOf(
--- a/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/Fuzzer.kt
+++ b/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/Fuzzer.kt
@@ -166,7 +166,7 @@ fun defaultModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>): Mode
 /**
  * Creates a model provider consisting of providers that do not make recursive calls inside them
  */
-fun nonRecursiveProviders(idGenerator: IdentityPreservingIdGenerator<Int>): ModelProvider {
+private fun nonRecursiveProviders(idGenerator: IdentityPreservingIdGenerator<Int>): ModelProvider {
     return ModelProvider.of(
         CollectionModelProvider(idGenerator),
         EnumModelProvider(idGenerator),
@@ -180,14 +180,24 @@ fun nonRecursiveProviders(idGenerator: IdentityPreservingIdGenerator<Int>): Mode
 }
 
 /**
- * Creates a model provider consisting of providers that will make no more than [recursion] nested recursive calls.
+ * TODO: write doc here
  */
-fun recursiveModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>, recursion: Int): ModelProvider {
+private fun recursiveModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>, recursionDepth: Int): ModelProvider {
     return ModelProvider.of(
-        ObjectModelProvider(idGenerator, recursion),
-        ArrayModelProvider(idGenerator, recursion)
+        ObjectModelProvider(idGenerator, recursionDepth),
+        ArrayModelProvider(idGenerator, recursionDepth)
     )
 }
+
+/**
+ * Creates a model provider from a list of default providers.
+ */
+fun defaultModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>, recursionDepth: Int = 1): ModelProvider =
+    if (recursionDepth >= 0)
+        nonRecursiveProviders(idGenerator).with(recursiveModelProviders(idGenerator, recursionDepth))
+    else
+        nonRecursiveProviders(idGenerator)
+
 
 fun defaultModelMutators(): List<ModelMutator> = listOf(
     StringRandomMutator,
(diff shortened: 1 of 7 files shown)


utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ObjectModelProvider.kt
```diff
@@ -39,25 +35,20 @@ private val logger by lazy { KotlinLogging.logger {} }
...
-        get() {
-            val modelProviderWithoutRecursion = modelProvider.exceptIsInstance<ObjectModelProvider>()
-            return if (recursion > 0) {
-                ObjectModelProvider(idGenerator, limit = 1, recursion - 1).with(modelProviderWithoutRecursion)
-            } else {
-                modelProviderWithoutRecursion.withFallback(NullModelProvider)
-            }
+    private val limit: Int =
+        when(recursion) {
+            1 -> Int.MAX_VALUE
```
Markoutte, line 50: The logic looks different: before when recursion is 1, 2, 3 and so on then limit is 1. Now it is harder to understand what's happening. The idea behind this was that the limit is articulated as: when we find constructors that accepts another objects then we construct them, but use no more than one available constructor. If recursion is 0, than we pass null to every object parameter in constructor. But usually, we look for constructors with primitive values.

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.utbot.fuzzer.nonRecursiveProviders
+import org.utbot.fuzzer.recursiveModelProviders
+
+// TODO: maybe use `recursionDepth` instead of `recursion` here and in [recursiveModelProviders]?
+abstract class RecursiveModelProvider(
+    protected val idGenerator: IdentityPreservingIdGenerator<Int>,
+    protected val recursion: Int
+): ModelProvider {
+    // TODO: currently it is var due to tests, maybe we can make it private val?
+    var modelProviderForRecursiveCalls: ModelProvider =
```
Markoutte, line 19: Looks like we have lost an ability to change model providers for ObjectModelProvider, because before the changed model provider is passed into recursive call. Now, it uses only default providers.

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ArrayModelProvider.kt
```diff
@@ -1,34 +1,87 @@
...
+
+                // Fuzz arrays with interesting lengths and default-valued elements
+                val lengths = generateArrayLengths(description)
+                yieldAllValues(indices, lengths.asSequence().map { length ->
+                    createFuzzedArrayModel(arrayClassId, length, null)
                 })
             }
     }
+
+    private fun generateArrayLengths(description: FuzzedMethodDescription): Set<Int> {
```
Markoutte, line 46: Looks like this can be simplified. It is too much work to generate some values in interval from 0 up to 10

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/ObjectModelProvider.kt
```diff
@@ -39,25 +35,20 @@ private val logger by lazy { KotlinLogging.logger {} }
...
-    private val idGenerator: IdentityPreservingIdGenerator<Int>,
-    private val limit: Int = Int.MAX_VALUE,
-    private val recursion: Int = 1
-) : ModelProvider {
+    idGenerator: IdentityPreservingIdGenerator<Int>,
+    recursion: Int = 1,
+) : RecursiveModelProvider(idGenerator, recursion) {
 
-    var modelProvider: ModelProvider = objectModelProviders(idGenerator)
+    // TODO: can we make it private val (maybe depending on recursion)?
```
Markoutte, line 42: It is manual option and a part of public API

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.utbot.fuzzer.fuzz
+import org.utbot.fuzzer.nonRecursiveProviders
+import org.utbot.fuzzer.recursiveModelProviders
+
+// TODO: maybe use `recursionDepth` instead of `recursion` here and in [recursiveModelProviders]?
+abstract class RecursiveModelProvider(
+    protected val idGenerator: IdentityPreservingIdGenerator<Int>,
+    protected val recursion: Int
+): ModelProvider {
+    // TODO: currently it is var due to tests, maybe we can make it private val?
```
Markoutte, line 18: It was a part of public API to change a set of model providers. Currently, it looks broken

Markoutte (changes requested):
Please, look at comments

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.utbot.fuzzer.IdentityPreservingIdGenerator
+import org.utbot.fuzzer.ModelProvider
+import org.utbot.fuzzer.ModelProvider.Companion.yieldAllValues
+import org.utbot.fuzzer.defaultModelProviders
+import org.utbot.fuzzer.exceptIsInstance
+import org.utbot.fuzzer.fuzz
+
+data class ModelConstructor(
+    val neededTypes: List<ClassId>,
+    val createModel: (List<FuzzedValue>) -> FuzzedValue
```
Markoutte, line 17: You can use notation like this:

```kotlin
 val createModel: (inputs: List<FuzzedValue>) -> FuzzedValue
```
to clarify name for parameter. Also, doc will be very useful here

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
```diff
@@ -0,0 +1,84 @@
...
+    var totalLimit: Int = 1000
+
+    var branchingLimit: Int = 10 //Int.MAX_VALUE
+
+    private fun getModelProvider(numOfBranches: Int): ModelProvider =
+        if (recursionDepthLeft > 0)
+            modelProviderForRecursiveCalls.map {
+                if (it is RecursiveModelProvider)
+                    it.copy(idGenerator, recursionDepthLeft - 1).apply {
+                        modelProviderForRecursiveCalls = this@RecursiveModelProvider.modelProviderForRecursiveCalls
```
Markoutte, line 37: Why don't we copy all these values in the `copy` method? It looks better as we want to get a copy of this provider. Here we can leave values that should be changed like totalLimit (but in copy it should be copied, I think)

utbot-fuzzers/src/test/kotlin/org/utbot/framework/plugin/api/ModelProviderTest.kt
```diff
@@ -412,6 +413,7 @@ class ModelProviderTest {
         }
     }
 
+    @Disabled
```
Markoutte, line 416: Test checks that Outer class is created using only one constructor from the inner and it is the simplest one. I think we should keep this behavior for ObjectModelProvider

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.utbot.fuzzer.defaultModelProviders
+import org.utbot.fuzzer.exceptIsInstance
+import org.utbot.fuzzer.fuzz
+
+data class ModelConstructor(
+    val neededTypes: List<ClassId>,
+    val createModel: (List<FuzzedValue>) -> FuzzedValue
+)
+
+abstract class RecursiveModelProvider(
```
Markoutte, line 20: Please, add docs

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/Fuzzer.kt
```diff
@@ -180,6 +179,26 @@ fun objectModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>): Model
...
+        ArrayModelProvider(idGenerator, recursionDepth)
+    )
+}
+
+/**
+ * Creates a model provider from a list of default providers.
+ */
+fun defaultModelProviders(idGenerator: IdentityPreservingIdGenerator<Int>, recursionDepth: Int = 1): ModelProvider =
+    if (recursionDepth >= 0)
+        nonRecursiveProviders(idGenerator).with(recursiveModelProviders(idGenerator, recursionDepth))
```
Markoutte, line 197: Note, that the other method called 'defaultModelProviders' without parameter with recursion has another set of model providers. Therefore. I'd recommend to change name for this method to avoid ambiguous calls.

Markoutte (changes requested):
Please, see comments

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
volivan239: Added name for parameters. Also, added docs here and in other places where needed

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
volivan239: Renamed to `createNewInstance` for more clarity

utbot-fuzzers/src/test/kotlin/org/utbot/framework/plugin/api/ModelProviderTest.kt
volivan239: Returned this behavior back

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/Fuzzer.kt
volivan239: This method should have replaced the method without `recursionDepth` parameter with no changes in behavior when this parameter is not specified. Seems like I accidently got duplication after rebase. Deleted method without parameter now. Could you please check that it didn't change the behavior?

Markoutte: It changes because `PrimitivesModelProvider` isn't used anymore. By default I want to keep an ability to set different providers by default, because I usually want to use shallower value set for it. Right now this change is only about PrimitivesModelProvider, but in the future I'd want to add some providers that are used for generating flat parameters and aren't used for recursive.

Markoutte (changes requested):
Looks good but let's discuss changes about `defaultModelProviders`

Markoutte (approved):

PR #766, opened 2022-08-23
Added the ClassId for ParametrizedType

EgorkaKulikov:
# Description

After upgrading the versions of JDK, Kotlin and Gradle we have a regression in parametrized tests generation

In method `createParametersDeclaration` we need to define the types of test method parameters.
If parameter type is `ParametrizedType`, we need to be able to convert in to `ClassId`. Such logic is added.

## Type of Change

# How Has This Been Tested?

## Automated Testing

Run  org.utbot.examples.algorithms.CorrectBracketSequencesTest or some other tests containing generics in PARAMETRIZED mode.


dtim (approved):
LGTM

Issue #549, opened 2022-07-19
UI. Gradle project. Suggest all Test sources roots for test generation


alisevych:
**Description**

In Gradle projects test sources root can be located anywhere. Suggest all found test source roots for test generation.

**Expected behavior**

In Gradle project Test sources root can be located anywhere, it can be configured in build.gradle file.
So we'll need to be able to generate tests to a directory located outside src.
![Test root out of content notification](https://user-images.githubusercontent.com/37301492/179707697-8c07a2ec-8c23-4355-9d6f-48147986ab5e.jpg)

**Environment**

Checked on IDEA Ultimate 2022.1.3
Gradle 6.3, 7.4

**Potential alternatives**

Current behavior forbids destination folder outside /src/test

**Context**

For example: https://github.com/volivan239/virt-mem-personal/blob/main/build.gradle
Now integrationTest directory can't be selected as destination folder:

alisevych:
@Vassiliy-Kudryashov Please have a look at the referenced project:
https://github.com/volivan239/virt-mem-personal
There is a source root in the root - and it is not found by UTBot
![image](https://user-images.githubusercontent.com/37301492/181272119-e4f71b96-bb44-4e66-bc7f-c149b88eae56.png)

PR #776, opened 2022-08-23
Follow up fix for PR-705

EgorkaKulikov:
# Description

A pull request https://github.com/UnitTestBot/UTBotJava/pull/705/ destroyed `main` branch.
It's first commit was correct, neither the second one.

First commit is reverted, made a little more clean.
Function `Project.isGradle` is converted to a property and renamed.

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Check that some typical scenarios work correctly if we run UTBot as a plugin.
Pay more attention to Gradle projects
Also take in mind the original issue https://github.com/UnitTestBot/UTBotJava/issues/549


volivan239 (approved):
Fixes problems that I had with `main`, LGTM

PR #795, opened 2022-08-26
Cannot find utbot-fuzzing.jar fix

Markoutte:
# Description

This PR tries to fix an action because of `Cannot expand ZIP '/__w/UTBotJava/UTBotJava/utbot-fuzzers/build/libs/utbot-fuzzers-2022.8.jar' as it does not exist`


victoriafomina (approved):
LGTM

Issue #131, opened 2022-06-02
byteToString works for hours sometimes	


alexeyfridman:
SAT-1453

String theory issues

dtim:
Temporarily disabled the test as flaky, will be fixed by PR #603

CaelmBleidd:
Probably, should be fixed with #1050

PR #797, opened 2022-08-26
Fixes for analysis of reflection and Unsafe #721

### LINKED ISSUE #721: IllegalStateException for the Unsafe.putAddress usage leading to JVM crash example - for JDK 11
**Description**

Code sample contains Unsafe.putAddress method call that leads to JVM crash in project with JDK 11.
When mocks are on - IllegalStateException in Concrete Executor is thrown.

**To Reproduce**

1. Run IntelliJ IDEA 2022.1.4 with UTBot installed - [from denis-fokin\gradle_version_upgrade](https://github.com/UnitTestBot/UTBotJava/actions/runs/2848081051)
2. Create a project with JDK 11.
3. Copy org/utbot/examples/exceptions/JvmCrashExamples.java from UTBotJava project, utbot-sample

**Expected behavior**

Tests on JVM crash are supposed to be generated. 
No Concrete Execution errors are expected.

**Actual behavior**

Concrete Execution errors thrown

The following test class is generated - **with plugin from denis-fokin\gradle_version_upgrade**

~~~
package org.utbot.examples.exceptions;

import org.testng.annotations.Test;

public class GradleUpgradeJvmCrashExamplesTest {
    ///region Test suites for executable org.utbot.examples.exceptions.JvmCrashExamples.crash

    ///region

    /**
     * <a href="file:///C:/Users/LWX118~1/AppData/Local/Temp//hs_err_pid6316.log">JVM crash report</a>
     */
    @Test(enabled = false, description = "Disabled due to possible JVM crash")
    public void testCrash1() throws Exception {
        JvmCrashExamples jvmCrashExamples = new JvmCrashExamples();

        // This invocation possibly crashes JVM
        jvmCrashExamples.crash(0);
    }
    ///endregion

    ///region Errors report for crash

    public void testCrash_errors() {
        // 
        // 208 occurrences of:
        // Concrete execution failed

        // 18 occurrences of:
        /* class soot.jimple.StaticFieldRef cannot be cast to class soot.jimple.internal.JimpleLocal (soot.jimple.StaticFieldRef and
        soot.jimple.internal.JimpleLocal are in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @8d8b598) */

    }
    ///endregion

    ///endregion

}
~~~

[Gradle upgrade branch with JDK 11.log](https://github.com/UnitTestBot/UTBotJava/files/9336292/Gradle.upgrade.branch.with.JDK.11.log)

[Unsafe - from main on JDK 11.log](https://github.com/UnitTestBot/UTBotJava/files/9336297/Unsafe.-.from.main.on.JDK.11.log)

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.1.4
Maven project with JDK 11

**Additional context**

Concrete Execution errors are thrown with **plugin published from main branch**
But from main - there is no test generated, the test class is empty.

### LINKED ISSUE #785: Implement a `UtSettings` option to enable or disable sandboxing
**Description**

Sandbox changes the behavior of the concrete executor, and it affects the analysis of code that contains reflection, `Unsafe` class methods etc.

There are currently two options to manage sandbox permissions.
  * Enable a specific set of operations or all operations in the sandbox configuration file.
  * Wrap the body of the method under test with `AccessController.doPrivileged` method call to explicitly enable privileged operations when the MUT is concretely executed.

It seems that these options are sufficient: end users would either agree that potentially dangerous code is not executed (especially implicitly during the test suite generation), or enable a set of "acceptable" operations or their specific invocations.

At the same time, writing unit tests for `UnitTestBot` itself requires more fine-grained configuration of the sandbox. Unit tests for `utbot-framework` should ensure that the symbolic engine and the code generator work with any combination of user settings. Consider a test suite  that checks that calls to `Unsafe` methods are correctly analyzed. We need to check three configurations:
  * Sandboxing is turned on, no unsafe operations are permitted, concrete executor signals an expected error.
  * Sandboxing is turned on, user code is wrapped with `AccessController.doPrivileged`.
  * Sandboxing is turned off, all code is enabled (or all necessary operations are explicitly enabled in the configuration file).

It seems inconvenient to create and drop configuration files when running specific tests. Using `doPrivileged` is OK but it checks a specific configuration, because adding code to MUT changes the behavior of the symbolic engine. As a result, we need a  way to control the behavior of the concrete executor on a per-test basis.

_Proposal:_
    * Add a new command line option to the concrete executor: `--disable-sandbox` (all operations are enabled if it is defined).
    * Add a new `UtSettings` option: `UtSettings.disableSandbox` (`false` by default). If it is `true`, then `--disable-sandbox` argument
      will be added to the concrete executor run command.
    * Add a new utility function `inline fun <reified T> withoutSandbox(block: () -> T): T` to run specific tests with sandboxing turned off.

Sandbox will be enabled by default, and it can be disabled only when necessary. Users can use this option to disable sandboxing as an alternative method to granting all permissions in the sandbox configuration file.

**Expected behavior**

It will be possible to disable sandbox for specific tests and to keep it enabled for all other tests. Users will have one more (simpler) configuration option but no additional power

**Environment**

This feature request is targeted to framework unit tests, although setting the option in the `.utbot/settings.properties` will  affect the plugin and CLI as well. It does not depend on any other environment options.

dtim:
# Description

Add support for symbolic/concrete analysis of reflection and `Unsafe` calls under JDK 11 and improve their analysis under JDK 8.

  * Updated system packages list to prevent mocking of `jdk.internal.*` classes on JDK 11.
  * Refactored `FieldId.isInaccessibleViaReflection` property to consistently use it in all relevant checks (engine, codegen, concrete executor, mocking), updated the class/field list to JDK 11.
  * Implemented a wrapper for `SecurityManager` that simulates the "enable all" policy to avoid long analysis of the real privilege checking code. We can lose branches, but the hypothesis is that we will keep "interesting" branches. Two corresponding methods have been added to the [Class] override to save a bit more of work.

As the concrete executor now implements a sandbox to disable unsafe and dangerous operations, `UtSettings.disableSandbox` option has been introduced. It allows to disable sandboxing without using non-trivial security manager configuration file. A utility function `withoutSandbox` can now be used to disable sandbox in a specific unit test.

Fixes #721
Fixes #785

**Note**: the `StringExamplesTest.testByteToString` unit test has been temporarily disabled as flaky. The change of its behavior seems to be related to this PR (new wrappers may change the set of generated constraints), but the test itself is buggy and very fragile due to limitations of the current `String` wrapper implementation (related issue: #131). The test will be re-enabled as a part of new string support (without Z3-backed `UtNativeString`), as its problems are rooted in the Z3 string theory and are not related to `SecurityManager` and other stuff affected by this PR.

## Type of Change

This PR adds a new way to disable sandbox, alters the list of classes and fields that can be mocked, can make some static fields not meaningful, and replaces the real analysis of privilege checking with a wrapper/overrides.

# How Has This Been Tested?

## Automated Testing

New tests have been added:
  * `org.utbot.examples.exceptions.JvmCrashExamplesTest#testCrashPrivileged`: checks that `Unsafe` call can be analyzed when the MUT contains an explicit call to `doPrivileged` method.
  * `org.utbot.examples.unsafe.UnsafeOperationsTest#checkGetAddressSizeOrZero`: checks a "safe" `Unsafe` call.
  * `org.utbot.examples.unsafe.UnsafeOperationsTest#checkGetAddressSizeOrZeroWithMocks`: checks a "safe" `Unsafe` call with mocking enabled.

## Manual Scenario 

It should be now possible to generate unit tests for the code that involves unsafe calls, e.g.:
```
    public int getAddressSizeOrZero() {
        try {
            Field f = Unsafe.class.getDeclaredField("theUnsafe");
            f.setAccessible(true);
            Unsafe unsafe = (Unsafe) f.get(null);
            return unsafe.addressSize();
        } catch (NoSuchFieldException | IllegalAccessException e) {
            throw new RuntimeException("Reflection failed");
        }
    }
```

It should be possible to generate tests for the "crash" example from the description of #721, but it is necessary to disable summary generation by setting `UtSettings.enableMachineLearningModule = false`. Reason: the only execution that can be generated results in JVM failure, but by default this execution is removed by the summarizer (related issue: #800).

Note: make sure that sandbox is either turned off or configured with all necessary privileges during testing, or the method under test wraps privileged calls to `doPrivileged`, otherwise only "disabled by sandbox" tests will be generated.

# Checklist:

- [X] No new warnings


utbot-framework/src/main/kotlin/org/utbot/engine/SecurityManagerWrapper.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.utbot.framework.util.nextModelName
+import soot.Scene
+import soot.SootClass
+import soot.SootMethod
+
+val utSecurityManagerClass: SootClass
+    get() = Scene.v().getSootClass(UtSecurityManager::class.qualifiedName)
+
+class SecurityManagerWrapper : BaseOverriddenWrapper(utSecurityManagerClass.name) {
+    private val baseModelName: String = "securityManager"
```
Damtev, line 17: Minor: maybe, move it to a companion object?

utbot-framework/src/test/kotlin/org/utbot/examples/exceptions/JvmCrashExamplesTest.kt
```diff
@@ -5,8 +5,18 @@ import org.utbot.examples.DoNotCalculate
...
+import org.utbot.framework.codegen.CodeGeneration
+import org.utbot.framework.plugin.api.CodegenLanguage
 
-internal class JvmCrashExamplesTest : UtValueTestCaseChecker(testClass = JvmCrashExamples::class) {
+internal class JvmCrashExamplesTest : UtValueTestCaseChecker(
+    testClass = JvmCrashExamples::class,
+    testCodeGeneration = true,
+    languagePipelines = listOf(
+        CodeGenerationLanguageLastStage(CodegenLanguage.JAVA),
+        CodeGenerationLanguageLastStage(CodegenLanguage.KOTLIN, CodeGeneration)
```
Damtev, line 17: Do we know why compilation and run for Kotlin are disabled?

utbot-framework/src/test/kotlin/org/utbot/examples/unsafe/UnsafeOperationsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    @Test
+    fun checkGetAddressSizeOrZeroWithMocks() {
+        withoutSandbox {
+            check(
+                UnsafeOperations::getAddressSizeOrZero,
+                eq(1),
+                { r -> r == 8 },
```
Damtev, line 38: Same about bits

Damtev (changes requested):

CaelmBleidd (approved):

Markoutte (approved):

sergeypospelov (approved):

utbot-framework/src/main/kotlin/org/utbot/engine/SecurityManagerWrapper.kt
dtim: Thank you, done.

utbot-framework/src/test/kotlin/org/utbot/examples/exceptions/JvmCrashExamplesTest.kt
dtim: It was an intermediate version I forgot to change back, thank you. Enabled default compilation/run settings.

utbot-framework/src/test/kotlin/org/utbot/examples/unsafe/UnsafeOperationsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.utbot.framework.codegen.CodeGeneration
+import org.utbot.framework.plugin.api.CodegenLanguage
+import org.utbot.framework.plugin.api.MockStrategyApi
+
+internal class UnsafeOperationsTest : UtValueTestCaseChecker(
+    testClass = UnsafeOperations::class,
+    testCodeGeneration = true,
+    languagePipelines = listOf(
+        CodeGenerationLanguageLastStage(CodegenLanguage.JAVA),
+        CodeGenerationLanguageLastStage(CodegenLanguage.KOTLIN, CodeGeneration)
```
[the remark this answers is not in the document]
dtim: Fixed

utbot-framework/src/test/kotlin/org/utbot/examples/unsafe/UnsafeOperationsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    )
+) {
+    @Test
+    fun checkGetAddressSizeOrZero() {
+        withoutSandbox {
+            check(
+                UnsafeOperations::getAddressSizeOrZero,
+                eq(1),
+                { r -> r == 8 },
+                coverage = DoNotCalculate
```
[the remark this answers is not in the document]
dtim: Coverage matcher fail as it the calculated result is only 15/21 instructions, 0/0 branches, and 0/0 lines. I am not sure whether it is a bug in the coverage calculation algorithm, or there really are non-covered instructions (it is possible as we are dealing with complex internal implementation of `Field.setAccessible` and other reflection machinery, as well as with unsafe native calls). I added a TODO comment and will investigate this later.

utbot-framework/src/test/kotlin/org/utbot/examples/unsafe/UnsafeOperationsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    fun checkGetAddressSizeOrZeroWithMocks() {
+        withoutSandbox {
+            check(
+                UnsafeOperations::getAddressSizeOrZero,
+                eq(1),
+                { r -> r == 8 },
+                coverage = DoNotCalculate,
```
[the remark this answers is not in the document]
dtim: The same unclear coverate matcher behavior, will investigate later.

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/process/ChildProcess.kt
```diff
@@ -48,14 +48,21 @@ private fun log(any: Any?) {
 
 private val kryoHelper: KryoHelper = KryoHelper(System.`in`, System.`out`)
 
+/**
+ * Command-line option do disable the sandbox
```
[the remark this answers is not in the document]
dtim: Fixed, thank you.

utbot-framework/src/test/kotlin/org/utbot/examples/unsafe/UnsafeOperationsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        CodeGenerationLanguageLastStage(CodegenLanguage.KOTLIN, CodeGeneration)
+    )
+) {
+    @Test
+    fun checkGetAddressSizeOrZero() {
+        withoutSandbox {
+            check(
+                UnsafeOperations::getAddressSizeOrZero,
+                eq(1),
+                { r -> r == 8 },
```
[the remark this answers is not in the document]
dtim: Yes, and I agree that this is an oversimplification. Changed it with `r!! > 0`.

utbot-framework/src/test/kotlin/org/utbot/examples/unsafe/UnsafeOperationsTest.kt
dtim: Same fix, thank you.

Damtev (approved):

Issue #800, opened 2022-08-27
Summarizer erases JVM crash executions


dtim:
**Description**

When the concrete executor tries to execute a branch that leads to a JVM crash, it creates a `UtExecution` instance that correspond to the failure. The test generator will process this execution in a special way: the generated test case will be disabled and it will contain a comment about the possible JVM crash.

Summarizer considers only two subclasses of `UtExecution`, `UtSymbolicExecution` and `UtFuzzedExecution`, but not the parent class itself. As a result, summarizer erases these failed (JVM crash) executions, and no tests are generated for them.

A related problem: failed executions are not erased during the code generation / summarization phases in `utbot-framework` unit tests, because the result of the call to summarizer is not assigned to anything. It seems that this inconsistency should be fixed as well.

**To Reproduce**

Turn fuzzing off (either via the plugin settings dialog window or by setting `UtSettings` option `useFuzzing` to `false`).

Generate a test suite for the following class using the plugin.
```
import sun.misc.Unsafe;

import java.lang.reflect.Field;
import java.security.AccessController;
import java.security.PrivilegedAction;

public class JvmCrashExample {
    public int crashPrivileged(int i) {
        return AccessController.doPrivileged((PrivilegedAction<Integer>) () -> {
            try {
                Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
                unsafeField.setAccessible(true);
                Unsafe unsafe = (Unsafe) unsafeField.get(null);
                unsafe.putAddress(0, 0);

                if (i == 0) {
                    return i;
                }

                return 1;
            } catch (NoSuchFieldException | IllegalAccessException e) {
                throw new RuntimeException(e);
            }
        });
    }
}
```

Note: this example uses `doPrivileged` to allow the concrete executor to run privileged code (the unsafe call) despite the sandbox limitations. If the crashing code is not wrapped in `doPrivileged`, and the operation is not explicitly enabled by a sandbox configuration file, a special test case will be generated instead with the annotation `@Disabled(value = "Disabled due to sandbox")`. To reproduce this issue, please either allow all unsafe operations to the concrete executor, or use `doPrivileged`.

**Expected behavior**

A disabled test case should be generated.

```
import org.junit.jupiter.api.Test;

public class JvmCrashExampleTest {
    ///region Test suites for executable unsafe.JvmCrashExample.crashPrivileged

    ///region

    /**
[...]

PR #804, opened 2022-08-29
Make UtExecution abstract, introduce UtFailedExecution as its child

dtim:
# Description

Make `UtExecution` abstract, introduce `UtFailedExecution` as its child.

*Rationale*

There is a `UtExecution` class that represent any execution regardless of the way it has been found. It has two subclasses, `UtSymbolicExecution` and `UtFuzzedExecution`. The separation of these entities is artificial: `UtSymbolicExecution` contains data provided by the symbolic engine and used by the summarizer that `UtFuzzedExecution` does not have. Having two distinct child classes instead of a single `UtExecution` allowed the summarizer to correctly handle executions coming from two sources, the symbolic engine and the fuzzer.

There is another special kind of execution: an execution that is produced by a concrete executor when running the method leads to a failure (e.g., JVM crash). It is not possible to produce any "after state" in this case, and the execution result is always a failure.

Currently these executions are created as just `UtExecution` instances, and summarizer ignores them (see #800). As a result, executions may be lost.

It is necessary to correctly process these failed executions in the summarizer, and to distinguish them from "normal" executions. It seems not very convenient to make explicit checks to distinguish between a parent class and its subclasses.

This PR uses the following approach.

1. Introduce a new class, `UtFailedExecution`, that is a subclass of `UtExecution` but has a restricted set of parameters so we can't create a "successful" failed execution instance by mistake.
2. Make `UtExecution` abstract to prevent creating "unclassified" execution instances.

That way the summarizer can explicitly handle failed executions and generate correct summaries for them.

*Limitations*

This solution seems not to be "final" in any way. We have to properly define the interface between execution producers (fuzzer, symbolic engine, concrete executor, or any other kind of producer that can be added in the future) and the summarizer. In particular, it seems right to separate executions not by their source but by a set of available information (e.g., basic executions, executions with attached Jimple/bytecode/source code information, summarizer-annotated executions and so on). Maybe the additional data can be attached in some general form as annotations instead. This "correct" refactoring is a major enterprise and can't be performed immediately.

At the same time, we need to fix the issue with lost executions, so an "intermediate" refactoring is necessary.

Related issue: #800

## Type of Change

- Refactoring (typos and non-functional changes) 

# How Has This Been Tested?

## Automated Testing

All existing unit tests should pass.

## Manual Scenario 

None (it will be possible to test the fix for #800, for which this refactoring is a prerequisite).

- [X] No new warnings

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -144,7 +144,7 @@ sealed class UtResult
  * - coverage information (instructions) if this execution was obtained from the concrete execution.
  * - comments, method names and display names created by utbot-summary module.
  */
-open class UtExecution(
+abstract class UtExecution(
     val stateBefore: EnvironmentModels,
     val stateAfter: EnvironmentModels,
     val result: UtExecutionResult,
@@ -223,6 +223,27 @@ class UtSymbolicExecution(
         )
     }
 }
+
+/**
+ * Execution that result in an error (e.g., JVM crash or another concrete execution error).
+ *
+ * Contains:
+ * - state before the execution;
+ * - result (a [UtExecutionFailure] or its subclass);
+ * - coverage information (instructions) if this execution was obtained from the concrete execution.
+ * - comments, method names and display names created by utbot-summary module.
+ *
+ * This execution does not contain any "after" state, as it is generally impossible to obtain
+ * in case of failure. [MissingState] is used instead.
+ */
+class UtFailedExecution(
+    stateBefore: EnvironmentModels,
+    result: UtExecutionFailure,
+    coverage: Coverage? = null,
+    summary: List<DocStatement>? = null,
+    testMethodName: String? = null,
+    displayName: String? = null
+) : UtExecution(stateBefore, MissingState, result, coverage, summary, testMethodName, displayName)
 
 open class EnvironmentModels(
     val thisInstance: UtModel?,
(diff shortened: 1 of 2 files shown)


utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -144,7 +144,7 @@ sealed class UtResult
  * - coverage information (instructions) if this execution was obtained from the concrete execution.
  * - comments, method names and display names created by utbot-summary module.
  */
-open class UtExecution(
+abstract class UtExecution(
```
amandelpie, line 147: @dtim @Markoutte I remember that month ago we discussed that it could be UTExecution which are not defeined on early stages what it is (fuzzing or symbolic). Could we do this changes? Do we really cover by UtSymbolic/UtFuzzing/UtFailed all variants of the UtExecution at this momemnt?

Damtev (approved):
LGTM

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
Markoutte: Looks like it covers all at the moment. Isn't it a problem when we add another implementation for UtExecution for a summarization?

dtim: I think that major changes in summarization/engine/fuzzer interfaces will require another refactoring anyway. In this PR I just want to distinguish more clearly between all possible execution types, to make it easier (as I see it, I may be wrong) for the summarizer to check for failed executions.

If the issue #800 can be fixed without this hierarchy change, it's OK for me to drop this PR and discuss a better/conceptually correct refactoring. I think that the final decision is up to @amandelpie

amandelpie (approved):

PR #854, opened 2022-09-04
Prefer symbolic executions during minimization

### LINKED ISSUE #843: Prioritize Symbolic Engine executions
**Description**

Sometimes we need to turn off *Symbolic Engine* and stop the test generation (and do not collect this execution) in case of mocking taking place when generating parameterized tests. However, *Fuzzer* can still add its executions which may lead to unwanted executions being processed later.
The suggestion is to prioritize *Symbolic Engine* executions in minimization process.

dtim:
# Description

Suppose we have two executions with the same coverage, one produced by the symbolic engine, and one produced by the fuzzer. From the user's point of view, they are equivalent, and the minimizer can keep any of them. Usually it will be fuzzed one, because fuzzer produces executions earlier that the symbolic engine, and the minimizer orders executions with the same coverage by their order in the execution list.

When parametrized tests are generated, fuzzed tests tend to be excluded due to mock processing issues. If a fuzzed tests has been selected by the minimizer instead of the symbolic one, the corresponding path remains uncovered: one (symbolic) test has been dropped by the minimizer, and another (fuzzed) test has been excluded by the code generator.

The coverage of the entire parametrized test suite may be improved is minimizer selects symbolic executions when possible. It does not guarantee the full coverage (it is possible to have only fuzzed tests for some paths), but we expect that the coverage may be improved that way.

To let users to control the minimizer behavior, new `UtSettings` option is introduced: `preferSymbolicExecutionsDuringMinimization`. When it is set to `true` (default), the minimizer will consider the execution source (specifically, its class: `UtSymbolicExecution` or `UtFuzzedExecution` (`UtFailedExecution` and any potential other direct `UtExecution` subclasses are assigned the lowest priority, but `UtFailedExecution` does not have any coverage anyway, so it is processed separately).

Fixes #843 

## Type of Change

This fix changes the set of generated tests. It should not decrease coverage, and is expected to increase coverage when generating parametrized tests (_please report a bug if coverage is decreased when `preferSymbolicExecutionsDuringMinimization` is `true`_).

This PR also changes the minimizer interface.

# How Has This Been Tested?

## Automated Testing

All existing minimizer tests should pass.

Two new minimizer tests have been added:
  * `org.utbot.framework.minimization.MinimizationGreedyEssentialTest#testWithSourcePriority`
  * `org.utbot.framework.minimization.MinimizationGreedyEssentialTest#testWithoutSourcePriority`

## Manual Scenario 

Make sure that fuzzer is enabled in `UtSettings`, and both fuzzer and symbolic engine are enabled in the plugin configuration window.

Create any method simple enough so both fuzzer and symbolic engine can cover all branches. An example:
```
package simple;

public class SimpleExample {
    public int foo(int a, int b) {
        if (a > b)
            return a - b;
        else
            return a + b;
    }
}
```

Set `UtSettings.preferSymbolicExecutionsDuringMinimization` to `true` and generate a test suite for `foo`. Test cases produced by the symbolic engine should be generated.

```
package simple;

import org.junit.Test;

import static org.junit.Assert.assertEquals;

public class SimpleExampleTest {
    ///region Test suites for executable simple.SimpleExample.foo

    ///region SUCCESSFUL EXECUTIONS for method foo(int, int)

    /**
     * <pre>
     * Test executes conditions:
     *     {@code (a > b): False }
     * returns from: {@code return a + b; }
     * </pre>
     */
    @Test
    public void testFoo_ALessOrEqualB() {
        SimpleExample simpleExample = new SimpleExample();

        int actual = simpleExample.foo(-240, -240);

        assertEquals(-480, actual);
    }

    /**
     * <pre>
     * Test executes conditions:
     *     {@code (a > b): True }
     * returns from: {@code return a - b; }
     * </pre>
     */
    @Test
    public void testFoo_AGreaterThanB() {
        SimpleExample simpleExample = new SimpleExample();

        int actual = simpleExample.foo(1, 0);

        assertEquals(1, actual);
    }
    ///endregion

    ///endregion
}
```

Set `UtSettings.preferSymbolicExecutionsDuringMinimization` to `false` and generate a test suite for `foo` again. This time, the plugin should generate test cases produced by the fuzzer.

```
package simple;

import org.junit.Test;

import static org.junit.Assert.assertEquals;

public class SimpleExampleTest {
    ///region Test suites for executable simple.SimpleExample.foo

    ///region

    @Test
    public void testFooReturnsZero() {
        SimpleExample simpleExample = new SimpleExample();

        int actual = simpleExample.foo(-1, 1);

        assertEquals(0, actual);
    }

    @Test
    public void testFooWithCornerCase() {
        SimpleExample simpleExample = new SimpleExample();

        int actual = simpleExample.foo(Integer.MAX_VALUE, 1);

        assertEquals(2147483646, actual);
    }
    ///endregion

    ///endregion
}
```

- [X] No new warnings

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/UtSettings.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/UtSettings.kt
@@ -378,6 +378,11 @@ object UtSettings : AbstractSettings(
      */
     var disableSandbox by getBooleanProperty(false)
 
+    /**
+     * Exclude fuzzer-produced executions during minimization if there is an equivalent symbolic execution
+     */
+    var preferSymbolicExecutionsDuringMinimization by getBooleanProperty(true)
+
 }
 
 /**
--- a/utbot-framework-test/src/test/kotlin/org/utbot/framework/minimization/MinimizationGreedyEssentialTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/framework/minimization/MinimizationGreedyEssentialTest.kt
@@ -68,4 +68,46 @@ class MinimizationGreedyEssentialTest {
         val minimizedExecutions = GreedyEssential.minimize(executions)
         assertEquals((1..size).toList(), minimizedExecutions.sorted())
     }
+
+    @Test
+    fun testWithSourcePriority() {
+        val executions = mapOf(
+            10 to listOf(1, 2, 3, 4, 5),
+            20 to listOf(2, 3, 4, 5, 6, 7),
+            21 to listOf(2, 3, 4, 5, 6, 7),
+            25 to listOf(1, 7, 2, 3, 5),
+            30 to listOf(8, 9),
+            50 to listOf(1, 8, 9),
+            60 to listOf(1, 9)
+        )
+
+        val sourcePriorities = mapOf(
+            10 to 0,
+            20 to 1,
+            21 to 0,
+            25 to 0,
+            30 to 1,
+            50 to 1,
+            60 to 0
+        )
+
+        val minimizedExecutions = GreedyEssential.minimize(executions, sourcePriorities)
+        assertEquals(listOf(21, 50), minimizedExecutions)
+    }
+
+    @Test
+    fun testWithoutSourcePriority() {
+        val executions = mapOf(
+            10 to listOf(1, 2, 3, 4, 5),
+            20 to listOf(2, 3, 4, 5, 6, 7),
+            21 to listOf(2, 3, 4, 5, 6, 7),
+            25 to listOf(1, 7, 2, 3, 5),
+            30 to listOf(8, 9),
+            50 to listOf(1, 8, 9),
+            60 to listOf(1, 9)
+        )
+
+        val minimizedExecutions = GreedyEssential.minimize(executions)
+        assertEquals(listOf(20, 50), minimizedExecutions)
+    }
 }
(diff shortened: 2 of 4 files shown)


dtim:
Following the discussions with @sergeypospelov and @Markoutte.

This PR generally implements the idea of providing additional execution priorities map to minimizer to specify that some executions are more "important" or "interesting" that others with the same coverage. It may be useful per se (although probably requires some more design), but it seems to be a hack w.r.t. the original issue (#843).

  * The temporary `UtExecution` class hierarchy is spreading into conceptually isolated parts, such as the minimizer.
  * It seems that this solution is only partial. Let us have two executions, a `UtSymbolicExecution` covering 6 instructions, and a `UtFuzzedExecution` that covers the same 6 instructions plus 2 other instructions, so it has a strictly better coverage. Although the symbolic execution has higher priority, the minimizer will select the fuzzed execution as its coverage is better. If we remove the fuzzed execution later during parametrized code generation, we will lose both executions, while the correct behavior would be to keep the symbolic execution, even it is not as good as the fuzzed one. At the same time, we definitely want to use the fuzzed execution in the regular case (when no parametrized tests are generated, so no fuzzed executions are removed).

As a potential alternative, we could use three stages:
  1. Select only symbolic executions.
  2. Add fuzzed executions that increase coverage (optional, turned off if generating parametrized tests).
  3. Minimize the resulting collection.

Issue #691, opened 2022-08-09
Android Studio: Mockito dependencies installation (and maybe JUnit)


nikitavlaev:
**Description**

In Android Studio `mockito-core` is not installed correctly in Gradle project.
This leads to the project build failure.

Also, other dependencies should be checked for that error (`mockito-inline`, `JUnit`).

**To Reproduce**

1. Create new project in Android Studio
2. Add Android library
3. Use plugin to generate tests
4. Choose to configure mockito
5. Experience the build failure

**Expected behavior**

Required dependencies are supposed to be installed correctly.

**Actual behavior**

Build failure.

**Visual proofs (screenshots, logs, images)**

To be added.

**Environment**

Clean Gradle project in Android Studio.

**Additional context**

None.

EgorkaKulikov:
This problem is actual for `mockito-core` and `junit5`

tyuldashev:
Tested with Android Studio Chipmunk | 2021.2.1 Patch 2 and UTBot 2022.10.471.

PR #872, opened 2022-09-07
Correct installation dependencies into Android Studio

EgorkaKulikov:
# Description

Introduce custom modifier specifically for Android Gradle projects

Fixes # ([691](https://github.com/UnitTestBot/UTBotJava/issues/691))

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Install all test frameworks and mock framework one by one into plugin and check that dependencies are correct. 
Do this check in Intellij Idea for all build systems and in Android Studio


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/UtAndroidGradleProjectModelModifier.kt
```diff
@@ -0,0 +1,140 @@
...
+        val listener = object : GradleSyncListener {
+            override fun syncSucceeded(project: Project) {
+                promise.setResult(null)
+            }
+
+            override fun syncFailed(project: Project, errorMessage: String) {
+                promise.setError(errorMessage)
+            }
+        }
+        GradleSyncExecutor(project).sync(request, listener)
```
Vassiliy-Kudryashov, line 136: Why don't you use GradleSyncInvoker.getInstance().requestProjectSync(...) instead? See original com.android.tools.idea.gradle.AndroidGradleJavaProjectModelModifier#doAndroidGradleSync

utbot-intellij/src/main/resources/META-INF/plugin.xml
```diff
@@ -35,6 +35,7 @@
         <registryKey defaultValue="false" description="Enable editing Kotlin test files" key="kotlin.ultra.light.classes.empty.text.range"/>
         <postStartupActivity implementation="org.utbot.intellij.plugin.ui.GotItTooltipActivity"/>
         <projectModelModifier implementation="org.utbot.intellij.plugin.util.UtProjectModelModifier"/>
+        <projectModelModifier implementation="org.utbot.intellij.plugin.util.UtAndroidGradleJavaProjectModelModifier" order="first"/>
```
Vassiliy-Kudryashov, line 38: Now it requires android plugin to be installed. Probably we'll need  <depends>org.jetbrains.android</depends> at line 13

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/UtAndroidGradleProjectModelModifier.kt
EgorkaKulikov: Discussed, it's because Java version compatibilities

utbot-intellij/src/main/resources/META-INF/plugin.xml
EgorkaKulikov: More than that, we have smth like `isBuildWithGradle`

EgorkaKulikov:
Replaced with PR-1180.

PR #914, opened 2022-09-13
No true branch test is generated for code with String.split(";\\s") #885

### LINKED ISSUE #885: No true branch test is generated for code with String.split(";\\s")
**Description**

No test is generated for the true branch of the condition with String.split(";\\s")

**To Reproduce**

1. IntelliJ IDEA with UTBot plugin installed
2. Fuzzing is on, slider is in default position
3. Add the following Java class:

```java
public class StringCharAt {
    public boolean isSplitted(String s) {
        if (s.split(";\\s").length > 1) {
            return true;
        }
        return false;
    }
}
```

4. Use plugin to generate tests
5. Open the generated test

**Expected behavior**

Tests are supposed to be generated for all code branches.

**Actual behavior**

There is only one successful test - with empty String, which is for the false branch.

The following test class is generated:
~~~java
public class StringCharAtTest {
    ///region Test suites for executable StringCharAt.isSplitted

    ///region FUZZER: SUCCESSFUL EXECUTIONS for method isSplitted(java.lang.String)

    @Test
    @DisplayName("isSplitted: s = empty string -> return false")
    public void testIsSplittedReturnsFalseWithEmptyString() {
        StringCharAt stringCharAt = new StringCharAt();

        boolean actual = stringCharAt.isSplitted("");

        assertFalse(actual);
    }
    ///endregion

    ///region SYMBOLIC EXECUTION ENGINE: ERROR SUITE for method isSplitted(java.lang.String)

    /**
     * <pre>
     * Test invokes:
     *     {@link String#split(String)} once
     *
     * throws NullPointerException in: s.split("<\\s").length > 1
     * </pre>
     */
    @Test
    @DisplayName("isSplitted: s.split(\"<\\\\s\").length > 1 -> ThrowNullPointerException")
    public void testIsSplitted_StringSplit() {
        StringCharAt stringCharAt = new StringCharAt();
        
        /* This test fails because method [StringCharAt.isSplitted] produces [java.lang.NullPointerException]
            StringCharAt.isSplitted(StringCharAt.java:11) */
        stringCharAt.isSplitted(null);
    }
    ///endregion

    ///endregion
}
~~~

**Environment**

Windows 10 Pro
JDK 11

**Additional context**

The same number of tests is generated for the following regex used : "<\\s", ";\\s"
If only "\\s" or ";" is used - all branches are covered.

Markoutte:
# Description

RegexModelProvider now recognize argument of method "String.split" call as a pattern to produce initial string from regex.

Fixes #885

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Try to generate test as in #885 

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RegexModelProvider.kt
+++ b/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RegexModelProvider.kt
@@ -56,9 +56,10 @@ object RegexModelProvider : ModelProvider {
 
     private fun FuzzedContext.isPatterMatchingContext(): Boolean {
         if (this !is FuzzedContext.Call) return false
+        val stringMethodWithRegexArguments = setOf("matches", "split")
         return when {
             method.classId == Pattern::class.java.id -> true
-            method.classId == String::class.java.id && method.name == "matches" -> true
+            method.classId == String::class.java.id && stringMethodWithRegexArguments.contains(method.name) -> true
             else -> false
         }
     }


utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RegexModelProvider.kt
```diff
@@ -56,9 +56,10 @@ object RegexModelProvider : ModelProvider {
 
     private fun FuzzedContext.isPatterMatchingContext(): Boolean {
         if (this !is FuzzedContext.Call) return false
+        val stringMethodWithRegexArguments = setOf("matches", "split")
         return when {
             method.classId == Pattern::class.java.id -> true
-            method.classId == String::class.java.id && method.name == "matches" -> true
+            method.classId == String::class.java.id && stringMethodWithRegexArguments.contains(method.name) -> true
```
dtim, line 62: There are two more methods of `String` that accept regexes:
  * `public String replaceFirst​(String regex, String replacement)`
  * `public String replaceAll(String regex, String replacement)`

dtim (approved):
The fix looks good. It seems that it is possible to cover a few more cases by extracting regexes from `replaceFirst` and `replaceAll` calls on `String`.

PR #921, opened 2022-09-13
Concrete executor logs

### LINKED ISSUE #870: Log of concrete execution doesn't print exception anymore
**Description**

At the moment, logs that are created by concrete execution don't contains printed exception anymore, but have too much info about data transfer between main process and child process.

**To Reproduce**

1. Enable logs using `~/.utbot/settings.properties`: `logConcreteExecutionErrors=true`
2. Run any test generation which leads to concrete execution failure. For example, #739
3. Find logs in temp folder: `~\AppData\Local\Temp\UTBot\utbot-childprocess-errors`
4. Open it.

**Expected behavior**

Log should contain some system information and an exception that is translated into broken test in main process.

**Actual behavior**

Log contains some child process problem (#848), many lines like this one:

```
16:59:27.308 | sent cmd: org.utbot.framework.concrete.UtConcreteExecutionResult@2efe6272
````

and no information about exception.

**Additional context**

Looks like this behavior was introduces with #748

### PR DESCRIPTION
**Pull request by Domonion**
# Description

Fixes #870
Moved logs in `RuntimeTraceStorage` and `DynamicClassTransformer` to ChildProcess.kt logging system

## Type of Change

### DIFF
--- a/utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/et/TraceHandler.kt
+++ b/utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/et/TraceHandler.kt
@@ -9,6 +9,7 @@ import org.objectweb.asm.MethodVisitor
 import org.objectweb.asm.Opcodes
 import org.objectweb.asm.Type
 import org.objectweb.asm.commons.LocalVariablesSorter
+import org.utbot.instrumentation.process.logError
 
 sealed class InstructionData {
     abstract val line: Int
@@ -111,6 +112,8 @@ class ProcessingStorage {
  * Storage to which instrumented classes will write execution data.
  */
 object RuntimeTraceStorage {
+    internal var alreadyLoggedIncreaseStackSizeTip = false
+
     /**
      * Contains ids of instructions in the order of execution.
      */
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/et/TraceHandler.kt [hunk 1]
```diff
@@ -151,7 +152,7 @@ object RuntimeTraceStorage {
             this.`$__trace__`[current] = id
             this.`$__counter__` = current + 1
         } else {
-            System.err.println("Stack overflow (increase stack size Settings.TRACE_ARRAY_SIZE)")
+            logDebug { "Stack overflow (increase stack size Settings.TRACE_ARRAY_SIZE)" }
```
sergeypospelov, line 155:
```suggestion
            if (alreadyLoggedIncreaseStackSizeTip) {
                logError { "Stack overflow (increase stack size Settings.TRACE_ARRAY_SIZE)" }
            }
```

sergeypospelov: `alreadyLoggedIncreaseStackSizeTip` can be a private util variable in `RuntimeStorage`.

sergeypospelov (changes requested):

sergeypospelov (approved):

PR #964, opened 2022-09-19
Fix wrapper concrete types

### LINKED ISSUE #957: Engine doesn't cover branch instanceof String
**Description**

Engine generates 1 execution for this method, but doesn't cover throw statement.
```java
public static void charSeqIsNotString(CharSequence cs) {
    if (cs instanceof String) {
        throw new IllegalStateException("CharSequence must be not String!");
    }
}
```

**To Reproduce**

1. Create the mentioned method
2. Make sure that fuzzing turned of (Tools -> UnitTestBot -> Code analysis toggle must be in the rightest position)
3. Use plugin to generate tests
4. Open the generated test
5. Run tests with coverage

**Expected behavior**

2 tests are supposed to be generated witch cover all statements.

**Actual behavior**

1 test generated that don't covers throw statement.

image

**Additional context**

This bug is a part of issue #820

SBOne-Kenobi:
# Description

The problem was in wrappers instantiation. Initialized concrete types were incorrect. I've provided method for wrappers witch returns set of possible types in resolving. Now all of set's sizes equals 1.

Fixes #957

## Type of Change

- Refactoring (typos and non-functional changes) 

# How Has This Been Tested?

## Automated Testing

org.utbot.examples.casts.InstanceOfExampleTest

## Manual Scenario 

Described in issue #957 

# Checklist:

- [ ] No new warnings


utbot-framework/src/main/kotlin/org/utbot/engine/CollectionWrappers.kt
```diff
@@ -405,6 +410,9 @@ private val UT_GENERIC_ASSOCIATIVE_CLASS
 private val UT_GENERIC_ASSOCIATIVE_SET_EQUAL_GENERIC_TYPE_SIGNATURE =
     UT_GENERIC_ASSOCIATIVE_CLASS.getMethodByName(UtGenericAssociative<*, *>::setEqualGenericType.name).signature
 
+val ARRAY_OBJECT_TYPE: Type
+    get() = ArrayType.v(OBJECT_TYPE, 1)
```
CaelmBleidd, line 414: I'm not sure whether it is the right way to get array type or not since it creates a new one instead of returning an already existed. Better to use `OBJECT_TYPE.arrayType`

utbot-framework/src/main/kotlin/org/utbot/engine/StreamWrappers.kt
```diff
@@ -81,6 +82,9 @@ abstract class StreamWrapper(
             }
     }
 
+    override fun getPotentialPossibleTypes(type: Type): Set<Type> =
+        setOf(STREAM_TYPE)
```
CaelmBleidd, line 86: Why do you use this type? Shouldn't it be something related to `utStreamClass.overriddenStreamClassId`?

utbot-framework/src/main/kotlin/org/utbot/engine/Strings.kt
```diff
@@ -348,6 +355,9 @@ sealed class UtAbstractStringBuilderWrapper(className: String) : BaseOverriddenW
             }
     }
 
+    override fun getPotentialPossibleTypes(type: Type): Set<Type> =
+        setOf(STRING_TYPE)
```
CaelmBleidd, line 359: Why? Does `StringBuilder` relate to`String`?

CaelmBleidd (changes requested):
I have many questions about types choice, I think we need additional documentation or explanations in the code, why it is done this way

utbot-framework/src/main/kotlin/org/utbot/engine/ArrayObjectWrappers.kt
```diff
@@ -415,6 +419,9 @@ class AssociativeArrayWrapper : WrapperInterface {
         return model
     }
 
+    override fun getPotentialPossibleTypes(type: Type): Set<Type> =
+        setOf(associativeArrayType)
```
dtim, line 423: It is OK to return our internal class as a possible concrete class?

utbot-framework/src/main/kotlin/org/utbot/engine/CollectionWrappers.kt
```diff
@@ -137,6 +139,9 @@ abstract class BaseContainerWrapper(containerClassName: String) : BaseOverridden
             }
     }
 
+    override fun getPotentialPossibleTypes(type: Type): Set<Type> =
+        setOf(Scene.v().getSootClass(chooseClassIdWithConstructor(type.classId).canonicalName).type)
```
dtim, line 143: I suppose that this code can return not only concrete classes but interfaces as well. It seems that it may be a bit dangerous. What do you think?

dtim (changes requested):
I don't fully understand the logic, and have several concerns about the selection of types. It would be nice to discuss your PR in more details.

utbot-framework/src/main/kotlin/org/utbot/engine/CollectionWrappers.kt
SBOne-Kenobi: It's ok, but I'll rewrite

utbot-framework/src/main/kotlin/org/utbot/engine/ArrayObjectWrappers.kt
SBOne-Kenobi: Yes, it's internal class, this method never called for this class. We use it only to override key-value storages implementation symbolically.

utbot-framework/src/main/kotlin/org/utbot/engine/CollectionWrappers.kt
SBOne-Kenobi: Are you sure? chooseClassIdWithConstructor must return class with callable constructor (I suppose) that we use for creation assemble model.

utbot-framework/src/main/kotlin/org/utbot/engine/ArrayObjectWrappers.kt
```diff
@@ -264,6 +264,13 @@ class RangeModifiableUnlimitedArrayWrapper : WrapperInterface {
         return resultModel
     }
 
+    override fun getPossibleConcreteTypes(type: Type): Set<Type> {
+        val possibleObjectTypes = Scene.v().classes.map { it.type }
+        return possibleObjectTypes.mapTo(mutableSetOf()) {
+            it.arrayType
+        }
```
CaelmBleidd, line 271: Now it contains all the types, including interfaces, abstract classes, artificial entities, our own classes, etc.

utbot-framework/src/main/kotlin/org/utbot/engine/Extensions.kt
```diff
@@ -424,3 +427,10 @@ val SootMethod.isUtMockAssumeOrExecuteConcretely
  */
 val SootMethod.isPreconditionCheckMethod
     get() = declaringClass.isOverridden && name == "preconditionCheck"
+
+val ClassId.sootType: Type
+    get() = when {
+        isPrimitive -> Scene.v().getType(name)
```
CaelmBleidd, line 433: Does it work this way? Did you check it?

CaelmBleidd (changes requested):
TBH, I don't have an opinion about this request. Probably, we should discuss it in person before merging

PR #967, opened 2022-09-19
Mark distinct test for Stream as slow

Damtev:
# Description

Mark `org.utbot.examples.stream.BaseStreamExampleTest#testDistinctExample` as slow (average 25 minutes).


victoriafomina (approved):
LGTM

PR #971, opened 2022-09-19
Check for ArrayStoreException on array update

### LINKED ISSUE #923: Detect `ArrayStoreException` on illegal array writes
**Description**

The symbolic engine does not check array store operations and does not detect assignments to array elements that result in `ArrayStoreException`. This limitation leads to lost executions on code that involve array assignments or calls to a large number of library methods (e.g., `System.arraycopy`, `Collection.toArray(T[])` etc).

**To Reproduce**

Create a sample class with the following code:
```
public class ArrayStoreExample {
    public int foo() {
        Integer[] integers = { 1, 2, 3 };
        Object[] objects = integers;
        objects[1] = "";
        return 1;
    }
}
```

Generate the test suite for this class (or just for `foo` method).

**Expected behavior**

A single test case with expected `ArrayStoreException` should be generated.

**Actual behavior**

No tests are generated for `foo`.

**Environment**

No specific environment: the check is currently missing and does not depend on any UnitTestBot configuration.

**Additional context**

`ArrayStoreException` is thrown to indicate that an attempt has been made to store the wrong type of object into an array of objects:

SAT-1088, SAT-1089

dtim:
# Description

This PR adds a check to `TraversalContext.traverseAssignLeftPart` to detect if `ArrayStoreException` should be thrown for the given array element assignment statement.

The exception is generated if the type of the value on the right-hand side of the assignment is not a subtype of the array elements type and is not null.

This change affects all code that explicitly assigns array elements, as well as all calls to `Objects.arraycopy` and similar `Arrays` static methods.

Fixes #923

## Type of Change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

New executions may be generated for the code that involves array modification.

# How Has This Been Tested?

## Automated Testing

All existing unit tests should pass.

A new test suite `org.utbot.examples.arrays.ArrayStoreExceptionExamplesTest` has been added, it should pass both with and without concrete execution.

## Manual Scenario 

Generate a test suite (with fuzzer turned off) for a method that assigns values to array elements of a incompatible reference type, e.g.:
```
public class Foo {
    public int foo() {
        Integer[] integers = { 1, 2, 3 };
        Object[] objects = integers;
        objects[1] = "";
        return 1;
    }
}
```

A single test should be generated, that corresponds to `ArrayStoreException`:
```
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.DisplayName;

public class FooTest {
    ///region Test suites for executable basic.Foo.foo

    ///region SYMBOLIC EXECUTION ENGINE: ERROR SUITE for method foo()

    /**
     * @utbot.classUnderTest {@link Foo}
     * @utbot.methodUnderTest {@link Foo#foo()}
     * @utbot.throwsException {@link ArrayStoreException} in: objects[1] = "";
     */
    @Test
    @DisplayName("foo: objects[1] = \"\" -> ThrowArrayStoreException")
    public void testFoo_IntegerValueOf() {
        Foo foo = new Foo();
        
        /* This test fails because method [basic.Foo.foo] produces [java.lang.ArrayStoreException: java.lang.String]
            basic.Foo.foo(Foo.java:9) */
        foo.foo();
    }
    ///endregion

    ///endregion
}
```

## Checklist

- [X] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
@@ -65,9 +65,6 @@ import org.utbot.engine.pc.mkNot
 import org.utbot.engine.pc.mkOr
 import org.utbot.engine.pc.select
 import org.utbot.engine.pc.store
-import org.utbot.engine.symbolic.HardConstraint
-import org.utbot.engine.symbolic.SoftConstraint
-import org.utbot.engine.symbolic.Assumption
 import org.utbot.engine.symbolic.emptyAssumption
 import org.utbot.engine.symbolic.emptyHardConstraint
 import org.utbot.engine.symbolic.emptySoftConstraint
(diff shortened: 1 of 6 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -767,6 +764,56 @@ class Traverser(
...
+
+            // It is related to [org.utbot.engine.pc.Z3TranslatorVisitor.filterInappropriateTypes] that removes
+            // internal engine classes for [java.lang.Object] type storage, and this logic is not fully
+            // consistent with the negation.
+
+            // Here we have a specific test for [java.lang.Object] as the type of array elements:
+            // any reference type may be assigned to elements of an Object array, so we should not generate
+            // [ArrayStoreException] in these cases.
+
+            // TODO: remove enclosing `if` when [UtIfExpression] negation is fixed
```
CaelmBleidd, line 807: Do we have an issue with it?

CaelmBleidd (approved):
LGTM

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
dtim: I created the issue: #1007.

PR #988, opened 2022-09-21
Fuzzer should generate tests for simple collection with generic types #875

### LINKED ISSUE #875: Fuzzer should generate tests for simple collection with generic types
**Description**

For simple methods fuzzer can and should generate test that fills collection with some data. Consider this example:

```java
public <T> Map<String, T> sortedMap(Map<String, T> source) {
    return new TreeMap<>(source);
}
```

Both fuzzer and symbolic execution create empty map. But this test is not very interesting. Fuzzer should add some values into a map to run it with them and check, how it affects the result.

Markoutte:
# Description

Introduces provider to generate collections with generic types. Now, fuzzer creates tests with values for such types as `Collection<Integer>`, `Map<MyClass, String>`, and recursive types like: `Collection<List<Double>>`, `Map<String, Set<Integer>>`.

Fixes #875

## Type of Change

# How Has This Been Tested?

## Automated Testing

Run:

```
org.utbot.framework.plugin.api.ModelProviderTest
org.utbot.framework.plugin.api.CollectionModelProviderTest
```

## Manual Scenario 

Here several examples which can be tested:

```java
public class CollectionExample {

    /**
     * Should create unsorted list that will be sorted as a result.
     */
    public static <T extends Number> Collection<T> sorted(Collection<T> source) {
        return source.stream().sorted().collect(Collectors.toList());
    }

    /**
     * Should create at least both answers: one that finds the key, and another returns null.
     */
    public static String getKeyForValue(Map<String, Number> map, Number value) {
        for (Map.Entry<String, Number> entry : map.entrySet()) {
            if (Objects.equals(entry.getValue(), value)) {
                return entry.getKey();
            }
        }
        return null;
    }

    /**
     * Should find a branch that returns true and size of array is greater than 1 (non-trivial).
     */
    public static boolean isDiagonal(Collection<Collection<Double>> matrix) {
        int cols = matrix.size();
        if (cols <= 1) {
            return false;
        }
        int i = 0;
        for (Collection<Double> col : matrix) {
            if (col.size() != cols) {
                return false;
            }
            int j = 0;
            for (Double value : col) {
                if (i == j && value == 0.0) return false;
                if (i != j && value != 0.0) return false;
                j++;
            }
            i++;
        }
        return true;
    }

    /**
     * Checks that different collections can be created. Part 1
     */
    public boolean allCollectionAreSameSize1(
            Collection<Integer> c,
            List<Integer> l,
            Set<Integer> s,
            SortedSet<Integer> ss,
            Deque<Integer> d,
            Iterable<Integer> i
    ) {
        if (c.size() != l.size()) {
            return false;
        }
        if (l.size() != s.size()) {
            return false;
        }
        if (s.size() != ss.size()) {
            return false;
        }
        if (ss.size() != d.size()) {
            return false;
        }
        if (d.size() != StreamSupport.stream(i.spliterator(), false).count()) {
            return false;
        }
        return true;
    }

    /**
     * Checks that different collections can be created. Part 2
     */
    public boolean allCollectionAreSameSize2(
            Iterable<Integer> i,
            Stack<Integer> st,
            NavigableSet<Integer> ns,
            Map<Integer, Integer> m,
            SortedMap<Integer, Integer> sm,
            NavigableMap<Integer, Integer> nm
    ) {
        if (StreamSupport.stream(i.spliterator(), false).count() != st.size()) {
            return false;
        }
        if (st.size() != ns.size()) {
            return false;
        }
        if (ns.size() != m.size()) {
            return false;
        }
        if (m.size() != sm.size()) {
            return false;
        }
        if (sm.size() != nm.size()) {
            return false;
        }
        return true;
    }

    /**
     * Should create TreeSet without any modifications as T extends Number is not Comparable
     */
    public <T extends Number> boolean testTreeSetWithoutComparable(NavigableSet<T> set) {
        if (set.size() > 5) {
            return true;
        }
        return false;
    }

    /**
     * Should create TreeSet with modifications as Integer is Comparable
     */
    public boolean testTreeSetWithComparable(NavigableSet<Integer> set) {
        if (set.size() > 5) {
            return true;
        }
        return false;
    }

    static class ConcreteList<T> extends LinkedList<T> {}

    /**
     * Should create concrete class
     */
    public boolean testConcreteCollectionIsCreated(ConcreteList<?> list) {
        if (list.size() > 5) {
            return true;
        }
        return false;
    }

    static class ConcreteMap<K, V> extends HashMap<K, V> { }

    /**
     * Should create concrete class
     */
    public boolean testConcreteMapIsCreated(ConcreteMap<?, ?> map) {
        if (map.size() > 5) {
            return true;
        }
        return false;
    }
}
```

# Checklist:

- [x] No new warnings


utbot-fuzzers/src/main/kotlin/org/utbot/fuzzer/providers/RecursiveModelProvider.kt
```diff
@@ -31,87 +34,86 @@ data class ModelConstructor(
...
  * Note that if [modelProviderForRecursiveCalls] has instances of [RecursiveModelProvider] then this provider will use
- * their copies created by [createNewInstance] rather than themselves (this is important because we have to change some
+ * their copies created by [newInstance] rather than themselves (this is important because we have to change some
  * properties like [recursionDepthLeft], [totalLimit], etc.)
- *
  * @property fallbackProvider provider that will be used instead [modelProviderForRecursiveCalls] after reaching maximum recursion level
- *
- * @property totalLimit maximum number of parameters produced by this provider
- *
+ * @property totalLimit maximum number of values produced by this provider
```
volivan239, line 40: Currently this is not true: model provider can produce up to `branchingLimit * totalLimit` models. I also think that this (combined with the fact that we no longer reduce `totalLimit` for recursive instances) may reduce the variety of generated data if one of the providers generates too many models and models generated by others won't appear in the result. Needs to be discussed

volivan239 (changes requested):
Looks like a powerful change, but I believe there still may be some room for improving variety of created models, needs discussion.

utbot-framework/src/main/kotlin/org/utbot/fuzzer/FuzzerFunctions.kt
```diff
@@ -292,3 +298,13 @@ private fun sootIfToFuzzedOp(unit: JIfStmt) = when (unit.condition) {
...
 
 private fun nextDirectUnit(graph: ExceptionalUnitGraph, unit: Unit): Unit? = graph.getSuccsOf(unit).takeIf { it.size == 1 }?.first()
+
+fun toFuzzerType(type: Type): FuzzedType {
+    return when (type) {
+        is WildcardType -> type.upperBounds.firstOrNull()?.let(::toFuzzerType) ?: FuzzedType(objectClassId)
+        is TypeVariable<*> -> type.bounds.firstOrNull()?.let(::toFuzzerType) ?: FuzzedType(objectClassId)
+        is ParameterizedType -> FuzzedType((type.rawType as Class<*>).id, type.actualTypeArguments.map { toFuzzerType(it) })
+        is Class<*> -> FuzzedType(type.id)
+        else -> error("Unknown type: $type")
```
sergeypospelov, line 308: Looks like you forgot `GennericArrayType`. Consider this method:
```java
public <T extends Integer>int get(List<T[]> arr) {
    int x = 0;
    if (arr.size() > 0) {
        x = arr.get(0)[0];
    }
    return x;
}
```

sergeypospelov (changes requested):

Issue #948, opened 2022-09-16
Incorrect property access rendering in Kotlin


volivan239:
**Description**

By default, Kotlin generates a private field and getter/setter for each property into bytecode. However, you can't use these getters/setters directly from Kotlin code, instead you should direct access syntax (like direct accessing field in Java). However, code produced by UTBot uses these setters/getters which lead to compilation error.

**To Reproduce**

Launch action on the following class:
```Kotlin
class Sample {
    var x: Int = 3;
    fun f(): Int {
        return x;
    }
}
```

**Expected behavior**

Tests are generated correctly (in particular, access to field is written like `sample.x = ...`)

**Actual behavior**

Generated tests are not compilable.

**Visual proofs (screenshots, logs, images)**

Among others, the following test is produced:
```Kotlin
@Test
fun testF1() {
    val sample = Sample()
    sample.setX(-255)

    val actual = sample.f()

    assertEquals(-255, actual)
}
```

volivan239:
Seems like a duplicate of #496

PR #1002, opened 2022-09-23
Replace Kotlin getters/setters with property access in codegen #496

### LINKED ISSUE #496: Compilation errors in code with setters generated on Kotlin
**Description**

Create Kotlin class

```kotlin
class A {
    var a: Int = 0;
}
```

Create Java class

```java
public class D {
    public boolean foo(A a) {
        return a.getA() > 0;
    }
}
```

and create tests for it. Assemble models are used in tests generation

**Expected behavior**

Generated tests code is compilable.

**Actual behavior**

The following code is generated

```kotlin
@Test
@DisplayName("foo: AGetA -> return a.getA() > 0")
fun testFoo_AGetAGreaterThanZero() {
    val d = D()
    val a = A()
    a.setA(1)
    
    val actual = d.foo(a)
    
    assertTrue(actual)
    }
```

image

volivan239:
# Description

Now codegen properly treats Kotlin's autogenerated getters/setters and uses direct property access instead of them.

Kotlin properties are treated as fields. However, instead of `FieldId.isAccessibleFrom` which is now private, methods `FieldId.canBeReadFrom` and `FieldId.canBeSetFrom` should be used. First one behaves for Java code in the same way as `isAccessibleFrom` did before, but for Kotlin it also returns true if there is an accessible getter. Behavior of `FieldId.canBeSetFrom` hasn't been changed for Java, but for Kotlin it now also returns true if there is accessible setter/getter pair. See [docs](https://kotlinlang.org/docs/java-interop.html#getters-and-setters) for more details.

These two methods guarantee that codegen produces code with field access where it is possible. However, some calls to getters/setters may be produced by `AssembleModelGenerator`, so this PR also adds `replaceCgExecutableCallWithFieldAccessIfNeeded` to replace such calls with direct accesses.

Fixes #496

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Works correctly on examples from #496 and #948

- [x] No new warnings


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
```diff
@@ -212,7 +217,12 @@ internal class CgVariableConstructor(val context: CgContext) :
                     instance[statementModel.fieldId] `=` declareOrGet(statementModel.fieldModel)
                 }
                 is UtExecutableCallModel -> {
-                    +createCgExecutableCallFromUtExecutableCall(statementModel)
+                    val call = createCgExecutableCallFromUtExecutableCall(statementModel)
+                    val callOrAccess: CgStatement = replaceCgExecutableCallWithFieldAccessIfNeeded(call)
+                    if (callOrAccess is CgExecutableCall)
```
EgorkaKulikov, line 222: Improve this code fragment to make it more readable, please

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
```diff
@@ -236,6 +246,7 @@ internal class CgVariableConstructor(val context: CgContext) :
         val initExpr = if (isPrimitiveWrapperOrString(type)) {
             cgLiteralForWrapper(params)
         } else {
+            // TODO: if instantiation chain could be a setter call, we need to replace it in Kotlin
```
EgorkaKulikov, line 249: We have an instantiation call, now instantiation chain right now. Anyway, I can't imagine a situation when the top-level instantiation call is a setter.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
```diff
@@ -261,6 +272,23 @@ internal class CgVariableConstructor(val context: CgContext) :
         return cgCall
     }
 
+    private fun replaceCgExecutableCallWithFieldAccessIfNeeded(call: CgExecutableCall): CgStatement {
+        if (call !is CgMethodCall || context.codegenLanguage != CodegenLanguage.KOTLIN)
```
EgorkaKulikov, line 276: For me it seems better to write code-specific logic in a way like
```
when (context.codegenlangauge) {
is Java -> return call
is Kotlin -> do some specific logic
}
```

But I do not insist on it.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
```diff
@@ -261,6 +272,23 @@ internal class CgVariableConstructor(val context: CgContext) :
...
+            return call
+
+        val caller = call.caller ?: return call
+
+        caller.type.fieldThisIsSetterFor(call.executableId)?.let {
+            return CgAssignment(caller[it], call.arguments.single())
```
EgorkaKulikov, line 282: Why don't you require that arguments contains single item here as in line 285?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
```diff
@@ -261,6 +272,23 @@ internal class CgVariableConstructor(val context: CgContext) :
...
+        }
+        caller.type.fieldThisIsGetterFor(call.executableId)?.let {
+            require(call.arguments.isEmpty())
```
EgorkaKulikov, line 285: require is more useful with a message with failure description

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/ClassIdUtil.kt
```diff
@@ -28,4 +31,16 @@ infix fun ClassId.isAccessibleFrom(packageName: String): Boolean {
...
     }
-}
\ No newline at end of file
+}
+
+/**
+ * Returns field of [this], such that [methodId] is a getter for it (or null if methodId doesn't represent a getter)
+ */
+internal fun ClassId.fieldThisIsGetterFor(methodId: MethodId): FieldId? =
+    allDeclaredFieldIds.firstOrNull { !it.isStatic && it.getter == methodId }
```
EgorkaKulikov, line 40: May be `singleOrNull`? Same for the second method.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/ClassIdUtil.kt
```diff
@@ -28,4 +31,16 @@ infix fun ClassId.isAccessibleFrom(packageName: String): Boolean {
...
+
+/**
+ * Returns field of [this], such that [methodId] is a setter for it (or null if methodId doesn't represent a setter)
+ */
+internal fun ClassId.fieldThisIsSetterFor(methodId: MethodId): FieldId? =
```
EgorkaKulikov, line 45: I suggest the name `fieldThatIsSetWith`.
Current name means that field is a setter isself, it seems strange.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/visitor/UtilMethods.kt
```diff
@@ -343,7 +343,7 @@ private fun createInstance(visibility: Visibility, language: CodegenLanguage): S
         }
         CodegenLanguage.KOTLIN -> {
             """
-            ${visibility by language}fun createInstance(className: String): kotlin.Any? {
+            ${visibility by language}fun createInstance(className: String): kotlin.Any {
```
EgorkaKulikov, line 346: To be honest, I don't understand this change, but can believe that is is possible :)

EgorkaKulikov (approved):

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
volivan239: Should be clearer now. Also, changed return type of `replaceCgExecutableCallWithFieldAccessIfNeeded` to be nullable and added docs to it

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
volivan239: Thanks for advice, deleted the question

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
volivan239: Fixed here, didn't check for other places

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
volivan239: I do -- at `call.arguments.single()`....

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgVariableConstructor.kt
volivan239: Added message

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/util/ClassIdUtil.kt
volivan239: Renamed this and the method below

Issue #1024, opened 2022-09-28
Incorrect reflection call leads to compilation errors


EgorkaKulikov:
**Description**

Try to generate tests for the following class, set timeout at least to 60s.

```java
import org.antlr.runtime.tree.TreeVisitor;
import org.antlr.runtime.tree.TreeVisitorAction;
import org.antlr.v4.parse.GrammarASTAdaptor;
import org.antlr.v4.tool.ast.GrammarAST;

public class Grammarin {
    public void setGrammarPtr(final org.antlr.v4.tool.Grammar g, GrammarAST tree) {
        if (tree != null) {
            TreeVisitor v = new TreeVisitor(new GrammarASTAdaptor());
            v.visit(tree, new TreeVisitorAction() {
                public Object pre(Object t) {
                    ((GrammarAST) t).g = g;
                    return t;
                }

                public Object post(Object t) {
                    return t;
                }
            });
        }
    }
}
```

The following code i generated:

```java
    @Test
    @DisplayName("setGrammarPtr: -> tree != null : True")
    public void testSetGrammarPtr_TreeNotEqualsNull_2() throws ClassNotFoundException, NoSuchFieldException, IllegalAccessException {
        Grammarin grammarin = new Grammarin();
        GrammarRootAST tree = new GrammarRootAST(null);
        tree.token = null;
        Class baseTreeClazz = Class.forName("org.antlr.runtime.tree.BaseTree");
        Field childrenField = baseTreeClazz.getDeclaredField("children");
        childrenField.setAccessible(true);
        childrenField.get(tree) = new ArrayList();

        grammarin.setGrammarPtr(null, tree);
    }
```

Line with getter leads to the compilation error
```
childrenField.get(tree) = new ArrayList(); - variable expected
```

PR #1037, opened 2022-09-29
UtBotFieldModificatorsSearcher fixed for default package and public fields

EgorkaKulikov:
# Description

Searcher of field modificators fixed for default package: in the previous implementation it considered all modificators appropriate if they are located in a default package.
Also public fields validation corrected.

Fixes # ([1024](https://github.com/UnitTestBot/UTBotJava/issues/1024))

## Type of Change

# How Has This Been Tested?

## Automated Testing

UtBotFieldModificatorsTest

## Manual Scenario 

Repeat the scenario described in the mentioned feature with fuzzer and concrete executor turned off.


volivan239 (approved):
I think this is OK as a temporary hack, please, add a comment with TODO about it.

EgorkaKulikov:
Replaced with a nornal fix, not like this.

Issue #1054, opened 2022-10-03
IntelliJ Plugin memory leakage


alisevych:
**Description**

Plugin memory leakage in IDEA is detected. 
There are many CgContext Objects created in CgComponents  - but it seems that they are not removed.

**To Reproduce**

1. Install UTBot plugin on IntelliJ Idea 2022.1.4
2. Use plugin to generate tests
3. Do not close IDEA and generate tests with UTBot several times

**Expected behavior**

CgContext Objects are supposed to be destroyed after usage.

**Actual behavior**

There are many CgContext Objects in Profiler presetn - after several UTBot test generation runs.
There are only getOrPut methods in CgComponents.
After a couple of days running IDEA with UTBot  - all 2 GB are used. IDEA is running out of memory.

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.1.4

**Additional context**

Discovered by @Markoutte

PR #1057, opened 2022-10-03
Try to fix memory leak in CgComponents

EgorkaKulikov:
# Description

The following chain of dependencies leads us to the memory leaks.

- we have an OBJECT `CgComponents'
- it has a map called `fieldStateManagers` that is never cleaned
- each key of type `CgContext` contains `allExecutions` field, that requires much memory

Actually, other maps have the same key, so they need to be replaced with better design solutions too. 

Fixes # ([1054](https://github.com/UnitTestBot/UTBotJava/issues/1054))

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Performance testing on the similar examples that were used to reproduce the issue.


Damtev (approved):
LGTM

PR #1074, opened 2022-10-04
Select Kotlin test src root by default when generating for Kotlin #949

### LINKED ISSUE #949: Change default test source root for Kotlin classes
**Description**

When launching action on a class from Kotlin, the default test source root that is selected in UI is `../src/test/java` even if `../src/test/kotlin` exists.

**Expected behavior**

 I believe that the kotlin tests source root should be selected by default in this case.

**Potential alternatives**

If there is no kotlin tests source root in the project, it would be nice to offer user to create it automatically (currently this isn't done, at least if there is java tests source root)

volivan239:
# Description

Now, if project has both Java and Kotlin tests source roots, UTBot by default selects one that matches code generation language. 

Note that currently there is no proper way to check whether source root is for Java or for Kotlin, so we mainly decide this by folder name (see `org.utbot.intellij.plugin.ui.utils.ModuleUtilsKt#getExpectedLanguageForTests` and comment there).

Fixes #949

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Repeated scenario from #949 -- works as expected.

- [x] No new warnings

### DIFF
--- a/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/components/TestFolderComboWithBrowseButton.kt
+++ b/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/components/TestFolderComboWithBrowseButton.kt
@@ -18,8 +18,8 @@ import javax.swing.DefaultComboBoxModel
 import javax.swing.JList
 import org.jetbrains.kotlin.idea.util.projectStructure.allModules
 import org.utbot.common.PathUtil
-import org.utbot.framework.plugin.api.CodegenLanguage
 import org.utbot.intellij.plugin.models.GenerateTestsModel
+import org.utbot.intellij.plugin.ui.utils.TestSourceRoot
 import org.utbot.intellij.plugin.ui.utils.addDedicatedTestRoot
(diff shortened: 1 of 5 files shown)


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/ui/components/TestFolderComboWithBrowseButton.kt
```diff
@@ -58,7 +59,17 @@ class TestFolderComboWithBrowseButton(private val model: GenerateTestsModel) :
...
+            suggestedModules.flatMap { module ->
+                module.suitableTestSourceRoots(language)
+            }
+        }
+
+        // testRoots for default codegen language should go before other testRoots (this impacts default-selected test root item)
+        val testRoots = with (testRootsByLanguage.entries) {
+            filter { it.key == model.codegenLanguage } + filter { it.key != model.codegenLanguage }
+        }.flatMap { it.value }.toMutableList()
+
```
Vassiliy-Kudryashov, line 72: Please use sorting instead of filtering

Vassiliy-Kudryashov (approved):

PR #1089, opened 2022-10-04
We create temporary files a lot, we have to clean it either

Vassiliy-Kudryashov:
# Description

New cleaning stage was added for files outside "UTBot" directory

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Check tempDir, it may contain a lot of small files named like:
- Graph-vis***
- {some digits}-0
- byteBuddyAgent***.jar
- mockitoboot***.jar
We never clean it but we have to.

- [x] No new warnings


Vassiliy-Kudryashov:
[Windows never automatically cleans the %TEMP% directory by default. In Windows 10 you have to enable this feature in Settings, and with earlier versions you must delete the files yourself or use programs like Disk Cleanup or cCleaner.
](https://superuser.com/questions/296824/when-is-a-windows-users-temp-directory-cleaned-out#:~:text=Windows%20never%20automatically%20cleans%20the,to%20clean%20up%20after%20themselves.)

utbot-core/src/main/kotlin/org/utbot/common/FileUtil.kt
```diff
@@ -72,6 +72,19 @@ object FileUtil {
             val creationTime = Files.readAttributes(it.toPath(), BasicFileAttributes::class.java).creationTime()
             TimeUnit.MILLISECONDS.toDays(currentTimeInMillis - creationTime.toMillis()) > daysLimit
         }.forEach { it.deleteRecursively() }
+        File(tempDirectoryPath).listFiles { file ->
+            val name = file.name
+            if ((name.matches(Regex("Graph-vis\\d{10,}"))
+                        || name.matches(Regex("\\d{10,}-0"))
+                        || name.matches(Regex("byteBuddyAgent\\d{10,}\\.jar"))
+                        || name.matches(Regex("cmd-args\\d{10,}"))
+                        || name.matches(Regex("mockitoboot\\d{10,}\\.jar")))
```
Damtev, line 81: I suggest extracting these regular expressions out of the loop for a little performance improvement.

Also, are you sure all these files are always created by UnitTestBot? Shouldn't we check a creator?

utbot-core/src/main/kotlin/org/utbot/common/FileUtil.kt
```diff
@@ -72,6 +72,19 @@ object FileUtil {
...
+            ) {
+                val creationTime = Files.readAttributes(file.toPath(), BasicFileAttributes::class.java).creationTime()
+                return@listFiles (TimeUnit.MILLISECONDS.toDays(currentTimeInMillis - creationTime.toMillis()) > daysLimit)
```
Damtev, line 84: Guess we can introduce a function for this usage and the same usage above

Damtev (changes requested):

utbot-core/src/main/kotlin/org/utbot/common/FileUtil.kt
Vassiliy-Kudryashov: "Graph-vis" comes from GraphVix.kt:32 -> moved inside UTBot directory
"cmd-args" comes from Api.kt:1317 -> moved inside UTBot directory

Got rid of:
"\\d{10,}-0" not sure what is the source
"mockitoboot" comes from library code InlineDelegateByteBuddyMockMaker.java:123 but I cannot see new ones anymore
"byteBuddyAgent" comes from library we use via mockito but i cannot see new ones either

utbot-core/src/main/kotlin/org/utbot/common/FileUtil.kt
Vassiliy-Kudryashov: Done

Damtev (approved):
LGTM

PR #1108, opened 2022-10-06
Make snapshot version calendar-dependant

Vassiliy-Kudryashov:
# Description

Current time affects snapshot version in format YYYY.MM like 2022.10

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

call "runIde" > Go to IDE Settings > check UnitTestBot plugin version, it should look like 2022.10-SNAPSHOT

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import java.text.SimpleDateFormat
 import org.gradle.api.JavaVersion.VERSION_11
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -9,7 +10,7 @@ val coroutinesVersion: String by project
 val collectionsVersion: String by project
 val junit5Version: String by project
 
-version = semVer ?: "1.0-SNAPSHOT"
+version = semVer ?: "${SimpleDateFormat("YYYY.MM").format(System.currentTimeMillis())}-SNAPSHOT"
```
denis-fokin, line 13: Please create a separate value variable for it and add a comment that this value is usually computed the similar way on CI.

Vassiliy-Kudryashov: Done

PR #1185, opened 2022-10-19
Dynamic invoke for string concatenation

### LINKED ISSUE #386: Support for dynamic invoke in the engine
**Description**

Engine currently does not support the `dynamicinvoke` bytecode instruction. It seems not to be very important for Java 1.8, but in is much widely used in Java 11.

An important case is string concatenation: starting from Java 9, it is implemented using `dynamicinvoke`. As a result, the plugin fails to generate tests for string concatenation when the code is compiled using JDK 9 or later (the primary concern is JDK 11).

**Expected behavior**

Tests are correctly generated for the code that is compiled with `dynamicinvoke`.

**Environment**

JDK 9 or later.

**Potential alternatives**

Possible approaches to support `dynamicinvoke` are under investigation.

sergeypospelov:
# Description

- Investigated the way of implementing `dynamicInvoke` bytecode instruction in a common case. In a few words, we need to symbolically analyze bootstrap methods, which is incredibly hard in known cases, since it operates with combinators and includes many native calls.
- Implemented a special case of `dynamicInvoke` -- `java.lang.invoke.StringConcatFactory` methods.
- Added a new tests covering the string concatenation mechanism in Java 11.

Partially fixes #386

## Type of Change

# How Has This Been Tested?

## Automated Testing

New tests: `org.utbot.examples.strings.StringConcatTest`

- [x] No new warnings


utbot-framework/src/main/kotlin/org/utbot/engine/ConstructedSootMethods.kt
```diff
@@ -169,4 +172,122 @@ fun unfoldMultiArrayExpr(assignStmt: JAssignStmt): ExceptionalUnitGraph {
...
+
+    // helper function for appending constants to delayedString
+    fun appendDelayedStringIfNotEmpty() {
+        if (delayedString.isEmpty()) {
+            return
+        }
+
+        val type = STRING_TYPE
+
+        val appendSootMethod = sbSootClass.getMethod("append", listOf(type), sbSootClass.type)
```
CaelmBleidd, line 234: This method here and in the cycle below could be extracted into a variable only once

utbot-framework/src/main/kotlin/org/utbot/engine/DynamicInvokeResolver.kt
```diff
@@ -0,0 +1,114 @@
...
+            constants
+        )
+
+        val invocationTarget = InvocationTarget(
+            null,
+            sootMethod
+        )
+
+        return Invocation(
+            null,
```
CaelmBleidd, line 108: const without a name

utbot-framework/src/main/kotlin/org/utbot/engine/JimpleCreation.kt
```diff
@@ -60,11 +72,10 @@ fun createSootMethod(
     argsTypes: List<Type>,
     returnType: Type,
     declaringClass: SootClass,
-    graphBody: JimpleBody
-) = SootMethod(name, argsTypes, returnType)
+    graphBody: JimpleBody,
+    isStatic: Boolean = true
+) = SootMethod(name, argsTypes, returnType, (if (isStatic) Modifier.STATIC else 0))
```
CaelmBleidd, line 77: redundant brackets

CaelmBleidd (approved):

Damtev (approved):
Code generation changes look good to me

PR #1192, opened 2022-10-19
Disable test generation for synthetic methods of data classes #1191

### LINKED ISSUE #1191: We shouldn't generate tests for synthetic methods of data classes
**Description**

Kotlin automatically generates some methods for data classes, however, we shouldn't test them unless they are overridden manually.

**To Reproduce**

Create data class with no methods and launch action on it:

```Kotlin
data class Datalass(val x: Int, val y: Int)
```

**Expected behavior**

Action is disabled

**Actual behavior**

Dialog appears with synthetic methods being proposed for testing

**Additional context**

The issue looks like a "Kotlin edition" of #457

volivan239:
# Description

Added filters so that autogenerated by Kotlin methods of data classes are not tested by plugin/cli/contestEstimator

Fixes #1191

## Type of Change

# How Has This Been Tested? 

## Manual Scenario 

Tested the behavior in plugin -- autogenerated methods are not added to memberlist (unless they are overridden).
In cli tests for autogenerated methods are also not generated.

- [x] No new warnings


utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
```diff
@@ -1,14 +1,18 @@
...
     if (method.classId.isEnum)
-        method.humanReadableName.substringBefore('(') in KnownSyntheticMethodNames.enumSyntheticMethodNames
-    else
-        false
+        return method.name in KnownSyntheticMethodNames.enumSyntheticMethodNames
+
+    if (method.classId.isData)
+        return KnownSyntheticMethodNames.dataClassSyntheticMethodNames.any { it.matches(method.name) }
+
+    return false
```
Damtev, line 14: `when` may be better here, I suppose

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
```diff
@@ -17,4 +21,14 @@ fun isKnownSyntheticMethod(method: ExecutableId): Boolean =
...
     /** List with names of enum methods that are autogenerated */
     val enumSyntheticMethodNames = listOf("values", "valueOf")
+
+    /** List with names of data classes methods that are autogenerated in Kotlin */
+    val dataClassSyntheticMethodNames = listOf(
+        "equals",
+        "hashCode",
+        "toString",
+        "copy",
+        "copy\\\$default",
```
Damtev, line 31: What is this method? Add some comments about it, please

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
```diff
@@ -17,4 +21,14 @@ fun isKnownSyntheticMethod(method: ExecutableId): Boolean =
 private object KnownSyntheticMethodNames {
     /** List with names of enum methods that are autogenerated */
     val enumSyntheticMethodNames = listOf("values", "valueOf")
+
+    /** List with names of data classes methods that are autogenerated in Kotlin */
```
Damtev, line 25: Mention that they are actually regexps, please

Damtev (changes requested):
Add a bit more docs, please

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
```diff
@@ -1,14 +1,18 @@
...
-fun isKnownSyntheticMethod(method: ExecutableId): Boolean =
+fun isKnownSyntheticMethod(method: ExecutableId): Boolean {
     if (method.classId.isEnum)
-        method.humanReadableName.substringBefore('(') in KnownSyntheticMethodNames.enumSyntheticMethodNames
-    else
-        false
+        return method.name in KnownSyntheticMethodNames.enumSyntheticMethodNames
+
+    if (method.classId.isData)
+        return KnownSyntheticMethodNames.dataClassSyntheticMethodNames.any { it.matches(method.name) }
```
Damtev, line 12: Am I right that CLI will not generate test methods for overridden "synthetic" methods in data classes (`equals`, for example)?

Damtev (changes requested):
What about overridden methods in data classes for CLI?

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
volivan239: Is it possible here? `isEnum` and `isData` are just properties...

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
volivan239: It is JVM-synthetic and needed when copy is called with default parameters, after last commit it is not needed here.

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
volivan239: Fixed, also changed the property name for more clarity

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
volivan239: Yes, and it seems difficult (or impossible) to check if the method is not compiler-generated but overridden by user (but if you know the way, I would appreciate hearing it)

utbot-framework/src/main/kotlin/org/utbot/framework/util/SyntheticMethods.kt
Damtev: Just a simple one, like
```kotlin
when {
    method.classId.isEnum -> ...
    method.classId.isData -> ...
    else -> ...
}
```

volivan239: Done

Damtev (approved):

PR #1228, opened 2022-10-25
Map don't work without concrete

### LINKED ISSUE #436: HashMap with Enum as key - test generation fails
**Description**

Test generation fails for a method with HashMap using Enum as key.

**To Reproduce**

1. IntelliJ IDEA with UTBotJava plugin installed
2. A project with JDK 8/11 is opened
3. Add the following class:

```
public class MapWithEnum {

    public void createMapWithEnum() {
        Map<WorkDays, Integer> map = new HashMap<>();
        map.put(WorkDays.Monday, 112);
        map.put(WorkDays.Tuesday, 354);
        map.put(WorkDays.Friday, 567);
        map.remove(WorkDays.Tuesday);
    }

    public enum WorkDays {
        Monday,
        Tuesday,
        Wednesday,
        Thursday,
        Friday
    }
}
```

4. Run Generating tests with UTBot... for this class

**Expected behavior**

Tests are supposed to be generated.

**Actual behavior**

"Failed to generate unit tests for class MapWithEnum" error popup is shown in IDEA.
It is shown immediately after test generation is started.
Map.remove() method is crucial for the issue - without it generation passes.

Map with Enum as key

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.1.3
checked on JDK 8 and 11, Gradle, Maven projects - same error

**Additional context**

Adding some input parameters to the method helps - fuzzer is generating tests.

Map.remove() method is crucial for the issue - without it generation passes.

Map with Enum as values - test generation passes.

CaelmBleidd:
# Description

At some moment we discovered that maps do not work as they suppose: sometimes concrete could fix it (for example, when a map was not a parameter), sometimes could not. This request fixes two particular problems: error in type indices for `select` operation and possible aliasing between nested arrays.

A problem with aliasing could be seen while debugging org.utbot.examples.collections.Maps#putInMapFromParameters. Without this fix it doesn't work for an empty map: `touched` and `values` might have the same address, and if we make two `put` operations, the second one will cause a problem -- memory update for keys will rewrite values as well, and we'll have wrong result values. With one put or with a map that was created during the execution everything worked fine (after the fix with types)

Fixes #436 

## Type of Change

# How Has This Been Tested?

## Automated Testing

org.utbot.examples.collections.Maps#createMapWithEnum
org.utbot.examples.collections.Maps#createMapWithString
org.utbot.examples.collections.Maps#containsKeyAndPuts
org.utbot.examples.collections.Maps#putInMapFromParameters
org.utbot.examples.collections.Maps#mapPutAndGet

## Manual Scenario 

Didn't check manually. 

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/ArrayObjectWrappers.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/ArrayObjectWrappers.kt
@@ -215,13 +215,17 @@ class RangeModifiableUnlimitedArrayWrapper : WrapperInterface {
                 createArray(addr, valueType, useConcreteType = false)
             }
 
-            listOf(
-                MethodResult(
-                    SymbolicSuccess(resultObject),
-                    typeRegistry.typeConstraintToGenericTypeParameter(addr, wrapper.addr, i = TYPE_PARAMETER_INDEX)
-                        .asHardConstraint()
-                )
-            )
+            val typeIndex = wrapper.asWrapperOrNull?.getOperationTypeIndex
+                ?: error("Wrapper was expected, got $wrapper")
+            val typeConstraint = typeRegistry.typeConstraintToGenericTypeParameter(
+                addr,
+                wrapper.addr,
+                i = typeIndex
+            ).asHardConstraint()
+
+            val methodResult = MethodResult(SymbolicSuccess(resultObject), typeConstraint)
+
+            listOf(methodResult)
         }
 
     @Suppress("UNUSED_PARAMETER")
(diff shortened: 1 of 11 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/ArrayObjectWrappers.kt
```diff
@@ -215,13 +215,16 @@ class RangeModifiableUnlimitedArrayWrapper : WrapperInterface {
...
             }
 
-            listOf(
-                MethodResult(
-                    SymbolicSuccess(resultObject),
-                    typeRegistry.typeConstraintToGenericTypeParameter(addr, wrapper.addr, i = TYPE_PARAMETER_INDEX)
-                        .asHardConstraint()
-                )
-            )
+            val typeIndex = wrapper.asWrapperOrNull?.getOperationTypeIndex ?: error("Unexpected wrapper $wrapper")
```
Damtev, line 218: Unexpected not wrapper?

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -3534,7 +3534,7 @@ class Traverser(
 
         // toplevel method
         // TODO: investigate very strange behavior when some constraints are not added leading to failing CodegenExampleTest::firstExampleTest fails
-        val terminalExecutionState = environment.state.copy(
+         val terminalExecutionState = environment.state.copy(
```
Damtev, line 3537: By mistake?

utbot-sample/src/main/java/org/utbot/examples/collections/Maps.java
```diff
@@ -25,6 +27,35 @@ String mapToString(long startTime, int pageSize, int pageNum) {
         return params.toString();
     }
 
+    Integer mapPutAndGet() {
```
Damtev, line 30: Looks like this method should suppress the same warning as methods below do

sergeypospelov (approved):

Issue #1231, opened 2022-10-25
Add symbolic executions with force mocks in parameterized tests mode


alisevych:
**Description**

If symbolic executions require force mocking they can't be added to the resulting test class as parametrized tests.
Currently they are just skipped. So the test class generated in parameterized mode do not include all the test cases generated without parameterization.

It is suggested to add symbolic executions with force mocks as standard tests.
Just as has been done for Fuzzer executions in #1137 

**Expected behavior**

The solution is to do the following in parameterized mode:

- create parameterized tests for symbolic executions without force mocks
- create standard tests for symbolic executions with force mocks

**Potential alternatives**

The solution is not ideal. There is the mix of parametrized and standard tests in one class.
The alternative - as it is working now - is to skip tests that require force mocking.

**Context**

Discussed with @EgorkaKulikov and @sofurihafe

alisevych:
@denis-fokin FYI

PR #1252, opened 2022-10-27
Generate standard tests for symbolic executions with mocking in parametrized mode

sofurihafe:
# Description

This PR adds more coverage by generating previously ignored test cases. Before, we skipped executions with mocking when generating parameterized tests. Now, we generate standard test in these cases.

This behavior is similar to our way of processing fuzzer executions (that may also contain force-mocking) in parametrized mode.

Fixes # ([1231](https://github.com/UnitTestBot/UTBotJava/issues/1231))

## Type of Change

# How Has This Been Tested?

## Automated Testing

UTBot samples.

## Manual Scenario 

Run *MockFinalClassTest* for *useFinalClassMethod()* method. Verify that now there are three generated tests instead of two (as it was when we excluded symbolic executions with mocking in them).

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgTestClassConstructor.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgTestClassConstructor.kt
@@ -39,6 +39,7 @@ import org.utbot.framework.plugin.api.ClassId
 import org.utbot.framework.plugin.api.MethodId
 import org.utbot.framework.plugin.api.UtExecutionSuccess
 import org.utbot.framework.plugin.api.UtMethodTestSet
+import org.utbot.framework.plugin.api.UtSymbolicExecution
 import org.utbot.framework.plugin.api.util.description
 import org.utbot.framework.plugin.api.util.humanReadableName
 import org.utbot.fuzzer.UtFuzzedExecution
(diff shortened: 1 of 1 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgTestClassConstructor.kt
```diff
@@ -211,16 +212,28 @@ internal class CgTestClassConstructor(val context: CgContext) :
...
+        testSet.executions
+            .filterIsInstance<UtSymbolicExecution>()
+            .filter { it.containsMocking }
+            .forEach { execution ->
+                testCaseTestMethods += methodConstructor.createTestMethod(methodUnderTest, execution)
+            }
 
         regions += CgSimpleRegion(
-            "FUZZER: EXECUTIONS for method ${methodUnderTest.humanReadableName}",
+            "EXECUTIONS for method ${methodUnderTest.humanReadableName}",
```
EgorkaKulikov, line 236: May be something like 
`EXECUTIONS for method ${methodUnderTest.humanReadableName} that cannot be presented in parametrized mode` ?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/model/constructor/tree/CgTestClassConstructor.kt
```diff
@@ -211,16 +212,28 @@ internal class CgTestClassConstructor(val context: CgContext) :
             )
         }
 
+        val testCaseTestMethods = mutableListOf<CgTestMethod>()
```
EgorkaKulikov, line 215: Could you move all this logic to a private method, please?

EgorkaKulikov (approved):

PR #1293, opened 2022-11-02
Stabilize the behavior of ConstructorAnalyzer

volivan239:
# Description

Removed `vistiedConstructors` from `ConstructorAnalyzer.analyze()` method. The intention of that container was to prevent from infinite recursion and to increase performance when analyzing chain of constructors. However, it lead to possible errors if the outer method was called more then once. Also, infinite recursion isn't actually possible since recursive constructor calls are forbidden both in Java and Kotlin. Performance changes seems to be miserable. 

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Tested on the following class -- previously incorrect behavior of `ConstructAnalyzer` lead to unnecessary reflection used for `expected` models, after these changes no reflection is used.

```Kotlin
class NullabilityIssuesExample(val x: Int) {
    fun copyWithX(new_x: Int?): NullabilityIssuesExample? {
        if (new_x == null)
            return null
        return NullabilityIssuesExample(x + new_x)
    }
}
```

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/modifications/ConstructorAnalyzer.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/modifications/ConstructorAnalyzer.kt
@@ -108,18 +108,11 @@ class ConstructorAnalyzer {
         return jimpleLocal.name.first() != '$'
     }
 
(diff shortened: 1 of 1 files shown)


volivan239:
Doesn't work and needs smarter approach -- see #1353

Issue #1204, opened 2022-10-20
Empty region names and missing javadoc for `Sort.arrayCopy`


tyuldashev:
**Description**

**To Reproduce**

```java
    public int[] arrayCopy() {
        int[] fst = {1, 2, 3};
        int[] snd = {4, 5, 6};

        System.arraycopy(fst, 0, snd, 0, 3);
        return snd;
    }
```

1. Open [`algorithms/Sort.java`](https://github.com/UnitTestBot/UTBotJava/blob/main/utbot-sample/src/main/java/org/utbot/examples/algorithms/Sort.java) sample from UTBot project 
2. Or use code sample from above.
3. In plugin settings set generation method to 95% Symbolic
4. Use plugin to generate tests
5. Open the generated test

**Expected behavior**

Every test is placed in region with name and has proper java doc description

**Actual behavior**

Quite many regions have no names and some tests have no javadocs

**Visual proofs (screenshots, logs, images)**

Here is one of generated test:
```java
public class SortTest {
    ///region Test suites for executable org.example.Sort.arrayCopy

    ///region

    @Test
    public void testArrayCopy1() {
        Sort sort = new Sort();

        int[] actual = sort.arrayCopy();

        int[] expected = {1, 2, 3};
        assertArrayEquals(expected, actual);
    }

...
```

amandelpie:
@tyuldashev the default behaviour - do not generate summaries/regions if something goes wrong (it should be logged by logger.error to the log). We are not guarantee at this moment that all tests will have regions and other summaries.

Sometimes it's a developer bug in summaries, sometimes it's the limitation of summaries functionality.

I suggest or copy the reason of crash from logs (if it's the same it's one bug if is different—there are a few bugs) or copy generated results (with missed summaries)

amandelpie:
On my laptop looks like it's reproduced and related to the problem #1227

amandelpie:
@tyuldashev @alisevych Looks like it was fixed by #1227 closed earlier, it's not reproduced on my side

amandelpie:
@tyuldashev @alisevych please reproduce and close, the bug is not reproduced on the master and closed earlier

For example, the following test is generated

```
public final class SortTest {
    ///region Test suites for executable custom.Sort.arrayCopy

    ///region FUZZER: SUCCESSFUL EXECUTIONS for method arrayCopy()

    /**
     * @utbot.classUnderTest {@link Sort}
[...]

PR #1303, opened 2022-11-03
Fixed FileNotFoundExecption during JavaParser work

### LINKED ISSUE #1227: Summaries are not generated due to FileNotFoundException
**Description**

Summaries are not generated due to FileNotFoundException  thrown in the [SourceCodeParser.kt:45](https://github.com/UnitTestBot/UTBotJava/blob/main/utbot-summary/src/main/kotlin/org/utbot/summary/ast/SourceCodeParser.kt#L45)

The error message is the following: "C:\Users\<username>\.jdks\corretto-11.0.16.1\lib\src.zip!\java.base\java\lang\Math.java (??????? ?? ??????? ????? ????????? ????)"

Need to say that the Math.java file is exist as a Zip entry in the archive with sources mentioned above

**To Reproduce**

1. Run runIde Gradle task
2. Copy to the open project the mentioned below snippet of code
3. Use plugin to generate tests for DoubleFunctions::hypo method

```
public class DoubleFunctions {
    public double hypo(double a, double b) {
        return Math.sqrt(Math.pow(a, 2) + Math.pow(b, 2));
    }

    @SuppressWarnings("ManualMinMaxCalculation")
    public double max(double a, double b) {
        return a > b ? a : b;
    }

    public double circleSquare(double r) {
        if (r < 0 || Double.isNaN(r) || r > 10000) {
            throw new IllegalArgumentException();
        }
        double square = Math.PI * r * r;
        if (square > 777.85) {
            return square;
        } else {
            return 0;
        }
    }

    public int numberOfRootsInSquareFunction(double a, double b, double c) {

        double result = b * b - 4 * a * c;
        if (result > 0) {
            return 2;
        } else if (result == 0) {
            return 1;
        }
        return 0;
    }
}
```

**Expected behavior**

Tests with summaries should be generated.

**Actual behavior**

Tests are generated without summaries.

image

**Environment**

The Coretto jdk 11.0.16.1 is installed and used in both, initial and debug IDEs

amandelpie:
# Description

Fixes #1227 #1204 

The main problem was the following: JavaParser tries to parse ZipEntry in zip archive with JDK sources.
ZipEntry is not a file.

To filter this situation, we decided to ignore these zipped sources from the standard library. In reality, we ignored it before @Domonion changes in the #1151 PR. 

We extended our scope and failed when faced with these issues.

The proposed solution helps to avoid situation with parsing of non-existing files and also improves the logging for some edge situations.

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

One of scenarios is described also in the ticket #1204 

### DIFF
--- a/utbot-summary/src/main/kotlin/org/utbot/summary/Summarization.kt
+++ b/utbot-summary/src/main/kotlin/org/utbot/summary/Summarization.kt
@@ -70,7 +70,10 @@ fun UtMethodTestSet.summarize(sourceFile: File?, searchDirectory: Path = Paths.g
 }
 
 fun UtMethodTestSet.summarize(searchDirectory: Path): UtMethodTestSet =
-    this.summarize(Instrumenter.adapter.computeSourceFileByClass(this.method.classId.jClass, searchDirectory), searchDirectory)
+    this.summarize(
+        Instrumenter.adapter.computeSourceFileByClass(this.method.classId.jClass, searchDirectory),
+        searchDirectory
+    )
 
 
 class Summarization(val sourceFile: File?, val invokeDescriptions: List<InvokeDescription>) {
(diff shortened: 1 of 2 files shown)


utbot-summary/src/main/kotlin/org/utbot/summary/Summarization.kt
```diff
@@ -395,10 +401,15 @@ private fun invokeDescriptions(testSet: UtMethodTestSet, searchDirectory: Path):
                 sootMethod.declaringClass.javaPackageName.replace(".", File.separator),
                 searchDirectory
             )
-            val ast = methodFile?.let {
-                SourceCodeParser(sootMethod, it).methodAST
+
+            if (methodFile != null && methodFile.exists()) {
```
denis-fokin, line 405: My personal preference is to use as few nulls as possible

Issue #1280, opened 2022-11-01
Do not suggest to test Kotlin protected methods


tyuldashev:
**Description**

UTBot suggests to generate tests for `protected` members of Kotlin class, but because they are visible only inside that class itself and its descendants the generated code cannot access such methods and fails at compilation.

**To Reproduce**

1. In UTBot settings set `Generated test language` to `Kotlin`
2. Invoke test generation for following class
```kotlin
class Calc {
    protected fun substract(a: Int, b: Int): Int {
        return a - b
    }
}
```

**Expected behavior**

UTBot either warn that there is NO methods it can test or create compilable code (if we can and want to access this method through reflection).

**Actual behavior**
UTBot allows to generate tests for that method and create uncompilable code.

**Visual proofs (screenshots, logs, images)**

![image](https://user-images.githubusercontent.com/10035839/199236428-c38fd27e-b24e-473e-967d-6e07df3a7d2f.png)

![image](https://user-images.githubusercontent.com/10035839/199237232-3932ef62-6bec-44d7-9726-b431d7064dcf.png)

EgorkaKulikov:
In current implementation we consider protected keyword in terms of Java.
May be we should improve `PsiClassHelper.filterTestableMethods`. Or suggest some more complicated separation between two protected terms.

PR #1373, opened 2022-11-15
Exclude protected kotlin methods from generate window

sofurihafe:
# Description

This PR makes protected kotlin methods being excluded from generate window, because we cannot access them as described in the following issue.

Fixes # ([1280](https://github.com/UnitTestBot/UTBotJava/issues/1280))

## Type of Change

# How Has This Been Tested?

## Automated Testing

utbot-samples

## Manual Scenario 

Try to generate the kotlin test in plugin for the following code:
```kotlin
class SimpleClass {
    protected fun substract(a: Int, b: Int): Int {
        return a - b
    }
}
```
Verify that there is no *substract* method in generate window.


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/PsiClassHelper.kt
```diff
@@ -23,6 +24,9 @@ private val PsiMember.isKotlinGetterOrSetter: Boolean
         return isGetter || isSetter
     }
 
+private val PsiMember.isKotlinAndProtected: Boolean
+    get() = this is KtLightMemberImpl<*> && this.hasModifierProperty(PsiModifier.PROTECTED)
```
volivan239, line 28: Minor: I think checking for `KtLightMember<*>` sounds a little bit more general than checking for `KtLightMemberImpl<*>`.

volivan239 (approved):
I think we should also exclude testing Kotlin protected members in cli (by similar checks in somewhere like `org.utbot.cli.GenerateTestsCommand#run`). However, right now even private methods are not filtered there, so it seems like another ticket

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/PsiClassHelper.kt
sofurihafe: Little accident. Changed to `KtLightMember<*>`.

Issue #1210, opened 2022-10-21
Enum class is addressed incorrectly


alisevych:
**Description**

When Enum class is created by className, it addresses not class but the Enum constant.

**To Reproduce**

1. Run the 'UTBotJava' project in IntelliJ Idea
2. Install [plugin build](https://github.com/UnitTestBot/UTBotJava/actions/runs/3296912146) or one of the latest builds from main. [Reproducing on more early build too](https://github.com/UnitTestBot/UTBotJava/actions/runs/3282086137)
3. Generate tests for utbot-sample/src/main/java/org/utbot/examples/enums
with No mocks, No parameterization.
4. Run the generated tests

**Expected behavior**

Assertions in generated tests are supposed to pass.

**Actual behavior**

There are assertion errors in tests - for every class in the package.
Please see the Class creation: `State$3` is addressing not the State enum, but the third constant in it.

**Visual proofs (screenshots, logs, images)**

~~~java
    ///region Test suites for executable org.utbot.examples.enums.ComplexEnumExamples.findState

    ///region FUZZER: SUCCESSFUL EXECUTIONS for method findState(int)

    @Test
    @DisplayName("findState: code > 0")
    public void testFindState() throws ClassNotFoundException, IllegalAccessException {
        ComplexEnumExamples complexEnumExamples = new ComplexEnumExamples();

        State actual = complexEnumExamples.findState(1);

        Class anonymousStateClazz = Class.forName("org.utbot.examples.enums.State$3");
        Object expected = getEnumConstantByName(anonymousStateClazz, "UNKNOWN");
        assertEquals(expected, actual);
    }
~~~

same for Symbolic execution:

```java
    ///region Test suites for executable org.utbot.examples.enums.State.findStateByCode

    ///region SYMBOLIC EXECUTION: SUCCESSFUL EXECUTIONS for method findStateByCode(int)

    /**
     * @utbot.classUnderTest {@link State}
     * @utbot.methodUnderTest {@link State#findStateByCode(int)}
     */
    @Test
    @DisplayName("findStateByCode: ")
    public void testFindStateByCode() throws ClassNotFoundException, IllegalAccessException {
        State actual = State.findStateByCode(255);

        Class anonymousStateClazz = Class.forName("org.utbot.examples.enums.State$1");
        Object expected = getEnumConstantByName(anonymousStateClazz, "OPEN");
        assertEquals(expected, actual);
    }

```

**Additional context**

[...]

Issue #617, opened 2022-07-30
No tests are generated for methods in enums with nested anonymous classes


dtim:
**Description**

No tests can't be generated by the symbolic engine for methods in enum classes whose values have corresponding anonymous classes.

Note: I am not currently sure that there are no cases where tests would be generated, this question has to be investigated. See #300 for a sample configuration for the contest estimator that will be useful for collecting data.

**To Reproduce**

Generate the test suite for the `State.findStateByCode()` method in the following code.
```
public enum State {
    OPEN(255) {
        @Override
        public String toString() {
            return "<open>";
        }
    },
    CLOSED(127) {
        @Override
        public String toString() {
            return "<closed>";
        }
    },
    UNKNOWN(0) {
        @Override
        public String toString() {
            return "<unknown>";
        }
    };

    private final int code;

    State(int code) {
        this.code = code;
    }
    
    public int getCode() {
        return code;
    }

    public static State findStateByCode(int code) {
        for (State state: values()) {
            if (state.getCode() == code) {
                return state;
            }
        }
        return UNKNOWN;
    }
}

```

**Expected behavior**

A non-empty test suite should be generated, no errors should be reported.

**Actual behavior**

No tests are generated, the error message is displayed: "UtBot failed to generate any test cases for class State".

**Environment**

This behavior does not depend on the specific test framework configuration and mocking settings. It can be reproduced in both Java 8 and Java 11, as well as on fuzzer settings.

**Additional context**

Enum constructor does not prevent test generation, anonymous nested classes do.

The issue is probably related to the limited support of anonymous classes in the code generator. It may be reproduced even with fuzzing turned on, and even with fuzzing only, so it seems that it is a codegen error. At the same time, anonymous class processing in the symbolic engine should be checked too.

CaelmBleidd:
Currently, it works fine
![image](https://user-images.githubusercontent.com/31047452/216070541-304d8301-6059-478c-8c65-f52e8fa9c49d.png)

PR #1408, opened 2022-11-22
Enum classes are instantiated correctly if some methods are overriden

EgorkaKulikov:
# Description

Sometimes enum constants override some methods of their enum
So we have an instance of anonymous class instead of enum, that can't be instantiated with `Class.forName`

Fixes # ([1210](https://github.com/UnitTestBot/UTBotJava/issues/1210))

## Type of Change

# How Has This Been Tested?

## Automated Testing

standard utbot-pipeline

## Manual Scenario 

standard enum regression testing

##Additional context

And advice from Dmitry Timofeev:


volivan239 (approved):
As `org.utbot.examples.enums.ComplexEnumExamplesTest#testFindState` is working correctly, it is now needed to think if #617 is closed by this.

Issue #1397, opened 2022-11-18
Enum instantiating with unnecessary reflection


alisevych:
**Description**

Some of tests for Enum generated by Symbolic execution are instantiating with unnecessary reflection.

**To Reproduce**

1. Run the `lrs` project in IntelliJ Idea
2. Install one of the latest plugin builds from main, [used 2022.11.546](https://github.com/UnitTestBot/UTBotJava/actions/runs/3494980756)
3. Use default settings
4. Use plugin to generate tests for `ThreeState` enum
5. Open the generated test

**Expected behavior**

Enums are supposed to be instantiated without reflection.

**Actual behavior**

There are several (not all) tests generated by Symbolic execution that use unneeded reflection.

**Visual proofs (screenshots, logs, images)**

Example:
~~~java
    ///region OTHER: ERROR SUITE for method mostPositive(java.lang.Iterable)

    @Test
    public void testMostPositive9() throws ClassNotFoundException, IllegalAccessException, NoSuchFieldException {
        ThreeState prevNO = ThreeState.NO;
        try {
            ThreeState no = ThreeState.NO;
            Class threeStateClazz = Class.forName("com.intellij.util.ThreeState");
            setStaticField(threeStateClazz, "NO", no);
            HashSet states = new HashSet();
            ThreeState threeState = ThreeState.YES;
            states.add(threeState);
            states.add(null);
            
            /* This test fails because method [com.intellij.util.ThreeState.mostPositive] produces [java.lang.NullPointerException]
                com.intellij.util.ThreeState.mostPositive(ThreeState.java:40) */
            ThreeState.mostPositive(states);
        } finally {
            setStaticField(ThreeState.class, "NO", prevNO);
        }
    }
    ///endregion
~~~

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.2.3
IntelliJ project, JDK 11

alisevych:
This bug is probably connected with
- #618

Issue #306, opened 2022-06-27
Static Enum values are needlessly overwritten


dtim:
**Description**

The code generator overwrites static Enum values with themselves in tests where an Enum constant is required.

**To Reproduce**

Generate a test suite for the `EnumExamples.isRedOrGreen` method below.

```
// Color.java
public enum Color {
    UNDEFINED,
    RED,
    YELLOW,
    GREEN
}

// EnumExamples.java
public class EnumExamples {
    public boolean isRedOrGreen(@NotNull Color color) {
        return color.equals(Color.YELLOW) || color.equals(Color.GREEN);
    }
}
```

**Expected behavior**

No static field assignment should take place in any of generated test cases.

Test cases could look like the following example:
```
    @Test
    @DisplayName("isRedOrGreen: return color.equals(Color.YELLOW) || color.equals(Color.GREEN) : False -> return color.equals(Color.YELLOW) || color.equals(Color.GREEN)")
    public void testIsRedOrGreen_ColorEqualsOrColorEquals() throws ClassNotFoundException, Exception {
        EnumExamples enumExamples = new EnumExamples();
        boolean actual = enumExamples.isRedOrGreen(Color.YELLOW);
        assertTrue(actual);
    }
```

**Actual behavior**

Each test case involves static field assignments with their original values.

**Visual proofs (screenshots, logs, images)**

```
    ///region SUCCESSFUL EXECUTIONS for method isRedOrGreen(enums.Color)

    /**
     * <pre>
     * Test returns from: {@code return color.equals(Color.YELLOW) || color.equals(Color.GREEN); }
     * </pre>
     */
    @Test
    @DisplayName("isRedOrGreen: return color.equals(Color.YELLOW) || color.equals(Color.GREEN) : False -> return color.equals(Color.YELLOW) || color.equals(Color.GREEN)")
    public void testIsRedOrGreen_ColorEqualsOrColorEquals() throws ClassNotFoundException, Exception {
        Color prevYELLOW = Color.YELLOW;
        try {
            Color yellow = Color.YELLOW;
            Class colorClazz = Class.forName("enums.Color");
            setStaticField(colorClazz, "YELLOW", yellow);
            EnumExamples enumExamples = new EnumExamples();

            boolean actual = enumExamples.isRedOrGreen(yellow);

            assertTrue(actual);
        } finally {
            setStaticField(Color.class, "YELLOW", prevYELLOW);
        }
    }

    /**
     * <pre>
[...]

PR #1416, opened 2022-11-23
Avoid remembering and restoring enum constants

EgorkaKulikov:
# Description

Usually we remember the state of a static variable before the main body of each generated test and restore this state at the end. But enum constants are such statics that do not require such actions.

Fixes # ([1397](https://github.com/UnitTestBot/UTBotJava/issues/1397))
Fixes # ([306](https://github.com/UnitTestBot/UTBotJava/issues/306))

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

- Scenarios described in both mentioned issues
- Standard enums regression


utbot-framework-test/src/test/kotlin/org/utbot/examples/enums/ComplexEnumExamplesTest.kt
```diff
@@ -74,6 +75,7 @@ class ComplexEnumExamplesTest : UtValueTestCaseChecker(
     }
 
     @Test
+    @Disabled("TODO: nested anonymous classes are not supported: https://github.com/UnitTestBot/UTBotJava/issues/617")
```
volivan239, line 78: I believe that the reason for disabling here differs from issue mentioned in 617, so maybe create an another issue?

volivan239 (approved):

PR #1429, opened 2022-11-25
Add hint for disabled fuzzing control #1411

### LINKED ISSUE #1411: Add hint for disabled fuzzing control
**Description**

In case useFuzzing=false in settings.properties we disable corresponding UI control. Such a behavior may be not obvious for a user and we have to provide some explanation hint do explain why the control is dsabled and how to enable it.

**To Reproduce**

Specify useFuzzing=false in {home}/.utbot/settings.properties

**Expected behavior**

Fuzzing ratio control disabled with an explanation hint

**Actual behavior**

Fuzzing ratio control is just disabled

java_1iIZVfECYk

Vassiliy-Kudryashov:
# Description

New comment line appears when fuzzing is disabled in settings.properties file

Fixes #1411 

## Type of Change

# How Has This Been Tested?

## Manual Scenario 

Disable fuzzing in settings.properties with the line `useFuzzing=false`
Check if comment appears under disabled fuzzing slider in in Settings > Tools > UtitTestBot

- [x] No new warnings

### DIFF
--- a/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/settings/SettingsWindow.kt
+++ b/utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/settings/SettingsWindow.kt
@@ -209,6 +209,11 @@ class SettingsWindow(val project: Project) {
                 addToRight(symLabel)
             }().constraints(CCFlags.growX)
         }
+        if (!UtSettings.useFuzzing) {
+            row("") {
+                comment("Fuzzing is disabled in ${UtSettings.getPath()}, see the key 'useFuzzing'" )
+            }
+        }
     }
 
     fun isModified(): Boolean {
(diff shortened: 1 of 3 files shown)


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/settings/SettingsWindow.kt
```diff
@@ -209,6 +209,11 @@ class SettingsWindow(val project: Project) {
                 addToRight(symLabel)
             }().constraints(CCFlags.growX)
         }
+        if (!UtSettings.useFuzzing) {
+            row("") {
+                comment("Fuzzing is disabled in ${UtSettings.getPath()}, see the key 'useFuzzing'" )
```
Markoutte, line 214: I think when the path is too long it is not very good. Also, we can do a link instead of path, which opens folder and selects the target file? wdyt?

Markoutte (approved):

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/settings/SettingsWindow.kt
Vassiliy-Kudryashov: Please check new chenges I've made

Markoutte (approved):

PR #1452, opened 2022-12-02
Added support for summaries on the nested classes

### LINKED ISSUE #710: Summaries are not generated for methods from nested classes of nested classes
**Description**

If we have a chain of at least three classes nested in each other (like `class A { class B { class C { int f(); } } }` ), summaries are not generated for the functions of the innermost of them (`A.B.C.f()` in our example).

**To Reproduce**

**UPD.** Nested classes are now supported in plugin, so the behavior can be reproduced just by launching plugin on the below class

As nested classes are currently supported only in contest estimator, the easiest way to reproduce the bug is through tests:

1. Add file `DeepNested.java` into `utbot-sample/src/main/java/org/utbot/examples/inner/` with the following content:

```Java
package org.utbot.examples.inner;

public class DeepNested {
    public class Nested1 {
        public class Nested2 {
            public int f() {
                return 0;
            }
        }
    }
}
```

2. Add file `DeepNestedTest.kt` into `utbot-summary-tests/src/test/kotlin/examples/inner/` with the following content:

```Kotlin
package examples.inner

import examples.SummaryTestCaseGeneratorTest
import org.junit.jupiter.api.Test
import org.utbot.examples.DoNotCalculate
import org.utbot.examples.inner.DeepNested
import org.utbot.framework.plugin.api.MockStrategyApi

class DeepNestedTest : SummaryTestCaseGeneratorTest(
    DeepNested::class,
) {
    @Test
    fun testF() {
        val summary1 = "Test returns from: return 0;"

        val method = DeepNested.Nested1.Nested2::f
        val mockStrategy = MockStrategyApi.NO_MOCKS
        val coverage = DoNotCalculate

        val summaryKeys = listOf(
            summary1
        )

        summaryCheck(method, mockStrategy, coverage, summaryKeys)
    }
}
```

**Expected behavior**

Test passes, as corresponding summaries are generated for tests for `A.B.C.f()`

**Actual behavior**

Test fails because no summaries are generated.

Test fails with assertion message

```
The following comments were produced by the UTBot, but were not found in the list of comments passed in the check() method:

-------------------------------------------------------------

null-------------------------------------------------------------


The following comments were passed in the check() method, but were not found in the list of comments produced by the UTBot:

-------------------------------------------------------------

Test returns from: return 0;
-------------------------------------------------------------
```

**Environment**

Nothing specific.

**Additional context**

For me it seems that problem is in `org.utbot.summary.ast.SourceCodeParser#traverseInnerClassDeclarations` and the fact that it doesn't make any recursive calls.

amandelpie:
# Description

This PR contains the implementation of the enabling summaries on the nested classes.
The existing solution has no ability to dig into the hierarchy of the nested classes (from parent to child)

The proposed solution tries to search in the space of children of each node with a special type related to the class

Fixes #710 

## Type of Change

## Automated Testing

The unit—test for 3-level hierarchy of nested classes is added


amandelpie:
@Markoutte could you please have a look, the same case like earlier but for the Symbolic Execution (unit test could be added)

Markoutte (approved):

PR #1472, opened 2022-12-07
some options moved from gradle.properties to GRADLE_OPTS #1470

### LINKED ISSUE #1470: Move options from gradle.properties to GRADLE_OPTS env
**Description**

Move the following options from gradle.properties to GRADLE_OPTS env:

```
org.gradle.daemon=false
org.gradle.parallel=false
kotlin.compiler.execution.strategy=in-process
org.gradle.caching=false
```

**Expected behavior**

Users take default from gradle.properties but it is optimized for CI usage

**Environment**

Not applicable

**Potential alternatives**

Not users should remove properties manually

**Context**

Not applicable

bissquit:
# Description

`org.gradle.daemon=false` - moved to `GRADLE_OPTS` but this option is not needed because we explicitly run gradle with `--no-daemon` option. Should I remove  `--no-daemon` or leave it as is?
`org.gradle.parallel=false` - moved to `GRADLE_OPTS`
`kotlin.compiler.execution.strategy=in-process` - moved to `GRADLE_OPTS`
`org.gradle.caching=false` - removed completely because we don't have caches on runners by default. To obtain cache we should use `actions/cache@v3` and in those cases we explicitly run gradle with `--build-cache` option
 
Fixes #1470

## Type of Change

# How Has This Been Tested?

No tests needed

## Automated Testing

Run workflow from a branch

## Manual Scenario 

Not applicable

- [ ] check automatic tests result
- [ ] run additional tests if needed


gradle.properties
```diff
@@ -75,11 +75,7 @@ openblasVersion=0.3.10-1.5.4
 arpackNgVersion=3.7.0-1.5.4
 
 # configuration for build server
```
denis-fokin, line 77: Could you add a commentary here, that some options in server environment are passed as as Gradle options and list the options. Configuration for local compilation - much faster

gradle.properties
```diff
@@ -75,11 +75,7 @@ openblasVersion=0.3.10-1.5.4
...
-org.gradle.daemon=false
-org.gradle.parallel=false
 org.gradle.jvmargs="-XX:MaxHeapSize=6144m"
-kotlin.compiler.execution.strategy=in-process
-org.gradle.caching=false
 
 # configuration for local compilation - much faster
 #org.gradle.daemon=true
```
denis-fokin, line 81: Can we enable the properties and overload them as Gradle environment options?

bissquit: `GRADLE_OPTS` environment variable has lower priority than gradle.properties file

denis-fokin: Maybe we should use GRADLE_OPTS here if they present?

gradle.properties
```diff
@@ -75,16 +75,24 @@ openblasVersion=0.3.10-1.5.4
...
 
 # configuration for local compilation - much faster
-#org.gradle.daemon=true
-#kotlin.daemon.jvm.options=-Xmx4g
-#org.gradle.parallel=true
-#org.gradle.caching=true
-#org.gradle.workers.max=8
-#org.gradle.jvmargs="-XX:MaxHeapSize=6144m"
\ No newline at end of file
+# overrided by some parameters in CI, read below about each option
```
Vassiliy-Kudryashov, line 87: "Overriden", not "overrided"

denis-fokin (approved):

victoriafomina (approved):
LGTM

Issue #1504, opened 2022-12-09
Improve executions priorities in minimization


EgorkaKulikov:
**Description**

As is known, it is better to avoid Fuzzer executions when we generate parametrized tests. 
Sometimes they are really required to increase coverage - the we put a separate region with them under a parameterized test
Sometimes we have a pair of absolutely similar executions, but one of them is symbolic and the second is fuzzer one. Thus we should better prefer symbolic.

**Expected behavior**

After the discussion with Sergey Pospelov and Dmitry Ivanov we decided to do the following:
- first of all we minimize symbolic executions
- if some instructions are not still covered, we use fuzzer executions

**Potential alternatives**

The alternative is to miss all fuzzer executions in parameterized tests. It seems to be worth.

PR #1514, opened 2022-12-14
Make symbolic execs preferable in minimization process

sofurihafe:
# Description

After some discussions it was decided to prioritize symbolic executions over fuzzed ones. Now, priority is based on source priority (symbolic is preferable), number of useful lines (more is preferable), and exec id.
For instance, if there are two identical executions differing only by their source, the one from symbolic engine will be preferable in minimization.

Fixes # ([1504](https://github.com/UnitTestBot/UTBotJava/issues/1504))

## Type of Change

# How Has This Been Tested?

Added a new test *testPriority1* in *MinimizationGreedyEssentialTest*.

## Automated Testing

utbot-samples, plus *testPriorityTest1*.

## Manual Scenario 

Generate a test for *max* method in *IntExamples* sample. Verify that there are two tests sourcing from symbolic engine in the generated test.


utbot-framework-test/src/test/kotlin/org/utbot/framework/minimization/MinimizationGreedyEssentialTest.kt
```diff
@@ -68,4 +68,18 @@ class MinimizationGreedyEssentialTest {
         val minimizedExecutions = GreedyEssential.minimize(executions)
         assertEquals((1..size).toList(), minimizedExecutions.sorted())
     }
+
+    @Test
+    fun testPriority1() {
```
EgorkaKulikov, line 73: Please rename it to smth like `testExecutionPriority`

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
```diff
@@ -264,4 +270,9 @@ private fun addExtraIfLastInstructionIsException(
...
  */
 private fun Throwable.exceptionToInfo(): String =
-    this::class.java.name + (this.stackTrace.firstOrNull()?.run { className + methodName + lineNumber } ?: "null")
\ No newline at end of file
+    this::class.java.name + (this.stackTrace.firstOrNull()?.run { className + methodName + lineNumber } ?: "null")
+
+/**
+ * Returns an execution priority. [UtSymbolicExecution] has the highest priority.
+ */
+private fun UtExecution.getExecutionPriority(): Int = if (this is UtSymbolicExecution) 0 else 1
```
EgorkaKulikov, line 278: It is better to write smth like
```
when (this) {
is Symbolic -> 0
isFuzzer -> 1
else -> error("...")
}

Add a comment, why symblic executions have higher priority, please.

EgorkaKulikov (approved):

CaelmBleidd:
>After some discussions it was decided to prioritize symbolic executions over fuzzed ones.

Why? I remember the last discussion of it with Dmitry Timofeev and @sergeypospelov, and they were against this solution because of many reasons. As for me, the most important ones were described by Dmitry in [his first attempt](https://github.com/UnitTestBot/UTBotJava/pull/854). I don't see any of the persons responsible for the minimization part or people who discussed it back then here in reviewers or in the issue discussion, so, the obvious question -- was has changed since that that moment, and what are the answers on Dmitry's (and mine, since we discussed it with him as well) concerns?

PR #1516, opened 2022-12-14
Add tests for fuzzing platform

Markoutte:
# Description

Fuzzing platform should generate reproducible tests on each run with new seeded random object and be cancelable by coroutine context. This PR fixes these problems and adds tests for core functions of fuzzing platform.

## Type of Change

# How Has This Been Tested?

New tests have been added:

* org.utbot.fuzzing.JavaFuzzingTest
* org.utbot.fuzzing.FuzzerSmokeTest
* org.utbot.fuzzing.ProvidersTest

## Manual Scenario 

Tests now have to be reproducible. It means, that fuzzing with more time to run has all those tests that fuzzing with less time to run has. Try to run fuzzing on `org.utbot.fuzzing.samples.Strings#test(String s)`  with different timeouts.

# Checklist:

- [x] No new warnings


Vassiliy-Kudryashov (approved):

utbot-fuzzers/src/test/kotlin/org/utbot/fuzzing/JavaFuzzingTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.utbot.fuzzer.FuzzedConcreteValue
+import org.utbot.fuzzing.samples.Stubs
+import org.utbot.fuzzing.utils.Trie
+
+class JavaFuzzingTest {
+
+    @Test
+    fun `string generates same values`() {
+        fun collect(): List<String> {
+            return runBlockingWithContext {
```
volivan239, line 19: Just thoughts: if we want reproducibility in many/most/all cases, maybe we should find a more general way to check it? E.g., introduce something like

```Kotlin
fun <T> getReproducibleResults(generator: () -> T): T {
    val probe1 = generator()
    val probe2 = generator()
    assertEquals(probe1, probe2)
    return probe1
}
```

Then we can wrap calls to fuzzer in tests in such function and process results as usual. (Just an example, there are lots of other ways to check reproducibility more general).

volivan239 (approved):

PR #1521, opened 2022-12-15
Add disjunction with null condition to isMock constraint

### LINKED ISSUE #1520: Mock label is assigned to null values
**Description**

We have an array that contains information about whether some object is mocked or not with symbolic addresses as its indices. When we create an object, we assign a mark corresponding provided mock information. Because of an error, we do it without the condition that they should not be taken into account if the object we create is null, and it cases UNSAT verdict.

**To Reproduce**

1. Open a test project with utbot-api dependency
2. Copy there a code from `visual proof` section below
3. Run generation with `Other packages` mock strategy

**Expected behavior**

One generated test.

**Actual behavior**

No generated tests.

```java
public int mocksForNullOfDifferentTypes(Integer intValue, Random random) {
    UtMock.assume(intValue == null);
    UtMock.assume(random == null);

    return 0;
}
```

**Environment**
Mocks are enabled, any Junit is ok, mock strategy `all other classes` of `from other packages`

**Additional context**

There is no additional context

CaelmBleidd:
# Description

Added a disjunction with nullability constraint to avoid an assignment to a null object contradicting conditions about whether it is mock or not.

Fixes #1520

## Type of Change

# How Has This Been Tested?

There is no difference between automatic tests and international tests for the engine.

## Automated Testing

org.utbot.examples.mock.CommonMocksExample#mocksForNullOfDifferentTypes

## Manual Scenario 

Did not check it manually.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
@@ -56,4 +56,14 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             coverage = DoNotCalculate
         )
     }
+
+    @Test
+    fun testMocksForNullOfDifferentTypes() {
+        check(
+            CommonMocksExample::mocksForNullOfDifferentTypes,
+            eq(1),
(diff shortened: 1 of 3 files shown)


sergeypospelov (approved):
LGTM

PR #1525, opened 2022-12-15
Remove access for jClass in UtMockFieldInfo

CaelmBleidd:
# Description

We use information from `jClass` when Mocker decides whether it should mock some object or not. Because of many reasons, these classes might be unavailable in Runtime and we have to work with it using `SootClass` instances. 

This problem was found during work on closed project and I don't have a way to reproduce it, but it is a problem that happen many times in the past and we should avoid accessing `jClass` whenever possible.

## Type of Change

# How Has This Been Tested?

There are no additional tests since they highly depend on classpath, project configuration, soot transformation and other options. Unfortunately, the exact combination of these options is lost and the problem occurred on a project with closed code.

## Automated Testing

There are no additional automatic tests.

## Manual Scenario 

There are no manual scenarios. 

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
@@ -188,12 +188,28 @@ class Mocker(
         if (isOverriddenClass(type)) return false  // never mock overriden classes
         if (type.isInaccessibleViaReflection) return false // never mock classes that we can't process with reflection
         if (isMakeSymbolic(mockInfo)) return true // support for makeSymbolic
-        if (type.sootClass.isArtificialEntity) return false // never mock artificial types, i.e. Maps$lambda_computeValue_1__7
-        if (!isEngineClass(type) && (type.sootClass.isInnerClass || type.sootClass.isLocal || type.sootClass.isAnonymous)) return false // there is no reason (and maybe no possibility) to mock such classes
-        if (!isEngineClass(type) && type.sootClass.isPrivate) return false // could not mock private classes (even if it is in mock always list)
+
+        val sootClass = type.sootClass
+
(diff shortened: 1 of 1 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -203,12 +219,22 @@ class Mocker(
...
+                .firstOrNull { it.name == fieldId.name && it.declaringClass.name == fieldId.declaringClass.name }
+                ?: error("Unexpected $fieldId is provided into shouldMock function")
+
+            val sootFieldType = sootField.type
+
             return when {
-                declaringClass.packageName.startsWith("java.lang") -> false
-                !mockInfo.fieldId.type.isRefType -> false  // mocks are allowed for ref fields only
-                else -> return strategy.eligibleToMock(mockInfo.fieldId.type, classUnderTest) // if we have a field with Integer type, we should not mock it
+                sootDeclaringClass.packageName.startsWith("java.lang") -> false // if we have a field with Integer type, we should not mock it
```
Damtev, line 230: Perhaps it would be better to check this condition a bit earlier - since it does not depends on `sootField` variable?

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -203,12 +219,22 @@ class Mocker(
                 return false
             }
 
+            val sootField = sootDeclaringClass
+                .fields
+                .firstOrNull { it.name == fieldId.name && it.declaringClass.name == fieldId.declaringClass.name }
```
Damtev, line 224: There already exists a variable for `fieldId.declaringClass`

Damtev (changes requested):

Damtev (approved):

PR #1526, opened 2022-12-15
Change unsafe access to jClass in updateGenericTypeInfo

CaelmBleidd:
# Description

In `org.utbot.engine.Traverser#updateGenericTypeInfo` we access `executable` for the method under test. It goes into `jClass` and extracts a method instance from it, which may cause an error if we don't have this class in the classpath. Since this function is not mandatory for analysis, we can just not update generic info and continue analysis instead of falling.

## Type of Change

# How Has This Been Tested?

There are no additional tests since they highly depend on classpath, project configuration, soot transformation and other options. Unfortunately, the exact combination of these options is lost and the problem occurred on a project with closed code.

## Automated Testing

There are no additional automatic tests.

## Manual Scenario 

There are no manual scenarios.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
@@ -1035,7 +1035,10 @@ class Traverser(
      * Stores information about the generic types used in the parameters of the method under test.
      */
     private fun updateGenericTypeInfo(identityRef: IdentityRef, value: ReferenceValue) {
-        val callable = methodUnderTest.executable
+        // If we don't have access to methodUnderTest's jClass, the engine should not fail
+        // We just won't update generic information for it
+        val callable = runCatching { methodUnderTest.executable }.getOrNull() ?: return
+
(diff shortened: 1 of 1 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -1035,7 +1035,10 @@ class Traverser(
      * Stores information about the generic types used in the parameters of the method under test.
      */
     private fun updateGenericTypeInfo(identityRef: IdentityRef, value: ReferenceValue) {
-        val callable = methodUnderTest.executable
+        // If we don't have access to methodUnderTest's jClass, the engine should not fail
+        // We just won't update generic information for it
+        val callable = runCatching { methodUnderTest.executable }.getOrNull() ?: return
```
Damtev, line 1040: `getOrElse { return }`?

Damtev (changes requested):

Damtev (approved):

Issue #1272, opened 2022-10-31
Test with comment 'This test fails because ... java.lang.ArithmeticException' does not fail


tyuldashev:
**Description**

UTBot could find situations which leads to arithmetic overflow and generate tests for such cases adding comment that such test would fail, but actually such test pass.

**To Reproduce**

1. Open `utbot`  project in IntelliJ Idea
2. In UTBot settings set `Overflow detection:` to `Treat overflows as errors`
3. Set `Test generated method` to 100 % `Symbolic`
4. Generate tests for `org.utbot.examples.math.OverflowExamples#intOverflow` method
5. Open the generated test

**Expected behavior**

Either UTBot generates test which causes ArithmeticException during execution 
Or it doesn't write that test should fail

**Actual behavior**

Test which are supposed to generate errors are generated, but during execution they pass.

Sample of such test:
```java
    @Test
    @DisplayName("intOverflow: x * x * x > 0 -> ThrowArithmeticException")
    public void testIntOverflow_ThrowArithmeticException() {
        OverflowExamples overflowExamples = new OverflowExamples();

        /* This test fails because method [org.utbot.examples.math.OverflowExamples.intOverflow] produces [java.lang.ArithmeticException: int  *  overflow] */
        overflowExamples.intOverflow(-11086, 0);
    }
```

**Visual proofs (screenshots, logs, images)**

![image](https://user-images.githubusercontent.com/10035839/199029014-c254a4d0-1901-4528-95cb-fa07a6696948.png)

**Additional context**

So this case is more about how to properly present results of UTBot finding. Currently there are at least two problems:
* incorrect comment that test fails
* as test pass it would be difficult to find such test in case of bulk test generation, so overflow problem could be left unnoticed.
Probably we should add `fail()` to such test, so it would actually fail and draw user attention.

denis-fokin:
Not enough resources, move to the next release

PR #1529, opened 2022-12-15
Avoid false positives in tests with overflow

EgorkaKulikov:
# Description

Currently generated unit tests for `OverflowExamples` from utbot-samples are false positives.
The only sign of something wrong is a comment about failing test. That actually passes...

We should add failing assertion and show that actual result is not reliable.

An approach of this fix is to suggest something that is a `SymbolicFailure`, but looks like a `SymbolicSuccess`

Fixes # ([1272](https://github.com/UnitTestBot/UTBotJava/issues/1272))

## Type of Change

# How Has This Been Tested?

Please, provide regression or integration tests for your code changes. If you don't do that, the reviewer can and highly likely **_will reject_** the PR. It is the contributor's responsibility to provide such tests or to reason why they are missing.

## Automated Testing

Specify tests that help to verify the change automatically.  

_Example:_ org.utbot.examples.algorithms.BinarySearchTest

## Manual Scenario 

Please, provide several scenarios that you went through to verify that the change worked as expected.


CaelmBleidd:
> Currently generated unit tests for OverflowExamples from utbot-samples are false positives. The only sign of something wrong is a comment about failing test. That actually passes...

If the problem is that we have tests that should fail, but they pass, it is a false negative. If we find errors in places where there is no overflow or we fail some tests where there is no overflow -- it is a false positive

EgorkaKulikov:
Replaced a week later with a better fix.

PR #1573, opened 2022-12-22
New treating of artificial errors

EgorkaKulikov:
# Description

Currently generated unit tests for OverflowExamples from utbot-samples are false negatives if the mode to treat overflows as errors is selected.

The only sign of something wrong is a comment about failing test. Test passes whenever it's behavior is unpredictable.

We should add failing assertion and show that actual result is not reliable. It requires changes in
- engine (new way to traverse and resolve overflows)
- engine tests (correct OverflowDetectionAndErrors test)
- codegen (new approach to render overflow detections, new assert introduced)
- reporting (support one more behavior type - artificial errors)
- summarization (new display names, test method names, summaries)

Everything except summarization is ready now.

Fixes # ([1272](https://github.com/UnitTestBot/UTBotJava/issues/1272))

## Type of Change

# How Has This Been Tested?

`OverflowAsErrorTest`

as far as no new scenarios were added to the symbolic engine, we just corrected current tests

## Manual Scenario 

Stanfard checks on both codegen languages and used test frameworks that generated code with failure assertion is compilable and test run results correlate with our expectations.

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
@@ -4,6 +4,8 @@ import org.utbot.common.PathUtil
 import org.utbot.common.WorkaroundReason
 import org.utbot.common.isStatic
 import org.utbot.common.workaround
+import org.utbot.engine.ArtificialError
+import org.utbot.engine.OverflowDetectionError
 import org.utbot.framework.assemble.assemble
 import org.utbot.framework.codegen.domain.ForceStaticMocking
 import org.utbot.framework.codegen.domain.ParametrizedTestSource
@@ -141,6 +143,7 @@ import org.utbot.framework.codegen.tree.CgTestClassConstructor.CgComponents.getS
 import org.utbot.framework.codegen.tree.CgTestClassConstructor.CgComponents.getTestFrameworkManagerBy
 import org.utbot.framework.codegen.tree.CgTestClassConstructor.CgComponents.getVariableConstructorBy
 import org.utbot.framework.plugin.api.UtExecutionResult
+import org.utbot.framework.plugin.api.UtOverflowFailure
 import org.utbot.framework.plugin.api.UtStreamConsumingFailure
 import org.utbot.framework.plugin.api.util.allSuperTypes
 import org.utbot.framework.plugin.api.util.baseStreamClassId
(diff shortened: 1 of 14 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
```diff
@@ -358,6 +367,25 @@ open class CgMethodConstructor(val context: CgContext) : CgContextOwner by conte
...
 
+    //TODO: ISSUE-1546 introduce some kind of language-specific string interpolation in Codegen
+    private fun prepareArtificialFailureMessage(executionResult: UtExecutionResult): CgVariable {
+        when (executionResult) {
+            is UtOverflowFailure -> {
+                val preparedMethodArguments = methodArguments.joinToString(separator = ", ") {
+                    when (it) {
+                        is CgVariable -> it.name
+                        is CgLiteral -> it.value.toString()
+                        else -> it.toString()
```
Damtev, line 378: Is it really possible for something else to appear here?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
```diff
@@ -358,6 +367,25 @@ open class CgMethodConstructor(val context: CgContext) : CgContextOwner by conte
         }
     }
 
+    //TODO: ISSUE-1546 introduce some kind of language-specific string interpolation in Codegen
+    private fun prepareArtificialFailureMessage(executionResult: UtExecutionResult): CgVariable {
```
Damtev, line 371: Are you sure in such creating a variable for the failing message? Perhaps it is better to use a constant string in the `fail` invocation?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
```diff
@@ -405,6 +433,7 @@ open class CgMethodConstructor(val context: CgContext) : CgContextOwner by conte
         if (exception is AccessControlException) return false
         // tests with timeout or crash should be processed differently
         if (exception is TimeoutException || exception is ConcreteExecutionFailureException) return false
+        if (exception is OverflowDetectionError) return false
```
Damtev, line 436: We can use `ArtificialError` here, I guess

Damtev (changes requested):
Actually, looks very good. Please, do not forget to fill the description of the PR

Damtev (approved):

PR #1585, opened 2022-12-23
Fixed improper verification of the mocha installation

### LINKED ISSUE #1582: Improper verification of the installation of the mocha test framework
**Description**

Currently, only global installations of npm packages are checked. Due to the pull request #1566, where the installation of mocha was moved to the local one, it is necessary to check whether mocha is installed locally in the repository.

1. Run UTBot plugin with JavaScript 
2. Try to generate unit-tests

**Expected behavior**

Checking the installed mocha package should occur correctly, if there is a local mocha installation, a window with an installation suggestion should not appear.

**Actual behavior**

For each iteration, the mocha installation window appears even if mocha is installed locally manually.

rudolf101:
# Description
Now the installed mocha package is checked correctly, if there is a local mocha installation, the window with the installation suggestion does not appear.
Fixes #1582

## Type of Change

# How Has This Been Tested?
Manually 

## Manual Scenario 
* If mocha is not installed on the system, the user will be prompted to download mocha
* Otherwise, the plugin window will appear without a download suggestion

- [x] No new warnings
- [x] All manual scenarios pass locally with my changes

### DIFF
--- a/utbot-intellij-js/src/main/kotlin/org/utbot/intellij/plugin/language/js/Utils.kt
+++ b/utbot-intellij-js/src/main/kotlin/org/utbot/intellij/plugin/language/js/Utils.kt
@@ -22,16 +22,22 @@ fun getFrameworkLibraryPath(npmPackageName: String, model: JsTestsModel?): Strin
     return input.substringBefore(npmPackageName) + npmPackageName
 }
 
-fun findFrameworkLibrary(npmPackageName: String, model: JsTestsModel): Boolean {
-    val (bufferedReader, _) = JsCmdExec.runCommand(
+private fun npmListByFlag(model: JsTestsModel, flag: String): String {
+    val (bufferReader, _) = JsCmdExec.runCommand(
         dir = model.project.basePath!!,
         shouldWait = true,
         timeout = 10,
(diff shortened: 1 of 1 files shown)


utbot-intellij-js/src/main/kotlin/org/utbot/intellij/plugin/language/js/Utils.kt
```diff
@@ -22,24 +22,30 @@ fun getFrameworkLibraryPath(npmPackageName: String, model: JsTestsModel?): Strin
...
         shouldWait = true,
         timeout = 10,
-        cmd = arrayOf(model.pathToNPM, "list", "-g")
+        cmd = arrayOf(model.pathToNPM, "list", flag)
     )
-    val checkForPackageText = bufferedReader.readText()
-    bufferedReader.close()
-    if (checkForPackageText == "") {
+    val packages = bufferReader.readText()
+    bufferReader.close()
```
denis-fokin, line 33: Is that correct that you close the reader for the input stream and do not close for the error stream (which is created in runCommand)?

denis-fokin (approved):

PR #1594, opened 2022-12-27
Pelevin/improve java fuzzing type system

Markoutte:
# Description

Adds implementation for `Iterator<*>` type and improve type system to support recursive or dependent types like:

```
public static  <T extends Iterator<T>> int size(T some)
```

## Type of Change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

## Automated Testing

`org.utbot.fuzzing.JavaFuzzingTest`

## Manual Scenario 

Try to generate tests with only fuzzing for these functions:

```java
public static  <T extends Iterator<T>> int size(T some) {
        int r = 0;
        while (some.hasNext()) {
            T next = some.next();
            while (next != null && next.hasNext()) {
                T n2 = next.next();
                while (n2 != null && n2.hasNext()) {
                    n2.next();
                    r++;
                }
                r++;
            }
            r++;
        }
        return r;
    }
```

```java
public class ManyRecursiveCalls<T extends Map<String, Number>> {

    private final T[] value;


    public ManyRecursiveCalls(T[] value) {
        this.value = value;
    }

    public double getSum() {
        double sum = 0;
        for (T numbers : value) {
            for (Number number : numbers.values()) {
                if (number.doubleValue() > 0) {
                    sum += number.doubleValue();
                }
            }
        }
        if (sum == 0.0) {
            throw new IllegalStateException();
        }
        return sum;
    }
}
```

- [x] No new warnings

### DIFF
--- a/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzing/JavaLanguage.kt
+++ b/utbot-fuzzers/src/main/kotlin/org/utbot/fuzzing/JavaLanguage.kt
@@ -11,7 +11,7 @@ import org.utbot.fuzzing.utils.Trie
 import java.lang.reflect.*
 import java.util.concurrent.TimeUnit
 import kotlin.system.measureNanoTime
-import java.util.IdentityHashMap
+import kotlin.random.Random
 
 private val logger = KotlinLogging.logger {}
 
@@ -20,7 +20,8 @@ typealias JavaValueProvider = ValueProvider<FuzzedType, FuzzedValue, FuzzedDescr
 class FuzzedDescription(
     val description: FuzzedMethodDescription,
     val tracer: Trie<Instruction, *>,
-    val typeCache: IdentityHashMap<Type, FuzzedType>,
+    val typeCache: MutableMap<Type, FuzzedType>,
+    val random: Random,
 ) : Description<FuzzedType>(
     description.parameters.mapIndexed { index, classId ->
         description.fuzzerType(index) ?: FuzzedType(classId)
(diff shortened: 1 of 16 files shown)


utbot-fuzzers/src/main/kotlin/org/utbot/fuzzing/JavaLanguage.kt
```diff
@@ -118,22 +127,85 @@ suspend fun runJavaFuzzing(
...
+        cache[t] = target
+        g += generics(t).map {
+            toFuzzerType(it, classId, generics, cache)
+        }
+    }
+    return target
+}
+
+private fun Type.findRealTypeUntilTypeVariable() : Type {
+    var type: Type = this
```
volivan239, line 179: The name of the method is misleading, I think it should be something like `Type.replaceWithUpperBoundUntilNotTypeVariable`, also docs would be great here

volivan239 (commented):
For third type parameter of the method below, `toFuzzerType` returns a finite type (see image below). Is this the expected behavior?

```Java
public static <A extends Iterable<B>, B extends List<A>, C extends List<? extends Iterable<A>>> A example(A c1, B c2, C c) {
    return c2.iterator().next();
}
```

image

utbot-fuzzers/src/main/kotlin/org/utbot/fuzzing/JavaLanguage.kt
Markoutte: Thx for the example! Not, this is not expected, fixed

volivan239 (approved):

PR #1609, opened 2022-12-28
Fix broken plugin name UnittestBot -> UnitTestBot.NET

Vassiliy-Kudryashov:
Fix 'until' version 232 -> 223


olganaumenko (approved):
Looks good to me

PR #1619, opened 2023-01-03
Add support for generic information retrieved from signatures of nested methods

### LINKED ISSUE #1613: Use generic information for nested methods parameters
**Description**

For now, UtBot processes generic information only for a method under test. There is an obvious way of improvement: we should save and use generic information for other methods using the same instruments

**Expected behavior**

If we have some method under test `foo(List<Integer>)` and another method `bar(List<Integer>)` that is called inside of the `foo`, we should use information that the first parameter passed into `bar` have an `Integer` type parameter

**Environment**

There is no specific environment.

**Potential alternatives**

Not suitable

**Context**

This issue might be considered a small part of #924, but none of the suggestions from there is implemented here.

CaelmBleidd:
# Description

This request adds support for generic information when we process nested methods. If we pass an object as a parameter and know type information from signatures about this parameter, we will add it to avoid path explosion because of the types.

Fixes #1613 

## Type of Change

# How Has This Been Tested?

Added a new package called `generics`

## Automated Testing

org.utbot.examples.types.Generics

## Manual Scenario 

You can run generation for the methods I mentioned with enabled visualization, and see, that there is no additional virtual invokes because of absent type information.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework-test/src/test/kotlin/org/utbot/examples/types/GenericsTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/examples/types/GenericsTest.kt
@@ -1,0 +1,63 @@
+package org.utbot.examples.types
+
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.Test
+import org.utbot.testcheckers.eq
+import org.utbot.testing.UtValueTestCaseChecker
+import org.utbot.testing.ignoreExecutionsNumber
+
+internal class GenericsTest : UtValueTestCaseChecker(
+    testClass = GenericsTest::class,
+    testCodeGeneration = false // TODO empty files are generated https://github.com/UnitTestBot/UTBotJava/issues/1616
+) {
+    @Test
+    fun mapAsParameterTest() {
+        check(
+            Generics::mapAsParameter,
+            eq(2),
+            { map, _ -> map == null },
+            { map, r -> map != null && r == "value" },
+        )
+    }
+
+    @Test
+    @Disabled("https://github.com/UnitTestBot/UTBotJava/issues/1620 wrong equals")
+    fun genericAsFieldTest() {
+        check(
+            Generics::genericAsField,
+            ignoreExecutionsNumber,
+            { obj, r -> obj?.field == null && r == false },
+            // we can cover this line with any of these two conditions
+            { obj, r -> (obj.field != null && obj.field != "abc" && r == false) || (obj.field == "abc" && r == true) },
+        )
+    }
+
+    @Test
+    fun mapAsStaticFieldTest() {
+        check(
+            Generics::mapAsStaticField,
+            ignoreExecutionsNumber,
+            { r -> r == "value" },
+        )
+    }
+
+    @Test
+    fun mapAsNonStaticFieldTest() {
+        check(
+            Generics::mapAsNonStaticField,
+            ignoreExecutionsNumber,
+            { map, _ -> map == null },
+            { map, r -> map != null && r == "value" },
+        )
+    }
+
+    @Test
+    fun methodWithRawTypeTest() {
+        check(
+            Generics::methodWithRawType,
+            eq(2),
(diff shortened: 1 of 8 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2084,6 +2095,19 @@ class Traverser(
             checkAndMarkLibraryFieldSpeculativelyNotNull(field, createdField)
         }
 
+        runCatching {
```
Damtev, line 2098: Do we need `runCatching` here because of `jClass` usage? Or by any other reason? Add some comments, please 

Also, it is obvious the purpose of this block of code just now, but it may be not easy to understand in the future. Could you comment on why we need this code (or, even better, introduce a function), please?

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2084,6 +2095,19 @@ class Traverser(
...
+            if (createdField !is ReferenceValue) return@runCatching
+
+            val jClass = field.declaringClass.id.jClass
+            val fields = (jClass.fields + jClass.declaredFields).toSet()
```
Damtev, line 2102: It looks like we skip here non-public inherited fields. Is it expected behavior?

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2650,6 +2674,22 @@ class Traverser(
...
+            }
+        }
+
+        if (invocation.instance != null) {
+            runCatching {
+                updateGenericTypeInfoFromMethod(method, invocation.instance, parameterIndex = 0)
+            }
+        }
+
         /**
```
Damtev, line 2693: Add some comments about a purpose of this code please, since it will not be obvious in the future

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2650,6 +2674,22 @@ class Traverser(
      * Returns results of native calls cause other calls push changes directly to path selector.
      */
     private fun TraversalContext.commonInvokePart(invocation: Invocation): List<MethodResult> {
+        val method = invocation.method.executableId
+
+        invocation.parameters.mapIndexed { index, param ->
+            if (param !is ReferenceValue) return@mapIndexed
+
+            runCatching {
```
Damtev, line 2682: Why do we need `runCatching` here?

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2650,6 +2674,22 @@ class Traverser(
...
+                updateGenericTypeInfoFromMethod(method, param, parameterIndex = index + 1)
+            }
+        }
+
+        if (invocation.instance != null) {
+            runCatching {
```
Damtev, line 2688: Same about `runCatching`

utbot-framework-test/src/test/kotlin/org/utbot/examples/types/GenericsTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    fun genericAsFieldTest() {
+        check(
+            Generics::genericAsField,
+            ignoreExecutionsNumber,
+            { _, r -> r == true },
+            { _, r -> r == false },
+        )
+    }
```
Damtev, line 32: It looks like this test does not distinguish situations where the field is not null and does not equal "abc" and a situation where an object is null or the field is null.

Damtev (changes requested):

Damtev (changes requested):

Damtev (approved):

Issue #88, opened 2022-05-31
Generics in Kotlin codegen


nikitavlaev:
**Description**

Extract type variables from the available structures inside the engine and put them into the result model.

**Expected behavior**

Kotlin code should be generated with generics where it is required.
Framework tests should pass with compilation for Kotlin enabled.
Additional testing on contest estimator is to be done.

**Environment**

Not applicable

**Potential alternatives**

Not applicable

**Context**

Generic types are much more often required in Kotlin than in Java(e.g. collections), so additional support for them is to be made.
Currently, generated Kotlin code does not use generics, and thus a lot of code does not compile.

PR #1621, opened 2023-01-04
Add missed mock generator for static fields

CaelmBleidd:
# Description

This request adds two missed generators for mocks that led to unexpected unsat results. 
It happened because when a static variable was on a right side of an assignment, it was created with the corresponding mock generator. Then, if some other variable related to the same class would be encountered on the left side of the assignment, it was created without such a generator, which led to a contradiction between `isMock` values in these two scenarios. 

There was one more missed generated related to final statics: when we reset statics, we should provide a corresponding mock generator.

## Type of Change

# How Has This Been Tested?

The same as automatic tests.

## Automated Testing

`org.utbot.examples.mock.MockStaticFieldExampleTest#testCheckMocksInLeftAndRightAssignPartFinalField`
`org.utbot.examples.mock.MockStaticFieldExampleTest#testCheckMocksInLeftAndRightAssignPart`

## Manual Scenario 

There is no specific manual scenario.

# Checklist:

- [x] No new warnings

### DIFF
--- a/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/MockStaticFieldExampleTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/MockStaticFieldExampleTest.kt
@@ -6,15 +6,29 @@ import org.utbot.framework.plugin.api.MockInfo
 import org.utbot.framework.plugin.api.MockStrategyApi.OTHER_PACKAGES
 import kotlin.reflect.KClass
 import org.junit.jupiter.api.Test
+import org.utbot.examples.mock.others.ClassWithStaticField
+import org.utbot.framework.plugin.api.CodegenLanguage
+import org.utbot.framework.plugin.api.UtCompositeModel
+import org.utbot.framework.plugin.api.UtNewInstanceInstrumentation
 import org.utbot.testcheckers.eq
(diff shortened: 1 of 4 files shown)


utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/MockStaticFieldExampleTest.kt
```diff
@@ -6,15 +6,28 @@ import org.utbot.framework.plugin.api.MockInfo
...
 import org.utbot.testing.singleMockOrNull
 import org.utbot.testing.value
 
-internal class MockStaticFieldExampleTest : UtValueTestCaseChecker(testClass = MockStaticFieldExample::class) {
+internal class MockStaticFieldExampleTest : UtValueTestCaseChecker(
+    testClass = MockStaticFieldExample::class,
+    testCodeGeneration = true,
+    pipelines = listOf(
+        TestLastStage(CodegenLanguage.JAVA, lastStage = TestExecution),
+        TestLastStage(CodegenLanguage.KOTLIN, lastStage = CodeGeneration)
```
Damtev, line 28: Please, mention the issue #88

Damtev (changes requested):

Damtev (approved):

PR #1658, opened 2023-01-11
Added constructing `this` instance for captured values of lambda in case of early return

Damtev:
# Description

If non-static lambda is a parameter in an execution but its init section was not analyzed (e.g., an exception was thrown earlier), `this` captured instance could be constructed incorrectly (with `null` value) or not be constructed at all. We need to prevent such cases and construct at least an empty object for `this` instance.

## Type of Change

# How Has This Been Tested?

## Automated Testing

Specify tests that help to verify the change automatically.  

`org.utbot.examples.lambda.ThrowingWithLambdaExampleTest#testAnyExample`.

## Manual Scenario 

`com.google.common.collect.TreeMultiset.forEachEntry` in Contest Estimator.

- [x] No new warnings

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/Resolver.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/Resolver.kt
@@ -131,7 +131,8 @@ class Resolver(
     private val typeResolver: TypeResolver,
     val holder: UtSolverStatusSAT,
     methodUnderTest: ExecutableId,
-    private val softMaxArraySize: Int
+    private val softMaxArraySize: Int,
+    private val objectCounter: ObjectCounter
 ) {
 
     private val classLoader: ClassLoader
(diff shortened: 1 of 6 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/Resolver.kt
```diff
@@ -529,7 +530,27 @@ class Resolver(
...
+                    if (thisInstanceField !in collectedFieldModels || collectedFieldModels[thisInstanceField] is UtNullModel) {
+                        // Non-static lambda has to have `this` instance captured as `cap0` field that cannot be null,
+                        // so if we do not have it as field or it is null (for example, an exception was thrown before initializing lambda),
+                        // we need to construct `this` instance by ourselves.
+                        // Since we do not know its fields, we create an empty object of the corresponding type that will be
+                        // constructed in codegen using reflection.
+                        val thisInstanceClassId = sootClass.name.substringBeforeLast("\$lambda").let {
+                            Scene.v().getSootClass(it)
+                        }.id
+                        val thisInstanceModel = UtCompositeModel(id = null, thisInstanceClassId, isMock = false)
```
CaelmBleidd, line 547: What do you mean `null` as its address? It is nullable due to very specific cases with mocks where we have to have a container for actions for an `object` that cannot be constructed, otherwise is it suppose to be an actual address in memory and some parts (for example, state comparison) are depends on it

CaelmBleidd (changes requested):

CaelmBleidd (approved):

PR #1724, opened 2023-01-20
Reduce the process of array construction to 200 stores

EgorkaKulikov:
## Title

Reduce the process of array construction to 200 stores

### Manual tests

Run ContestEstimator with the following settings:

```
projectFilter = listOf("guava-26.0")
Classes:
com.google.common.primitives.Booleans
com.google.common.primitives.Shorts
```

Verify that there are no significant regression in broken tests count after reducing array construction.


EgorkaKulikov:
Fortunately, a better fix was applied.

PR #1761, opened 2023-01-31
Taint analysis configuration format

### LINKED ISSUE #1760: Taint analysis configuration format
**Description**

Develop a configuration format for the future UTBot taint analysis.

IdeaSeeker:
## Description

Add the taint analysis configuration format. More detailed information in the TaintAnalysis.md

Fixes #1760

## Self-check list

### DIFF
--- a/docs/TaintAnalysis.md
+++ b/docs/TaintAnalysis.md
@@ -1,12 +1,14 @@
 # Taint analysis
 
+[//]: # (TODO)
+
 ## Configuration
 
 There are four _rule types_: sources, passes, cleaners and sinks.
 
 The rule of each type contains
 * a method name,
-* a method description.
+* a method description
 
 ### Method name
 
@@ -112,10 +114,13 @@ sinks:
           check: [ arg1, arg3 ]
           marks: [ sql-injection, xss ]
 ```
+***
 
 For all the rule types, method descriptions can optionally contain
 * a method signature,
 * runtime conditions.
+
+**Signature**
 
 Method `signature` (optional) is a list of _argument types_ (at compile time):
 
@@ -124,6 +129,8 @@ Method `signature` (optional) is a list of _argument types_ (at compile time):
 Please note that
 - the type name is written in `<>`,
 - `_` means any type.
+
+**Conditions**
 
 Runtime `conditions` (optional) are conditions that must be met to trigger this rule.
 
--- a/docs/TaintAnalysis.md
+++ b/docs/TaintAnalysis.md
@@ -5,7 +5,7 @@
 ### Basic
 
 - There are 4 sections: sources, passes, cleaners and sinks. Each of them contains a list of methods descriptions (rules). Fully qualified method names can be written in one line, or it can be defined by nested parts — first the package names, then the class name, and finally the method name. Note that regular expressions in names are not supported.
-- Each rule can contain a method signature — a list of argument types (in compile-time): `signature: [  <int>, _, <java.lang.String> ]`
+- Each rule can contain a method signature — a list of argument types (at compile-time): `signature: [  <int>, _, <java.lang.String> ]`
   - name of the type is written in `<>`.
   - `_` means any type.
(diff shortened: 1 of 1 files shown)


docs/TaintAnalysis.md
```diff
@@ -0,0 +1,64 @@
...
+    remove-from: this
+    marks: [sensitive-data, sql-injection]
+  com.company.method8:
+    remove-from: [arg1, return]
+    marks: nullable
+
+sinks:
+  org.example:
+    log:
+      check-in: arg1
```
CaelmBleidd, line 59: I guess it'd be called just `check` since there are no checks for out-arguments

docs/TaintAnalysis.md
```diff
@@ -0,0 +1,64 @@
...
+- Each method description can contain the key `conditions` (optionally).
+- Keys like `add-to` or `remove-from` can contain a list of values or only one value. 
+  Possible values:
+  - this
+  - arg1
+  - arg2
+  - ...
+  - return
+- The `marks` key can also contain a list of marks or only one mark.
+- Fully qualified name of the method may be determined in parts.
```
CaelmBleidd, line 14: Please, specify a supported scope: 
1. Do we support regexps anywhere?
2. How `condition` should look like? Which constructions should be supported?
3. If conditions are specified for the corresponding argument, we cannot add a condition with two involved of them, can we? For example, it is a taint then and only then when either the first argument is a string and the second one is a number or vice versa
4. What happens when we have several rules for the same method? What if they have a contradiction?

docs/TaintAnalysis.md
```diff
@@ -0,0 +1,64 @@
...
+
+**Example:**
+
+```YAML
+sources:
+  com:
+    abc.method1:
+      add-to: [this, return] 
+      marks: nullable
```
CaelmBleidd, line 23: It is better to change the example since `nullable` for `this` looks odd

docs/TaintAnalysis.md
```diff
@@ -0,0 +1,64 @@
...
+    bca.method2:
+      add-to: return
+      marks: [sensitive-data, sql-injection]
+
+passes:
+  com.abc.method2:
```
CaelmBleidd, line 29: How to choose between overloads?

docs/TaintAnalysis.md
```diff
@@ -0,0 +1,64 @@
...
+      arg0:
+        not: ""
+      arg1: <int> # type
+      arg2: [1, 2, 3] # arg2 should be equal to one of: 1, 2 or 3 
+      arg3:
+        not: [4, 5, 6]
+      arg4: [<float>, <java.lang.String>]
+      arg5:
+        not: [<int>, <boolean>]
+      return: false
```
CaelmBleidd, line 41: What does this condition mean? For me, it looks like it makes the whole formula unsatisfiable because I think about it not as about equality, but as about a conjunct value itself

CaelmBleidd (changes requested):

docs/TaintAnalysis.md
```diff
@@ -0,0 +1,178 @@
...
+
+- Values and types can be negated using the `not` key, as well as combined using lists (`or` semantics).
+- Nesting is allowed.
+
+**Example:**
+
+```yaml
+conditions:
+  this: [ "in", "out" ] # this should be equal to one of: "in" or "out"
+  arg1: [ <int>, <float> ] # arg1 should be int or float
```
CaelmBleidd, line 123: How to combine conditions?
For example, division by zero could be raised only if `arg0 is Integer and arg0 == 0`

docs/TaintAnalysis.md
```diff
@@ -0,0 +1,178 @@
...
+  arg2: { not: 0 } # arg2 should not be equal to 0
+  arg3:
+    not: [ 1, 2, 3, 5, 8 ] # should not be equal to any of the listed numbers
+  arg4: [ "", { not: <java.lang.String> } ] # should be an empty string or not a string at all
```
CaelmBleidd, line 127: Same here, the condition might be `If it is a String, it must be an empty one"

CaelmBleidd (changes requested):

CaelmBleidd (approved):

PR #1788, opened 2023-02-13
Initilial implementation of Spring test classes code generation

sofurihafe:
## Description

This PR adds support for necessary test class fields rendering (see *InjectMocks*, *Mock* below).
Also, in case of mocks it constructs them in test method (see *(when(...))*).

```java
@Service
public class OrderService {
    @Autowired
    private OrderRepository orderRepository;

    public List<Order> getOrders() {
        return orderRepository.findAll();
    }
}
```

```java
public final class OrderServiceTest {
    @InjectMocks
    public OrderService orderService;

    @Mock
    public OrderRepository orderRepositoryMock;

    public AutoCloseable mockitoCloseable;

    @BeforeEach
    public void setUp() {
        mockitoCloseable = openMocks(this);
    }

    @AfterEach
    public void tearDown() throws Exception {
        mockitoCloseable.close();
    }

    @Test
    public void testGetOrders1() {
        (when(orderRepositoryMock.findAll())).thenReturn(((List) null));

        List actual = orderService.getOrders();

        assertNull(actual);
    }
}
```

Fixes nothing, adds new feature.

### Automated tests

No automated tests are present at this moment.

### Manual tests

Tested manually on a third-party Spring project.

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/Domain.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/Domain.kt
@@ -3,6 +3,7 @@ package org.utbot.framework.codegen.domain
 import org.utbot.framework.DEFAULT_EXECUTION_TIMEOUT_IN_INSTRUMENTED_PROCESS_MS
 import org.utbot.framework.codegen.domain.builtin.mockitoClassId
 import org.utbot.framework.codegen.domain.builtin.ongoingStubbingClassId
+import org.utbot.framework.codegen.domain.context.CgContext
 import org.utbot.framework.codegen.domain.models.CgClassId
 import org.utbot.framework.codegen.tree.argumentsClassId
 import org.utbot.framework.plugin.api.BuiltinClassId
@@ -11,6 +12,8 @@ import org.utbot.framework.plugin.api.CodeGenerationSettingBox
 import org.utbot.framework.plugin.api.CodeGenerationSettingItem
 import org.utbot.framework.plugin.api.MethodId
 import org.utbot.framework.plugin.api.TypeParameters
+import org.utbot.framework.plugin.api.UtModel
+import org.utbot.framework.plugin.api.idOrNull
 import org.utbot.framework.plugin.api.isolateCommandLineArgumentsToArgumentFile
 import org.utbot.framework.plugin.api.util.booleanArrayClassId
 import org.utbot.framework.plugin.api.util.booleanClassId
@@ -773,3 +776,17 @@ object SpringBoot : DependencyInjectionFramework(
     id = "spring-boot",
     displayName = "Spring Boot"
 )
+
+/**
+ * Extended id of [UtModel], unique for whole test set.
+ *
+ * Allows to distinguish models from different executions,
+ * even if they have the same value of `UtModel.id`.
+ */
+data class ModelId(
+    private val id: Int?,
+    private val executionId: Int,
+)
+
+fun UtModel.withId(executionId: Int = -1) = this to ModelId(this.idOrNull(), executionId)
+
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/Domain.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/Domain.kt
@@ -788,5 +788,5 @@ data class ModelId(
     private val executionId: Int,
 )
 
-fun UtModel.withId(executionId: Int = -1) = this to ModelId(this.idOrNull(), executionId)
+fun UtModel.withExecutionId(executionId: Int = -1) = ModelId(this.idOrNull(), executionId)
 
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
@@ -571,15 +571,7 @@ data class CgContext(
         }
     }
 
-    override fun getIdByModel(model: UtModel): ModelId {
-        if (model !in modelIds) {
-            modelIds[model] = model.withExecutionId()
-        }
-
-        return modelIds.getOrElse(model) {
-            error("ModelId for $model should have also been created")
-        }
-    }
+    override fun getIdByModel(model: UtModel): ModelId = modelIds.getOrPut(model) { model.withExecutionId() }
 
     private fun createClassIdForNestedClass(testClassModel: SimpleTestClassModel): ClassId {
         val simpleName = "${testClassModel.classUnderTest.simpleName}Test"
@@ -599,6 +591,7 @@ data class CgContext(
         requiredUtilMethods.clear()
         valueByModel.clear()
         valueByModelId.clear()
+        modelIds.clear()
         mockFrameworkUsed = false
     }
 
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
@@ -22,6 +22,7 @@ import kotlinx.collections.immutable.PersistentSet
 import kotlinx.collections.immutable.persistentListOf
 import kotlinx.collections.immutable.persistentMapOf
 import kotlinx.collections.immutable.persistentSetOf
+import org.utbot.framework.codegen.domain.ModelId
 import org.utbot.framework.codegen.domain.ProjectType
 import org.utbot.framework.codegen.domain.models.CgMethodTestSet
 import org.utbot.framework.codegen.domain.builtin.TestClassUtilMethodProvider
@@ -29,6 +30,7 @@ import org.utbot.framework.codegen.domain.builtin.UtilClassFileMethodProvider
 import org.utbot.framework.codegen.domain.builtin.UtilMethodProvider
 import org.utbot.framework.codegen.domain.models.SimpleTestClassModel
 import org.utbot.framework.codegen.domain.models.CgParameterKind
+import org.utbot.framework.codegen.domain.withId
 import org.utbot.framework.codegen.services.access.Block
 import org.utbot.framework.codegen.tree.EnvironmentFieldStateCache
 import org.utbot.framework.codegen.tree.importIfNeeded
@@ -197,7 +199,7 @@ interface CgContextOwner {
     var valueByModel: IdentityHashMap<UtModel, CgValue>
 
     // use it to compare stateBefore and result variables - in case of equality do not create new variable
-    var valueByModelId: MutableMap<Int?, CgValue>
+    var valueByModelId: MutableMap<ModelId, CgValue>
 
     // parameters of the method currently being generated
     val currentMethodParameters: MutableMap<CgParameterKind, CgVariable>
@@ -225,6 +227,12 @@ interface CgContextOwner {
      * Result models required to create generic execution in parametrized tests.
      */
     var successfulExecutionsModels: List<UtModel>
+
+    /**
+     * Gives a unique identifier to model in test set.
+     * Determines which execution current model belongs to.
+     */
+    var modelIds: Map<UtModel, ModelId>
 
     fun block(init: () -> Unit): Block {
         val prevBlock = currentBlock
@@ -311,7 +319,10 @@ interface CgContextOwner {
         model?.let {
             valueByModel[it] = variable
             (model as UtReferenceModel).let { refModel ->
-                refModel.id.let { id -> valueByModelId[id] = variable }
+                refModel.id.let {
+                    val modelId = getIdByModel(model)
+                    valueByModelId[modelId] = variable
+                }
             }
         }
     }
@@ -428,6 +439,8 @@ interface CgContextOwner {
 
     val getLambdaMethod: MethodId
         get() = utilMethodProvider.getLambdaMethodMethodId
+
+    fun getIdByModel(model: UtModel): ModelId
 }
 
 /**
@@ -478,6 +491,8 @@ data class CgContext(
     override lateinit var statesCache: EnvironmentFieldStateCache
     override lateinit var actual: CgVariable
     override lateinit var successfulExecutionsModels: List<UtModel>
+
+    override var modelIds: Map<UtModel, ModelId> = mapOf()
 
     /**
      * This property cannot be accessed outside of test class file scope
@@ -556,6 +571,15 @@ data class CgContext(
         }
     }
 
+    override fun getIdByModel(model: UtModel): ModelId {
+        if (model !in this.modelIds) {
+            this.modelIds += model.withId()
+        }
+
+        return modelIds[model]
+            ?: error("ModelId for $model should have also been created")
+    }
+
     private fun createClassIdForNestedClass(testClassModel: SimpleTestClassModel): ClassId {
         val simpleName = "${testClassModel.classUnderTest.simpleName}Test"
         return BuiltinClassId(
@@ -580,7 +604,7 @@ data class CgContext(
 
     override var valueByModel: IdentityHashMap<UtModel, CgValue> = IdentityHashMap()
 
-    override var valueByModelId: MutableMap<Int?, CgValue> = mutableMapOf()
+    override var valueByModelId: MutableMap<ModelId, CgValue> = mutableMapOf()
 
     override val currentMethodParameters: MutableMap<CgParameterKind, CgVariable> = mutableMapOf()
 
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
@@ -232,7 +232,7 @@ interface CgContextOwner {
      * Gives a unique identifier to model in test set.
      * Determines which execution current model belongs to.
      */
-    var modelIds: PersistentMap<UtModel, ModelId>
+    var modelIds: MutableMap<UtModel, ModelId>
 
     fun block(init: () -> Unit): Block {
         val prevBlock = currentBlock
@@ -492,7 +492,7 @@ data class CgContext(
     override lateinit var actual: CgVariable
     override lateinit var successfulExecutionsModels: List<UtModel>
 
-    override var modelIds: PersistentMap<UtModel, ModelId> = persistentMapOf()
+    override var modelIds: MutableMap<UtModel, ModelId> = mutableMapOf()
 
     /**
      * This property cannot be accessed outside of test class file scope
@@ -573,7 +573,7 @@ data class CgContext(
 
     override fun getIdByModel(model: UtModel): ModelId {
         if (model !in modelIds) {
-            modelIds.put(model, model.withExecutionId())
+            modelIds[model] = model.withExecutionId()
         }
 
         return modelIds.getOrElse(model) {
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
@@ -30,7 +30,7 @@ import org.utbot.framework.codegen.domain.builtin.UtilClassFileMethodProvider
 import org.utbot.framework.codegen.domain.builtin.UtilMethodProvider
 import org.utbot.framework.codegen.domain.models.SimpleTestClassModel
 import org.utbot.framework.codegen.domain.models.CgParameterKind
-import org.utbot.framework.codegen.domain.withId
+import org.utbot.framework.codegen.domain.withExecutionId
 import org.utbot.framework.codegen.services.access.Block
 import org.utbot.framework.codegen.tree.EnvironmentFieldStateCache
 import org.utbot.framework.codegen.tree.importIfNeeded
@@ -232,7 +232,7 @@ interface CgContextOwner {
      * Gives a unique identifier to model in test set.
      * Determines which execution current model belongs to.
      */
-    var modelIds: Map<UtModel, ModelId>
+    var modelIds: PersistentMap<UtModel, ModelId>
 
     fun block(init: () -> Unit): Block {
         val prevBlock = currentBlock
@@ -492,7 +492,7 @@ data class CgContext(
     override lateinit var actual: CgVariable
     override lateinit var successfulExecutionsModels: List<UtModel>
 
-    override var modelIds: Map<UtModel, ModelId> = mapOf()
+    override var modelIds: PersistentMap<UtModel, ModelId> = persistentMapOf()
 
     /**
      * This property cannot be accessed outside of test class file scope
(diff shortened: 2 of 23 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/Domain.kt
```diff
@@ -191,6 +191,14 @@ abstract class TestFramework(
     abstract val nestedClassesShouldBeStatic: Boolean
     abstract val argListClassId: ClassId
 
+    abstract val beforeEachAnnotation: String
```
EgorkaKulikov, line 194: We should not forget that these annotations are different in JUnit4, JUnit5 and TestNg, so take it into account in naming, please, and locate corresponding overrides it test framework related classes.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/CgElement.kt
```diff
@@ -152,11 +153,17 @@ class CgClass(
  */
 class CgClassBody(
     val classId: ClassId,
+    val fields: List<CgElement>,
     val methodRegions: List<CgMethodsCluster>,
     val staticDeclarationRegions: List<CgStaticsRegion>,
     val nestedClassRegions: List<CgNestedClassesRegion<*>>
 ) : CgElement
 
+class CgFieldDeclaration(
```
EgorkaKulikov, line 162: Possibly we should have `classId` of the class containing field here too...

sofurihafe: Done.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/Domain.kt
sofurihafe: There were equivalents of BeforeEach annotation from JUnit5.
Renamed and removed them from abstract vals.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/TestClassModel.kt
```diff
@@ -29,7 +30,7 @@ class SpringTestClassModel(
     classUnderTest: ClassId,
     methodTestSets: List<CgMethodTestSet>,
     nestedClasses: List<SimpleTestClassModel>,
-    val injectingMocksClass: ClassId? = null,
-    val mockedClasses: Set<ClassId> = setOf(),
+    val injectedMockModels: List<UtModel> = listOf(),
```
EgorkaKulikov, line 33: May be a Set here?

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
```diff
@@ -21,27 +21,28 @@ class SpringTestClassModelBuilder: TestClassModelBuilder() {
...
-    ): Set<ClassId> {
+    ): Pair<List<UtModel>, List<UtModel>> {
+        val thisInstances = mutableListOf<UtModel>()
         val allModelsInExecution = mutableListOf<UtModel>()
 
         for (testSet in testSets) {
             for (execution in testSet.executions) {
-                execution.stateBefore.thisInstance?.let { allModelsInExecution += it }
-                execution.stateAfter.thisInstance?.let { allModelsInExecution += it }
+                execution.stateBefore.thisInstance?.let { allModelsInExecution += it; thisInstances += it }
```
EgorkaKulikov, line 44: Lets split it: one statement per line

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
```diff
@@ -53,16 +54,18 @@ class SpringTestClassModelBuilder: TestClassModelBuilder() {
...
         allModelsInExecution.forEach { model -> collectRecursively(model, allConstructedModels) }
 
-        return allConstructedModels
+        val injectedModels = thisInstances.distinct()
+        val mockedModels = allConstructedModels
             .filter { it.isMockComposite() || it.isMockAssemble() }
-            .map { it.classId }
-            .filter {  it != classUnderTest }
-            .toSet()
+            .filter { it != classUnderTest }
```
EgorkaKulikov, line 60: This check is out of date. May be something like `it !in injectedModels`

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
```diff
@@ -53,16 +54,18 @@ class SpringTestClassModelBuilder: TestClassModelBuilder() {
...
+            .distinct()
 
+        return injectedModels to mockedModels
     }
 
     private fun collectRecursively(currentModel: UtModel, allModels: MutableSet<UtModel>) {
-        if (currentModel in allModels) {
+        // TODO: come up with a better way to compare UtCompositeModels
+        if (allModels.any { it === currentModel }) {
```
EgorkaKulikov, line 68: It's seems to be ok, an `IdentityHashMap` used in `AssembleModelGenerator` is used for keys reference equality there too. I suggest to remove TODO.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/renderer/CgJavaRenderer.kt
```diff
@@ -92,6 +93,11 @@ internal class CgJavaRenderer(context: CgRendererContext, printer: CgPrinter = C
     }
 
     override fun visit(element: CgClassBody) {
+        // render class fields
+        for (field in element.fields) {
+            field.accept(this)
+        }
```
EgorkaKulikov, line 99: `println()` here ?

EgorkaKulikov:
Currently crashed on the following method from out sample project:

```java
public List<Order> getOrders(boolean flag) {
        if (flag) {
            return orderRepository.findAll();
        }

        return new ArrayList<>();
    }
```

Crash reason is:

```
        // 
        // 1 occurrences of:
        /* class org.utbot.framework.plugin.api.UtAssembleModel cannot be cast to class org.utbot.framework.plugin.api.UtCompositeModel (org.utbot.framework.plugin.api.UtAssembleModel and
        org.utbot.framework.plugin.api.UtCompositeModel are in unnamed module of loader 'app') */
```

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringVariableConstructor.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    private fun findCgValueByModel(model: UtModel, collection: Map<List<UtModel>, CgValue>): CgValue? =
+        collection
+            .filter { it.key.contains(model) }
+            .firstOrNull()
+            ?.value
+
+    companion object {
+        val mockedModelsVariables: MutableMap<List<UtModel>, CgValue> = mutableMapOf()
```
EgorkaKulikov, line 47: We should find a better way than a companion, but it does not seem obvious. Need to discuss.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -32,7 +37,11 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
 
             // TODO: support inner classes here
 
-            // TODO: create class variables with Mock/InjectMock annotations using testClassModel
+            val groupedInjectedModelsByClassId = testClassModel.injectedMockModels.groupBy { it.classId }
```
EgorkaKulikov, line 40: I think we should do this `groupBy` in a place where we prepare `SpringTestClassModel`. 
And store `Map<ClassId, List<UtModel>> in it is this structure is actually required here.

Similarly for `groupedMockedModelsByClassId`.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +86,37 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
+        groupedModelsByClassId: Map<ClassId, List<UtModel>>,
+        annotationClassId: ClassId,
+        isMock: Boolean
+    ): MutableList<CgFieldDeclaration> {
+        val constructedDeclarations = mutableListOf<CgFieldDeclaration>()
+
+        val annotation = statementConstructor.annotation(annotationClassId)
+        groupedModelsByClassId.forEach { (classId, listOfUtModels) ->
+            val name = nameGenerator.nameFrom(classId) + if (isMock) "Mock" else ""
+            val declaration =
```
EgorkaKulikov, line 99:
```java
val declaration = CgDeclaration(classId, name, initializer = null)
```

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +86,37 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
         return if (regions.any()) regions else null
     }
 
-    private fun constructMockitoCloseables(): Pair<CgDeclaration, CgMethodsCluster> {
+    private fun constructClassFields(
+        groupedModelsByClassId: Map<ClassId, List<UtModel>>,
+        annotationClassId: ClassId,
+        isMock: Boolean
```
EgorkaKulikov, line 92: It seems to be a reduntant parameter. It should be possible to take an information, should we mock or not, from `groupedModelsByClassId`.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +86,37 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
         return if (regions.any()) regions else null
     }
 
-    private fun constructMockitoCloseables(): Pair<CgDeclaration, CgMethodsCluster> {
+    private fun constructClassFields(
+        groupedModelsByClassId: Map<ClassId, List<UtModel>>,
```
EgorkaKulikov, line 90: Seems that just one UtModel will be enough here

sofurihafe: No.
Later, we do the following:
https://github.com/UnitTestBot/UTBotJava/blob/f213dded608dee0e8009908cb9d3d6dec8cbb165/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt#L110

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/TestClassModel.kt
sofurihafe: It's Map<ClassId, List<UtModel>> now. I guess, it's not relevant.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
sofurihafe: Done.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
sofurihafe: Done.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/renderer/CgJavaRenderer.kt
sofurihafe: Done.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
sofurihafe: Ok.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
sofurihafe: Done.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
sofurihafe: Removed.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +86,37 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
+            val declaration =
+                CgDeclaration(
+                    classId,
+                    name,
+                    null
+                )
+
+            constructedDeclarations += CgFieldDeclaration(declaration, annotation)
+
+            val createdValue = variableConstructor.getOrCreateVariable(listOfUtModels.first(), name)
```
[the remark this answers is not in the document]
sofurihafe: There can't be empty list by using **groupBy** method. But added a little check. Just in case.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +84,32 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
+        groupedModelsByClassId: Map<ClassId, List<UtModel>>,
+        annotationClassId: ClassId
+    ): MutableList<CgFieldDeclaration> {
+        val constructedDeclarations = mutableListOf<CgFieldDeclaration>()
+
+        val annotation = statementConstructor.annotation(annotationClassId)
+        groupedModelsByClassId.forEach { (classId, listOfUtModels) ->
+            val model = listOfUtModels.firstOrNull() ?: return@forEach
+            val name = nameGenerator.nameFrom(classId) + if (model.isMockModel()) "Mock" else ""
+            val declaration = CgDeclaration(classId, name, null)
```
Damtev, line 97: Add named arguments to constants, please. And add some comments the `null` argument is used here, please

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +84,32 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
+
+            constructedDeclarations += CgFieldDeclaration(declaration, annotation)
+
+            val createdValue = variableConstructor.getOrCreateVariable(model, name)
```
Damtev, line 101: `name` passed here is just a base name - it will be used in the name generator and does not have to be the real name of the created variable

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +84,32 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
+            when (annotationClassId) {
+                injectMocksClassId -> CgSpringVariableConstructor.injectedMocksModelsVariables += listOfUtModels to createdValue
+                mockClassId -> CgSpringVariableConstructor.mockedModelsVariables += listOfUtModels to createdValue
+                else -> error("Unexpected annotation ClassId")
```
Damtev, line 105: Please, pass the `annotationClassId` to the error description

Damtev, line 105: Minor: this condition does not depend on loop body - you can introduce a local function with a required behavior before the loop

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +84,32 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
+            }
+        }
+
+        return constructedDeclarations
+    }
+
+    private fun constructMockitoCloseables(): Pair<CgFieldDeclaration, CgMethodsCluster> {
```
Damtev, line 112: Why? These closeables were supposed to be local variables for static mocks in each test method

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringVariableConstructor.kt
```diff
@@ -0,0 +1,51 @@
...
+                )
+            }
+
+            return alreadyCreatedMock
+        }
+
+        return super.getOrCreateVariable(model, name)
+    }
+
+    private fun findCgValueByModel(model: UtModel, collection: Map<List<UtModel>, CgValue>): CgValue? =
```
Damtev, line 41: Actually, `Map` is not a collection

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringVariableConstructor.kt
```diff
@@ -0,0 +1,51 @@
...
+        collection
+            .filter { it.key.contains(model) }
```
Damtev, line 43: `contains` in a `List` requires linear time - consider using `Set` here

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringVariableConstructor.kt
```diff
@@ -0,0 +1,51 @@
...
+            .firstOrNull()
+            ?.value
+
+    companion object {
+        val mockedModelsVariables: MutableMap<List<UtModel>, CgValue> = mutableMapOf()
+        val injectedMocksModelsVariables: MutableMap<List<UtModel>, CgValue> = mutableMapOf()
```
Damtev, line 49: Please, do not use such statics - it often leads to unexpected and hard-to-catch errors

utbot-python/src/main/kotlin/org/utbot/python/framework/codegen/model/constructor/visitor/CgPythonRenderer.kt
```diff
@@ -111,6 +112,12 @@ internal class CgPythonRenderer(
         println("")
     }
 
+    override fun visit(element: CgFieldDeclaration) {
+        element.annotation?.accept(this)
+        renderVisibility(element.declaration.variableType)
+        element.declaration.accept(this)
+    }
```
Damtev, line 119: Same about override

Damtev (changes requested):

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgSpringTestClassConstructor.kt
```diff
@@ -77,7 +84,32 @@ class CgSpringTestClassConstructor(context: CgContext): CgAbstractTestClassConst
...
-    private fun constructMockitoCloseables(): Pair<CgDeclaration, CgMethodsCluster> {
+    private fun constructClassFields(
+        groupedModelsByClassId: Map<ClassId, List<UtModel>>,
+        annotationClassId: ClassId
+    ): MutableList<CgFieldDeclaration> {
+        val constructedDeclarations = mutableListOf<CgFieldDeclaration>()
+
+        val annotation = statementConstructor.annotation(annotationClassId)
+        groupedModelsByClassId.forEach { (classId, listOfUtModels) ->
+            val model = listOfUtModels.firstOrNull() ?: return@forEach
```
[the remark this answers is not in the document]
sofurihafe: **return@forEach** in this case works like **continue** in for-loop.

For example,
```kotlin
val a = listOf(1, 2, 3, 4, 5, 6, 7, 8)
a.forEach { if (it == 3) return@forEach else println(it) }
```
will produce the following output:
```kotlin
1
2
4
5
6
7
8
```

Actually, we do not expect **listOfUtModels** to be empty at all because of previous grouping by class ids. And, **listOfUtModels.first()** looks a little rough, so it was added just in case of unexpected situation.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/renderer/CgJavaRenderer.kt
```diff
@@ -388,11 +397,14 @@ internal class CgJavaRenderer(context: CgRendererContext, printer: CgPrinter = C
...
-        when {
-            classId.isPublic -> print("public ")
-            classId.isProtected -> print("protected ")
-            classId.isPrivate -> print("private ")
+    override fun renderVisibility(modifier: VisibilityModifier) {
+        when (modifier) {
+            VisibilityModifier.PUBLIC -> print("public ")
+            VisibilityModifier.PRIVATE -> print("private ")
+            VisibilityModifier.PROTECTED -> print("protected ")
+            VisibilityModifier.INTERNAL -> print("internal ")
```
Damtev, line 405: it looks like we can define corresponding literals in this enum and just render it here

EgorkaKulikov: Unfortunately it seems to be not very good idea because modifiers rendering is language specific. For example `public` in Kotlin will not be rendered at all.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/context/CgContext.kt
```diff
@@ -197,7 +199,7 @@ interface CgContextOwner {
     var valueByModel: IdentityHashMap<UtModel, CgValue>
 
     // use it to compare stateBefore and result variables - in case of equality do not create new variable
-    var valueByModelId: MutableMap<Int?, CgValue>
+    var valueByModelId: MutableMap<ModelId, CgValue>
```
Damtev, line 202: Also, this field should be cleared in the `clearClassScope` method, I guess

Damtev (changes requested):

Damtev (approved):

alisevych:
@sofurihafe @EgorkaKulikov 
Please do not forget to add ctg- and comp labels in future PRs

PR #1843, opened 2023-02-22
Added python function name to logging (#1842)

### LINKED ISSUE #1842: Add logging for starting/finishing fuzzing for specific function
**Description**

Currently from `idea.log` it's not possible to understand on which method UTBot Python works at the moment, so when I see some errors there I cannot link it to specific function under test which is problematic.

**Expected behavior**

Could we add a log entry for beginning [and ending] of test generation for each function, so the log would look something like:
```
Fuzzing started for function: use_dic(my_dict) -> int
....
Fuzzing finished for function: use_dic(my_dict) -> int
```
Exact method name in the log could be different of course, but better if it would be closer to what's written by user, so it would be easier to find.

**Potential alternatives**

We also could add function name to each log entry, when it's relevant. But this seems less pretty, as we don't always need that information every time we look in the log.

**Context**
When trying to investigate root cause of some issues, like #1830 it's needed to understand which error in the log belong to which function and which are not related.

tamarinvs19:
## Description

New log message with function name

Fixes #1842

### Manual tests

1. Generate test for python function
2. In logs with level `INFO`

## Self-check list

### DIFF
--- a/utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
+++ b/utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
@@ -110,6 +110,7 @@ class PythonTestCaseGenerator(
         var generated = 0
         val typeInferenceCancellation = { isCancelled() || System.currentTimeMillis() >= until || missingLines?.size == 0 }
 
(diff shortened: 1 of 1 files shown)


tyuldashev (approved):
LGTM

Issue #1560, opened 2022-12-20
UtBot mocks interface even when `Do not mock` is selected


tyuldashev:
**Description**
UtBot tries to mock parameter which type is interface without existing implementations. As `Do not mock` selected no Mockito dependencies are added and so generated code cannot be executed.

**To Reproduce**

1. Open some test project which has NO Mockito dependencies in it
2. Generate tests for code below, select `Do not mock` option

```java
public class CommonMocksExample {
    public Object mockInterfaceWithoutImplementors(InterfaceWithoutImplementors value) {
        return value.visit(this);
    }

}

interface InterfaceWithoutImplementors {
    Object visit(Object value);
}
```

**Expected behavior**

Generated tests have no mocks and can be executed. Or no tests are generated.

**Actual behavior**

Generated tests shown below. It contains references to Mockito `when` and `actual`, but because there is no Mockito dependency in the project, such tests cannot be executed.

```java
    @Test
    @DisplayName("mockInterfaceWithoutImplementors: InterfaceWithoutImplementorsVisit -> return value.visit(this)")
    public void testMockInterfaceWithoutImplementors_InterfaceWithoutImplementorsVisit() {
        CommonMocksExample commonMocksExample = new CommonMocksExample();
        InterfaceWithoutImplementors valueMock = mock(InterfaceWithoutImplementors.class);
        (when(valueMock.visit(any()))).thenReturn(null);

        Object actual = commonMocksExample.mockInterfaceWithoutImplementors(valueMock);

        assertNull(actual);
    }
```

**Visual proofs (screenshots, logs, images)**
![image](https://user-images.githubusercontent.com/10035839/208697991-e10645c4-6d35-4c3f-981f-71cc2188d591.png)

**Environment**
IDEA 2022.2.3
UTBot 2022.12.637 from unit-test-bot/rc16122022 branch

**Additional context**

That seems quite old bug, as it's reproducible in 2022.10.580 version of UTBot too.

tyuldashev:
@CaelmBleidd Do you think some simple fix is possible here? Or the fix would definitely require significant amount of instigation/work?

Regarding expected result: in case we have to mock in this situation then we need to add dependency to avoid red code.

CaelmBleidd:
[...]

PR #1889, opened 2023-03-03
Prune branches with unexpected mocks

EgorkaKulikov:
## Description

Fixes # ([1560](https://github.com/UnitTestBot/UTBotJava/issues/1560))

We need to stop symbolic execution if we called `forceMock` function and `Do not mock` strategy is used.
One of possible scenarios is when there are no implementations of an interface in project while `NO_MOCKS` strategy is selected.

To obtain this, some branch pruning is introduced.
Job cancellations are removed from mock listeners.

### Automated tests

Corrected test:  `testMockInterfaceWithoutImplementorsWithNoMocksStrategy`
Added test: `testMockInterfaceWithoutImplementorsWithMockingStrategy`

### Manual tests

Scenario from issue.

## Self-check list

### DIFF
--- a/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
@@ -7,14 +7,27 @@ import org.utbot.testing.DoNotCalculate
 import org.utbot.testing.UtValueTestCaseChecker
 
 internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = CommonMocksExample::class) {
+
     @Test
-    fun testMockInterfaceWithoutImplementors() {
+    fun testMockInterfaceWithoutImplementorsWithNoMocksStrategy() {
+        checkMocks(
+            CommonMocksExample::mockInterfaceWithoutImplementors,
+            eq(1),
+            { v, mocks, _ -> v == null && mocks.isEmpty() },
+            coverage = DoNotCalculate,
+            mockStrategy = MockStrategyApi.NO_MOCKS,
+        )
+    }
+
+    @Test
+    fun testMockInterfaceWithoutImplementorsWithMockingStrategy() {
         checkMocks(
             CommonMocksExample::mockInterfaceWithoutImplementors,
             eq(2),
             { v, mocks, _ -> v == null && mocks.isEmpty() },
             { _, mocks, _ -> mocks.singleOrNull() != null },
-            coverage = DoNotCalculate
+            coverage = DoNotCalculate,
+            mockStrategy = MockStrategyApi.OTHER_CLASSES,
         )
     }
 
--- a/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
@@ -3,10 +3,12 @@ package org.utbot.examples.mock
 import org.utbot.framework.plugin.api.MockStrategyApi
 import org.junit.jupiter.api.Test
 import org.utbot.testcheckers.eq
-import org.utbot.testing.DoNotCalculate
 import org.utbot.testing.UtValueTestCaseChecker
+import org.utbot.testing.atLeast
 
 internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = CommonMocksExample::class) {
+
+    //TODO: coverage values here require further investigation by experts
 
     @Test
     fun testMockInterfaceWithoutImplementorsWithNoMocksStrategy() {
@@ -14,8 +16,8 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             CommonMocksExample::mockInterfaceWithoutImplementors,
             eq(1),
             { v, mocks, _ -> v == null && mocks.isEmpty() },
-            coverage = DoNotCalculate,
             mockStrategy = MockStrategyApi.NO_MOCKS,
+            coverage = atLeast(75),
         )
     }
 
@@ -26,8 +28,8 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             eq(2),
             { v, mocks, _ -> v == null && mocks.isEmpty() },
             { _, mocks, _ -> mocks.singleOrNull() != null },
-            coverage = DoNotCalculate,
             mockStrategy = MockStrategyApi.OTHER_CLASSES,
+            coverage = atLeast(75),
         )
     }
 
@@ -40,7 +42,7 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             { fst, _, mocks, _ -> fst == null && mocks.isEmpty() },
             { _, _, mocks, _ -> mocks.isEmpty() }, // should be changed to not null fst when 1449 will be finished
             mockStrategy = MockStrategyApi.OTHER_PACKAGES,
-            coverage = DoNotCalculate
+            coverage = atLeast(75)
         )
     }
 
@@ -55,7 +57,7 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             // node == node.next
             // node.next.value == node.value + 1
             mockStrategy = MockStrategyApi.OTHER_CLASSES,
-            coverage = DoNotCalculate
+            coverage = atLeast(13)
         )
     }
 
@@ -66,7 +68,7 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             eq(1),
             { r -> r == -420 },
             mockStrategy = MockStrategyApi.OTHER_CLASSES,
-            coverage = DoNotCalculate
+            coverage = atLeast(70),
         )
     }
 
@@ -76,7 +78,7 @@ internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = Common
             CommonMocksExample::mocksForNullOfDifferentTypes,
             eq(1),
             mockStrategy = MockStrategyApi.OTHER_PACKAGES,
-            coverage = DoNotCalculate
+            coverage = atLeast(75)
         )
     }
 }
--- a/utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
@@ -160,7 +160,9 @@ class Mocker(
     private val mocksDesired: Boolean = strategy != MockStrategy.NO_MOCKS
 
     fun construct(value: ObjectValue?): MockedObjectInfo =
-        value?.let { if (mocksDesired || mockAlways(it.type) ) ExpectedMock(it) else UnexpectedMock(it) } ?: NoMock
+        value
+            ?.let { if (mocksDesired || mockAlways(it.type)) ExpectedMock(it) else UnexpectedMock(it) }
+            ?: NoMock
 
     /**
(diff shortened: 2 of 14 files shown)


utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2684,28 +2692,33 @@ class Traverser(
...
-                    val constraints = setOf(typeConstraint.isOrNullConstraint())
-
-                    // TODO add memory updated for types JIRA:1523
-                    // TODO isMock????
-                    InvocationTarget(mockedObject, method, constraints)
+                        // TODO add memory updated for types JIRA:1523
+                        InvocationTarget(wrapperOrInstance, method, constraints)
+                    }
+                    is ExpectedMock -> {
+                        val mockedObject = mockedObjectInfo.value!!
```
Damtev, line 2713: It's better to make `value` in `MockedObjectInfo` abstract and override with `non-null type in `ExpectedMock` to avoid such non-null assertion calls

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2672,10 +2680,10 @@ class Traverser(
         } ?: findMethodInvocationTargets(types, methodSubSignature)
 
         return methodInvocationTargets
-            .map { (method, implementationClass, possibleTypes) ->
+            .mapNotNull { (method, implementationClass, possibleTypes) ->
```
Damtev, line 2683: I do not see any nulls in the body

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2684,28 +2692,33 @@ class Traverser(
...
+                    is ExpectedMock -> {
+                        val mockedObject = mockedObjectInfo.value!!
+                        val typeConstraint = typeRegistry.typeConstraint(mockedObject.addr, mockedObject.typeStorage)
+                        val constraints = setOf(typeConstraint.isOrNullConstraint())
+
+                        // TODO add memory updated for types JIRA:1523
+                        // TODO isMock????
+                        InvocationTarget(mockedObject, method, constraints)
+                    }
+                    is UnexpectedMock -> unreachableBranch("If it ever happens, it should be investigated")
```
Damtev, line 2721: That behavior is very unclear. Please, add some comments to clarify.

Also, please change the message

Damtev (changes requested):
There are some suspicious parts

utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
```diff
@@ -7,14 +7,27 @@ import org.utbot.testing.DoNotCalculate
...
 internal class CommonMocksExampleTest: UtValueTestCaseChecker(testClass = CommonMocksExample::class) {
+
     @Test
-    fun testMockInterfaceWithoutImplementors() {
+    fun testMockInterfaceWithoutImplementorsWithNoMocksStrategy() {
+        checkMocks(
+            CommonMocksExample::mockInterfaceWithoutImplementors,
+            eq(1),
+            { v, mocks, _ -> v == null && mocks.isEmpty() },
+            coverage = DoNotCalculate,
```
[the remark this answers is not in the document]
EgorkaKulikov: Actually, all tests in this file are marked with `DoNotCalculate`, I just added similarly.
I tried to avoid this setting, both newly created and modified tests fail with a message

```
Coverage matcher 'full coverage' fails for (branches: 0/0, instructions: 3/4, lines: 0/0) (at least: 75) ==> expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:46)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:187)
```

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -137,21 +157,56 @@ class Mocker(
...
+    fun mock(type: RefType, mockInfo: UtMockInfo): MockedObjectInfo {
+        val objectValue = if (shouldMock(type, mockInfo)) createMockObject(type, mockInfo) else null
+        return construct(objectValue, mockInfo)
+    }
 
     /**
-     * Creates mocked instance of the [type] using mock info. Unlike to [mock], it does not
-     * check anything and always returns the constructed mock.
+     * Creates mocked instance of the [type] using [mockInfo]
+     * it does not check anything and always returns the constructed mock.
```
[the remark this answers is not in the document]
EgorkaKulikov: Why?

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -1470,7 +1472,13 @@ class Traverser(
             }
 
             val mockInfo = mockInfoGenerator.generate(addr)
-            val mockedObject = mocker.forceMock(type, mockInfoGenerator.generate(addr))
+            val mockedObjectInfo = mocker.forceMock(type, mockInfoGenerator.generate(addr))
+
+            val mockedObject = mockedObjectInfo.value ?: error("Mocked value cannot be null after force mock")
```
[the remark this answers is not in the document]
EgorkaKulikov: No... objectValue is never null in `forceMock` method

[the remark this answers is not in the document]
EgorkaKulikov: Anyway, I've rewritten this place

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
EgorkaKulikov: I will discuss with Alexey

utbot-framework-test/src/test/kotlin/org/utbot/examples/mock/CommonMocksExampleTest.kt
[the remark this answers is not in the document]
Damtev: Well, in such cases we can use `AtLeast(75)`

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -128,6 +129,34 @@ data class UtStaticMethodMockInfo(
     val methodId: MethodId
 ) : UtMockInfo(methodId.classId, addr)
 
+/**
+ * A wrapper for [ObjectValue] to store additional onfo.
```
CaelmBleidd, line 133: Typo? `info`

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -128,6 +129,34 @@ data class UtStaticMethodMockInfo(
...
+    abstract val value: ObjectValue?
+}
+
+object NoMock: MockedObjectInfo() {
+    override val value: ObjectValue? = null
+}
+
+/**
+ * Represents a mock that occurs when mock strategy allows it
+ * or when an object type is in a set that requires force mocking.
```
CaelmBleidd, line 145: Please, add a link using `[]` to this set

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -128,6 +129,34 @@ data class UtStaticMethodMockInfo(
...
+ */
+class ExpectedMock(objectValue: ObjectValue): MockedObjectInfo() {
+    override val value: ObjectValue = objectValue
+}
+
+/**
+ * Represents a mock that occurs when it is not recommended
```
CaelmBleidd, line 152: I think, `not recommended` is not exactly the condition. Probably, `when it is not allowed` would be better, since it is strictly prohibited to generate mocks in such cases

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -128,6 +129,34 @@ data class UtStaticMethodMockInfo(
...
+ * E.g. mock framework is not installed or
+ * mock startegy is DO_NOT_MOCK and class is not in mockAlways set.
```
CaelmBleidd, line 154: Use `[]` to add a link to the set and to the `DO_NOT_MOCK`, please

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -137,21 +166,56 @@ class Mocker(
...
 ) {
+    private val mocksAreDesired: Boolean = strategy != MockStrategy.NO_MOCKS
+
     /**
-     * Creates mocked instance of the [type] using mock info if it should be mocked by the mocker,
-     * otherwise returns null.
+     * Constructs [MockedObjectInfo]: enriches given value with
+     * an information if this mock is expected or not.
+     */
+    fun construct(value: ObjectValue?, mockInfo: UtMockInfo): MockedObjectInfo {
```
CaelmBleidd, line 177: Shouldn't it be a private function? What is a case of usage it from outside?

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
```diff
@@ -137,21 +166,56 @@ class Mocker(
...
+        if (value == null) {
```
CaelmBleidd, line 178: I guess, if you name it as `mockedValue`, it will be easier to understand

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -1470,7 +1475,20 @@ class Traverser(
             }
 
             val mockInfo = mockInfoGenerator.generate(addr)
-            val mockedObject = mocker.forceMock(type, mockInfoGenerator.generate(addr))
+            val mockedObjectInfo = mocker.forceMock(type, mockInfoGenerator.generate(addr))
+
+            val mockedObject: ObjectValue = when (mockedObjectInfo) {
+                is NoMock -> error("Value must be mocked after the fore mock")
```
CaelmBleidd, line 1481: Typo, `force`

CaelmBleidd (changes requested):

utbot-framework/src/main/kotlin/org/utbot/engine/Mocks.kt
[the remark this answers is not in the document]
CaelmBleidd: Added a proposition about possible documentation. Information about checks is still correct because added checks are not about the need to mock, but about whether they are expected or not

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
CaelmBleidd: If it is unclear, you can provide a comment. Something like: 


Currently, it is unclear how this could happen. Perhaps, the answer is somewhere in the following situation: you have an interface with an abstract method `foo`, and it has an abstract inheritor with the implementation of the method, but this inheritor doesn't have any concrete inheritors. It looks like in this case we would mock this instance (because it doesn't have any possible concrete type), but it is impossible since either this class cannot present in possible types of the object on which we call `foo` (since they contain only concrete types), or this class would be already mocked (since it doesn't contain any concrete implementors).


Something like this. But tbh I think it can be just replaced with `-> TODO("If it ever happens, it should be investigated")`

CaelmBleidd (approved):
LGTM

Damtev (approved):

PR #1893, opened 2023-03-03
Update time management of python test generation

### LINKED ISSUE #1831: UTBot Python cannot find Division by Zero problem
**Description**

Even when all types are specified UTBot Python doesn't find case when Division by Zero happens. Both UTBotJava and UTBotCPP can find such problem and it seems quite classic example of the problem which could be revealed with static analysis, so I would expect that UTBot Python also find the error.

**To Reproduce**

```python
def div(a: int, b: int) -> int:
    return a / b
```

1. Invoke code generation on code above

**Expected behavior**

Test which exposes 'Division by Zero' problem is generated among the others.

**Actual behavior**

No such test is generated, the only test looks like that:
```python
    def test_div(self):
        """
        a = 4
        b = 111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
        """
        actual = basics.div(4, 111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111)
        
        self.assertEqual(0.0, actual)
```

### LINKED ISSUE #1874: Allow UTBotPython to proceed test generation till timeout when function has full annotation
**Description**

When we have only one type we can use all time to generate tests for this annotation instead of splitting time for other inferred annotations (and wasting time).

**Expected behavior**

If function is annotated and all types are full (no one missed part, no one `typing.Any`, no one generic, no one protocol) UtBot will proceed test generation till timeout.

**Potential alternatives**

In future we can make option for this feature and add different modes of test generation:
* cover all branches and exit
* use all possible time

tamarinvs19:
## Description

New test generation strategy is:
* If function is fully annotated and type inference algorithm cannot infer type we use all time from UI-settings to fuzzing and test generation; 
* If type inference and infer type we try to fuzz tests for one annotation with the following restrictions:
    - max 150 successfully executable tests (with or without exeception)
    - max 10 invalid tests (when we generate incorrect arguments and have `TypeError` or `AttributeError`; or we cannot generate any valid representative of the inferred type)
    - and 5 additional tests (as a precaution)
    - until time limit

In future we can modify strategies and add it to UI. For example, wait max coverage.

Fixes #1874
Fixes #1831

#### Fully annotated function
For example,
```python
def div(a: int, b: int) :
    return a / b
```
__Expected__: two tests (valid and with `ZeroDivisionError`), test generation process lasted all time from UI-settings

Or,
```python
def str_test(x: str):
    if x[0] == 'b':
        if x[1] == 'a':
            if x[2] == 'd':
                return 'Very bad!'
```
__Expected__: all time generation, one test with `x = 'ba'` or longer (with timeout 60s) and all covered branches with timeout 180s

#### Without or partially annotated function
For example,
```python
def str_test(x):
    if x[0] == 'b':
        if x[1] == 'a':
            if x[2] == 'd':
                return 'Very bad!'
```
__Expected__: all time generation, one test with `x = 'b'`, no one with `'ba'` and longer (__regardless of timeout__)

## Self-check list


utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
```diff
@@ -277,14 +263,15 @@ class PythonTestCaseGenerator(
...
-            if (isCancelled()) {
+            if (typeInferenceCancellation()) {
                 return@breaking
             }
 
-            algo.run(hintCollector.result, isCancelled, annotationHandler)
+            algo.run(hintCollector.result, typeInferenceCancellation, annotationHandler)
 
             val existsAnnotation = method.definition.type
             if (existsAnnotation.arguments.all { it.pythonTypeName() != "typing.Any" }) {
```
tochilinak, line 273: Seems like this is wrong. Example:
```python
def f(x: list): pass
```

In this case we need type inference for intermal type.

We can get information about whether function signature has Any's inside type inference algorithm. In that case initial state has no Any nodes, and the main loop makes only one iteration (initial state -> no new states -> finish).

tochilinak (changes requested):
Problem with setting `limitManager.mode = TimeoutMode`.

tamarinvs19:
> Problem with setting `limitManager.mode = TimeoutMode`.

Fixed

utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
```diff
@@ -277,14 +263,15 @@ class PythonTestCaseGenerator(
...
 
-            algo.run(hintCollector.result, isCancelled, annotationHandler)
+            val iterationNumber = algo.run(hintCollector.result, typeInferenceCancellation, annotationHandler)
 
-            val existsAnnotation = method.definition.type
-            if (existsAnnotation.arguments.all { it.pythonTypeName() != "typing.Any" }) {
+            if (iterationNumber == 1) {
+                limitManager.mode = TimeoutMode
+                val existsAnnotation = method.definition.type
                 annotationHandler(existsAnnotation)
```
tochilinak, line 275: Initially this code was needed to run fuzzing with initial types (even when those are not full). Will we run fuzzing with `list[Any]` for the following function?
```python
def f(x: list): ...
```

tamarinvs19: Maybe. But only if all arguments have not-`typing.Any` annotations. And if all annotations are full we run fuzzing twice

tochilinak: Are you sure? Type inference does not run annotationHandler on initial signature.

tamarinvs19: Now I run fuzzing with initial annotations only after type inference) But I can change program and run fuzzing twice with initial annotations, but I think it isn't a good solution

tamarinvs19: I added handling of an initial annotation after first expansion in BaselineAlgorithm. 
- if annotation is full we cannot expand it and stop type inference and then start fuzzing;
- if annotation is not full we try to expand and first of all try to fuzz initial annotation (for example, if it is `list[typing.Any]` we can generate empty list), after that we continue type inference

tochilinak (approved):

utbot-python/src/main/kotlin/org/utbot/python/PythonEngine.kt
```diff
@@ -141,17 +140,20 @@ class PythonEngine(
...
+            val manager = try {
+                PythonWorkerManager(
+                    serverSocket,
+                    pythonPath,
+                    until,
+                    { constructEvaluationInput(it) },
+                    timeoutForRun.toInt()
+                )
+            } catch (_: TimeoutException) {
+                return@flow
```
Markoutte, line 155: Should be logged somehow

utbot-python/src/main/kotlin/org/utbot/python/PythonEngine.kt
```diff
@@ -195,7 +197,6 @@ class PythonEngine(
 
                         is PythonEvaluationSuccess -> {
                             val coveredInstructions = evaluationResult.coverage.coveredInstructions
-                            coveredInstructions.forEach { coveredLines.add(it.lineNumber) }
 
                             val summary = arguments
                                 .zip(methodUnderTest.arguments)
```
Markoutte, line 202: On line 215 (sorry, that I comment another line, because I was not able to find how to comment 215) there's no check that:

```
is ValidExecution -> {
  val trieNode: Trie.Node<Instruction> = description.tracer.add(coveredInstructions)
```

Returns the node the was already found. You can check `trieNode.count > 1` to ignore duplicates and minimise the count of tests for user. At the moment similar tests are generated with different arguments, but same trace.

Markoutte (changes requested):

utbot-python/src/main/kotlin/org/utbot/python/PythonEngine.kt
tamarinvs19: Logging was already there but one level higher. I moved it here

utbot-python/src/main/kotlin/org/utbot/python/PythonEngine.kt
tamarinvs19: Fixed

Markoutte (approved):

PR #1900, opened 2023-03-06
Remove old fuzzer dependencies from JavaScript module

### LINKED ISSUE #1792: Replacing fuzzer dependency for JS, Python and Go modules
**Description**

The JS, Python, and Go modules currently depend on `api(project(":utbot-fuzzers"))`. Looks like this dependency should be replaces to `implementation(project(":utbot-fuzzing"))`. In the future the better option is to get rid of all utbot-fuzzer module's classes. 

It is assumed that utbot-fuzzers will be modified and used only for Java/Kotlin

**Expected behavior**

The utbot-js, utbot-python and utbot-go modules depend on `implementation(project(":utbot-fuzzing"))` and do not contain imports from `org.utbot.fuzzer`

**Context**

Much of the `utbot-fuzzers` module is specific to Java, for the most part, the JS, Python and Go modules use only `FuzzedValue` and `UtFuzzedExecution`.

rudolf101:
## Description

Now `utbot-js` module depends on `utbot-fuzzing` and do not contain any imports from `org.utbot.fuzzer`

Fixes #1792 

### Manual tests

Various scenarios have been tested. You can find the examples in the folder `/utbot-js/samples' and try to run UTBot.

## Self-check list

### DIFF
--- a/utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
+++ b/utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
@@ -1,6 +1,7 @@
 package fuzzer
 
 import framework.api.js.JsClassId
+import framework.api.js.JsUtFuzzedExecution
 import framework.api.js.util.isClass
 import org.utbot.framework.plugin.api.UtTimeoutException
 import org.utbot.fuzzer.FuzzedConcreteValue
@@ -13,7 +14,7 @@ import org.utbot.fuzzing.Feedback
 import org.utbot.fuzzing.utils.Trie
 
 sealed interface JsFuzzingExecutionFeedback
-class JsValidExecution(val utFuzzedExecution: UtFuzzedExecution) : JsFuzzingExecutionFeedback
+class JsValidExecution(val utFuzzedExecution: JsUtFuzzedExecution) : JsFuzzingExecutionFeedback
 
 class JsTimeoutExecution(val utTimeout: UtTimeoutException) : JsFuzzingExecutionFeedback
 
--- a/utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
+++ b/utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
@@ -95,5 +95,5 @@ fun UtModel.fuzzed(block: JsFuzzedValue.() -> Unit = {}): JsFuzzedValue = JsFuzz
 object JsIdProvider {
     private var id = AtomicInteger(0)
 
-    fun get() = id.incrementAndGet()
+    fun createId() = id.incrementAndGet()
 }
--- a/utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
+++ b/utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
@@ -3,15 +3,14 @@ package fuzzer
 import framework.api.js.JsClassId
 import framework.api.js.JsUtFuzzedExecution
 import framework.api.js.util.isClass
+import org.utbot.framework.plugin.api.ClassId
+import org.utbot.framework.plugin.api.UtModel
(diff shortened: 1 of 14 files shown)


utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
```diff
@@ -57,3 +58,42 @@ class JsFeedback(
 data class JsStatement(
     val number: Int
 )
+
+data class JsFuzzedValue(
```
Markoutte, line 62: Look like JsFuzzedValue is not really needed by the code, because the filed `summary` is never read and used. Only the `model` is required by the further code as I can see.

utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
```diff
@@ -57,3 +58,42 @@ class JsFeedback(
...
+        NE -> EQ
+        GT -> LE
+        LT -> GE
+        LE -> GT
+        GE -> LT
+        Unknown -> Unknown
+    }
+}
+
+fun UtModel.fuzzed(block: JsFuzzedValue.() -> Unit = {}): JsFuzzedValue = JsFuzzedValue(this).apply(block)
```
Markoutte, line 93: Taking into account the previous comment you can remove this because UtModel can be used directly by the js codegen.

utbot-js/src/main/kotlin/fuzzer/providers/ObjectValueProvider.kt
```diff
@@ -31,10 +29,13 @@ class ObjectValueProvider : ValueProvider<JsClassId, FuzzedValue, JsMethodDescri
...
 
-    private fun createValue(classId: JsClassId, constructorId: JsConstructorId): Seed.Recursive<JsClassId, FuzzedValue> {
+    private fun createValue(
+        classId: JsClassId,
+        constructorId: JsConstructorId
+    ): Seed.Recursive<JsClassId, JsFuzzedValue> {
         return Seed.Recursive(
             construct = Routine.Create(constructorId.parameters) { values ->
-                val id = idGenerator.createId()
+                val id = JsIdProvider.get()
```
Markoutte, line 38: `createId` is clearer than `get`. When calling method with name `get` usually everyone would expect that it returns same value for every call.

Markoutte (changes requested):

utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
```diff
@@ -42,7 +43,7 @@ class JsMethodDescription(
 class JsFeedback(
     override val control: Control = Control.CONTINUE,
     val result: Trie.Node<JsStatement> = Trie.emptyNode()
-) : Feedback<JsClassId, FuzzedValue> {
+) : Feedback<JsClassId, JsFuzzedValue> {
 
     override fun equals(other: Any?): Boolean {
         val castOther = other as? JsFeedback
```
Markoutte, line 49: This equals/hashCode is not good for fuzzing, because it groups result only by status, but doesn't use JsStatement. Please, look at java and pythod feedbacks. They look like this:

```
data class PythonFeedback(
    override val control: Control = Control.CONTINUE,
    val result: Trie.Node<Instruction> = Trie.emptyNode(),
) : Feedback<Type, PythonFuzzedValue>
```

utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
rudolf101: Removed JsFuzzedValue

utbot-js/src/main/kotlin/fuzzer/JsFuzzerApi.kt
rudolf101: Fixed

Markoutte (approved):

PR #1922, opened 2023-03-10
Some fixes in Python fuzzer logic

tochilinak:
## Description

Changes:
- New logic for functions with default arguments: first try to fuzz only positional arguments, then all of them.
- Added new Python tree node: FakeNode (instead of Java `null`). If fuzzer provides object with this node, we don't start execution with this data.
- BaselineAlgorithm now provides random type each 6 iterations.
- Logic for abstract classes is now the same as for protocols.

### Manual tests

Default arguments: https://github.com/TheAlgorithms/Python/blob/master/machine_learning/gradient_descent.py (function `_error`).

Random type: https://github.com/TheAlgorithms/Python/blob/master/data_structures/queue/circular_queue.py  (function `enqueue` of class `CircularQueue`). Mypy does not pass any types for `data` parameter, bacause it consideres `self.array` to be `list[None]`.

Test for abstract class:

```python
from typing import AbstractSet

def id_(x: AbstractSet[int]):
    return x
```

Generated output:
```python
import sys
sys.path.append(r'../../../utbot-python/samples/easy_samples')
import builtins
import general
import unittest


class TestTopLevelFunctions(unittest.TestCase):
    # region Test suites for executable general.id_
    
    # region FUZZER
    
    def test_id_(self):
        """
        x = builtins.set[typing.Any]
        """
        actual = general.id_(set())
        
        self.assertEqual(set(), actual)
    # endregion
    
    # endregion
```

## Self-check list

### DIFF
--- a/utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
+++ b/utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
@@ -29,6 +29,7 @@ import org.utbot.python.utils.TimeoutMode
 import java.io.File
 
 private val logger = KotlinLogging.logger {}
+private const val RANDOM_TYPE_FREQUENCY = 6
 
 class PythonTestCaseGenerator(
     private val withMinimization: Boolean = true,
@@ -220,9 +221,11 @@ class PythonTestCaseGenerator(
                 originalDef.meta.name,
                 originalDef.meta.args.take(shortType.arguments.size)
             )
-            val additionalVars = originalDef.meta.args.drop(shortType.arguments.size).fold("") { acc, arg ->
-                acc + "\n" + "${arg.name}: ${pythonAnyType.pythonTypeRepresentation()}"  // TODO: better types
-            }
+            val additionalVars = originalDef.meta.args
+                .drop(shortType.arguments.size)
+                .joinToString(separator="\n", prefix="\n") { arg ->
+                    "${arg.name}: ${pythonAnyType.pythonTypeRepresentation()}"  // TODO: better types
+                }
             method.definition = PythonFunctionDefinition(shortMeta, shortType)
             val missingLines = methodHandler(method, typeStorage, coveredLines, errors, executions, null, firstUntil, additionalVars)
             method.definition = originalDef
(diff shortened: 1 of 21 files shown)


Markoutte (approved):

utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
```diff
@@ -196,6 +196,40 @@ class PythonTestCaseGenerator(
...
+            val firstUntil = (until - now) / 2 + now
+            val originalDef = method.definition
+            val shortType = meta.removeNonPositionalArgs(originalDef.type)
+            val shortMeta = PythonFuncItemDescription(
+                originalDef.meta.name,
+                originalDef.meta.args.take(shortType.arguments.size)
+            )
+            val additionalVars = originalDef.meta.args.drop(shortType.arguments.size).fold("") { acc, arg ->
+                acc + "\n" + "${arg.name}: ${pythonAnyType.pythonTypeRepresentation()}"  // TODO: better types
+            }
```
tamarinvs19, line 225: I suggest to simplify this code:
```kotlin
val additionalVars = originalDef.meta.args
                .drop(shortType.arguments.size)
                .joinToString(separator="\n", prefix="\n") { arg ->
                "${arg.name}: ${pythonAnyType.pythonTypeRepresentation()}"  // TODO: better types
            }
```

utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
```diff
@@ -257,7 +292,9 @@ class PythonTestCaseGenerator(
                 getOffsetLine(sourceFileContent, method.ast.beginOffset),
                 getOffsetLine(sourceFileContent, method.ast.endOffset)
             ),
-            mypyConfigFile
+            mypyConfigFile,
+            additionalVars,
+            randomTypeFrequency = 6
```
tamarinvs19, line 297: Maybe move 6 to constant?

tochilinak: Fixed

utbot-python/src/main/kotlin/org/utbot/python/PythonTestCaseGenerator.kt
tochilinak: Fixed

tamarinvs19 (approved):

PR #1957, opened 2023-03-15
Refactor name and module management in python code generation and fuzzing

### LINKED ISSUE #1850: Tests are not generated for classes that have inner classes
**Description**

Tests are not generated for classes that have inner classes.

**To Reproduce**

1. Add this code to Python project:

```python

class Outer:
    class Inner:
        a = 1

        def inc(self):
            self.a += 1

    def __init__(self):
        self.inner = Outer.Inner()

    def inc1(self):
        self.inner.inc()
```

2. Use plugin to generate tests

**Expected behavior**

Tests are supposed to be generated.

**Actual behavior**

Tests were not generated.

When test generation is started from CLI with `--verbosity=DEBUG` key, the following error is shown:

```
13:57:44.822 | DEBUG | PythonEngine | Traceback (most recent call last):
  File "/tmp/UTBot/python-test-generation-07779351610767258346/tmp_27_run_inc1.py", line 5, in <module>
    import easy_samples.sample_classes.Outer
ModuleNotFoundError: No module named 'easy_samples.sample_classes.Outer'; 'easy_samples.sample_classes' is not a package
```

**Additional context**

We have incorrect logic of getting module name for PythonClassId in `utbot-python/src/main/kotlin/org/utbot/python/framework/api/python/util/PythonIdUtils.kt:4`:

```kotlin
fun moduleOfType(typeName: String): String? {
    val lastIndex = typeName.lastIndexOf('.')
    return if (lastIndex == -1) null else typeName.substring(0, lastIndex)
}
```

In this case it splits `easy_samples.sample_classes.Outer.Inner` into (`easy_samples.sample_classes.Outer`, `Inner`), but should be (`easy_samples.sample_classes`, `Outer.Inner`).

`utbot_mypy_runner` splits this correctly.

tamarinvs19:
## Description

Module and name of python object have been separated to different fields in `PythonClassId`. Now we have two constructors:
* from module and object name (without module prefix)
* from full name (it expect that object name doesn`t contain dots)

Text type representations have been replaced to constants from util file.

Also fixed problem with objects which have same id and name but different fields in state after / state before. 

Fixes #1850

### Manual tests

See #1850

## Self-check list


utbot-python/src/main/kotlin/org/utbot/python/framework/api/python/PythonApi.kt
```diff
@@ -74,7 +74,7 @@ class PythonTreeModel(
 
     override fun equals(other: Any?): Boolean {
         if (other is PythonTreeModel) {
-            return tree == other.tree
+            return tree.softEquals(other.tree)
```
tochilinak, line 77: I think this might lead to infinite recursion

tochilinak (approved):

utbot-python/src/main/kotlin/org/utbot/python/utils/RequirementsUtils.kt
```diff
@@ -1,12 +1,12 @@
 package org.utbot.python.utils
 
 object RequirementsUtils {
-    val requirements: List<String> =
-        RequirementsUtils::class.java.getResource("/requirements.txt")
-            ?.readText()
-            ?.split('\n')
-            ?.filter { it.isNotEmpty() }
-            ?: error("Didn't find /requirements.txt")
+    val requirements: List<String> = listOf(
```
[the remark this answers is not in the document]
tamarinvs19: Fixed

utbot-python/src/main/kotlin/org/utbot/python/framework/api/python/PythonApi.kt
tamarinvs19: Replaced by special method `comparePythonTree`

PR #1967, opened 2023-03-16
Rider plugin bugfixes

mxprshn:
- Check for installed .NET SDKs instead of runtimes, as they are necessary for publish
- Remove logging from popup notification method
- Fix RPC timeout for test generation
- Use -a flag instead of -r in Publisher because incorrect RIDs appeared


Domonion (approved):
discuss

Domonion (approved):
lgtm

PR #2013, opened 2023-03-22
change build naming #1981

### LINKED ISSUE #1981: Improve build and publish plugin workflow
**Description**

We need to patch old release and publish its new version.

**Expected behavior**

In order to achieve the purpose we need:
- To have possibility to build plugin with an arbitrary version (year, month, minor version).

  Build naming should follow the naming convention: 
  [yyyy.mm.\<minor-version-indicator\>] where:
  - month doesn't contain leading zeros
  - minor version is not added to the first release in the month
  - minor version starts from 1
  
  We want to use default version name (as it's being calculated now) in most cases.
  
- We also want to have possibility to build from an arbitrary commit (by hash).

**Environment**

GitHub Actions

**Potential alternatives**

No alternatives.

**Context**

No context.

bissquit:
## Description

Fixes #1981

Add build options to set custom year, month and commit hash to build plugin

### Automated tests

Run different input combination from branch and compare results with the same runs from main

### Manual tests

Not applicable

## Self-check list

- [ ] Check minor releases
- [ ] Check different postfix
- [ ] Check how version with custom year and month is generated

### DIFF
--- a/.github/workflows/publish-plugin-and-cli-from-branch.yml
+++ b/.github/workflows/publish-plugin-and-cli-from-branch.yml
@@ -18,6 +18,18 @@ on:
           description: "Upload artifacts or not"
           required: false
           default: false
+        commit_sha:
+          required: false
+          type: string
+          description: "(optional) Commit SHA"
+        custom_year:
+          required: false
+          type: string
+          description: "(optional) Year"
+        custom_month:
+          required: false
+          type: string
+          description: "(optional) Month"
 
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


victoriafomina (approved):

alisevych (approved):
LGTM

alisevych (approved):
That's great. Thank you!

.github/workflows/publish-plugin-and-cli-from-branch.yml
```diff
@@ -3,45 +3,23 @@ name: "[M] Plugin and CLI: publish as archives"
...
-        version-postfix:
-          type: string
-          description: "It adds postfix (alpha or beta) to version (optional)."
-          required: false
-          default: no-postfix
         upload-artifact:
           type: string
           description: "Upload artifacts or not"
           required: false
           default: false
```
alisevych, line 10: Can we have `default: true`? Because we manually run the workflow in order to generate plugin.

bissquit: It's actually `true` by default for `workflow_dispatch` (manual workflow run). I suppose you've checked `workflow_call` section

PR #2024, opened 2023-03-23
Fix casts in overflow checking function

### LINKED ISSUE #2022: Test generation is cancelled due to JVM crash when overflow detection is on
**Description**

When option "Treat overflow as errors" is on than in some cases test generation is aborted due to JVM crash.

Bad code (will crash):

```java
void test(byte data )
{
    byte result = (byte)(data + 1);
    System.out.println(result);
}
```

Good code (does not crash):

```java
void wonderwuffle(int data )
{
    int result = (int) (data + 1L);
    System.out.println(result);
}
```

**To Reproduce**

1. Enable option "Treat overflow as errors"
2. Enable only symbolic execution test generation
3. Use code above
4. Try to generate test

**Expected behavior**

Some tests are generated.

**Actual behavior**

Test generation is cancelled due to JVM crash and hs_err_pid*.log file is generated.

**Environment**

No special environment.

**Additional context**

Looks like that problem in method `TraverserKt.intOverflowCheck`.

CaelmBleidd:
## Description

Added an explicit cast for arguments of the check function. They were wrapped into a primitive value that didn't change their symbolic type. That led to an error where we process primitive values of different sorts.

Fixes #2022 

### Automated tests

`org.utbot.examples.math.OverflowAsErrorTest#testByteWithIntOverflow`

### Manual tests

I tested the example from the issue and a similar one. Note that an example from the issue doesn't contain overflow since `+` casts both operands to an int type, therefore, since the second one is a constant and the first one is a byte value, an integer overflow can never occur. If we transform the second value into a symbolic value, we'll find an overflow error.

image

## Self-check list

### DIFF
--- a/utbot-framework-test/src/test/kotlin/org/utbot/examples/math/OverflowAsErrorTest.kt
+++ b/utbot-framework-test/src/test/kotlin/org/utbot/examples/math/OverflowAsErrorTest.kt
@@ -70,6 +70,22 @@ internal class OverflowAsErrorTest : UtValueTestCaseChecker(
     }
 
     @Test
+    fun testByteWithIntOverflow() {
+        withTreatingOverflowAsError {
+            checkWithException(
+                OverflowExamples::byteWithIntOverflow,
+                eq(2),
+                { x, y, r ->
+                    runCatching {
+                        Math.addExact(x.toInt(), y)
+                    }.isFailure && r.isException<OverflowDetectionError>()
+                },
+                { x, y, r -> Math.addExact(x.toInt(), y).toByte() == r.getOrThrow() }
+            )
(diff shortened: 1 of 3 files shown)


Markoutte (approved):
I think we need discuss more about this feature, because, it can have strange behavior from user point of view. For example, we have this test be reported as it has overflow, but in real life it does not:

```
    public static int foo(byte a, short b) {
        return a + b;
    }
```

Despite this, I think this fix should be merged as is, because it fixes critical problem.

CaelmBleidd:
> I think we need discuss more about this feature, because, it can have strange behavior from user point of view

Agree, created an issue for it

Issue #2034, opened 2023-03-24
Introduce more complicated NPE analysis for Spring applications


EgorkaKulikov:
**Description**

Let's generate tests for the following code snippet 

```java
  public UserServiceImpl(UserDao userDao) {
        this.userDao = userDao;
    }

    @Transactional
    @Override
    public void add(User user) {
        userDao.add(user);
    }
```

The following test is generated

```java
@Test
  public void testListUsers_UserDaoListUsers_1() {
      /* This test fails because method [web.service.UserServiceImpl.listUsers] produces [java.lang.NullPointerException]
          web.service.UserServiceImpl.listUsers(UserServiceImpl.java:28) */
      userServiceImpl.listUsers();
  }
```

Actually, this test passes :)
And this behavior is correct because NPE is impossible on objects injected with `Mock` and `InjectMocks`.

**Expected behavior**

NPE tests should be generated only when this exception is really possible. We need to tune Symbolic engine for that.

PR #2052, opened 2023-03-27
Avoid false NPE tests for Spring classes

EgorkaKulikov:
## Description

Fixes # ([2034](https://github.com/UnitTestBot/UTBotJava/issues/2034))

This PR enlarges the speculative not null concept to Spring applications

### Manual tests

Check that there are no tests with the message "this test fails ..." that are actually passing in Spring applications.

## Self-check list

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -1275,7 +1275,7 @@ class SpringApplicationContext(
     // Classes representing concrete types that are actually used in Spring application
     private val springInjectedClasses: Set<ClassId>
         get() {
-            if (!areInjectedClassesInitialized && springInjectedClassesStorage.isEmpty()) {
+            if (!areInjectedClassesInitialized) {
                 springInjectedClassesStorage += beanQualifiedNames
                     .map { fqn -> utContext.classLoader.loadClass(fqn) }
                     .filterNot { it.isAbstract || it.isInterface || it.isLocalClass || it.isMemberClass && !it.isStatic }
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -1270,10 +1270,11 @@ class SpringApplicationContext(
     private val shouldUseImplementors: Boolean,
 ): ApplicationContext(mockInstalled, staticsMockingIsConfigured) {
 
+    // Classes representing concrete types that are actually used in Spring application
     private val springInjectedClasses: Set<ClassId>
         get() {
             if (springInjectedClassesStorage.isEmpty()) {
-                springInjectedClassesStorage = beanQualifiedNames
+                springInjectedClassesStorage += beanQualifiedNames
                     .map { fqn -> utContext.classLoader.loadClass(fqn) }
                     .filterNot { it.isAbstract || it.isInterface || it.isLocalClass || it.isMemberClass && !it.isStatic }
                     .mapTo(mutableSetOf()) { it.id }
@@ -1282,7 +1283,9 @@ class SpringApplicationContext(
             return springInjectedClassesStorage
         }
 
-    private var springInjectedClassesStorage = mutableSetOf<ClassId>()
+    // This is a service field to model the lazy behavior of [springInjectedClasses].
+    // Do not call it outside the getter.
+    private val springInjectedClassesStorage = mutableSetOf<ClassId>()
 
     override val typeReplacementMode: TypeReplacementMode
         get() = if (shouldUseImplementors) KnownImplementor else NoImplementors
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -1270,10 +1270,12 @@ class SpringApplicationContext(
     private val shouldUseImplementors: Boolean,
 ): ApplicationContext(mockInstalled, staticsMockingIsConfigured) {
 
+    private var areInjectedClassesInitialized : Boolean = false
+
     // Classes representing concrete types that are actually used in Spring application
     private val springInjectedClasses: Set<ClassId>
         get() {
-            if (springInjectedClassesStorage.isEmpty()) {
+            if (!areInjectedClassesInitialized && springInjectedClassesStorage.isEmpty()) {
                 springInjectedClassesStorage += beanQualifiedNames
                     .map { fqn -> utContext.classLoader.loadClass(fqn) }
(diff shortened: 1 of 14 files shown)


utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -1244,13 +1269,20 @@ class SpringApplicationContext(
...
-    private val springInjectedClasses: List<ClassId> by lazy {
-        beanQualifiedNames
-            .map { fqn -> utContext.classLoader.loadClass(fqn) }
-            .filterNot { it.isAbstract || it.isInterface || it.isLocalClass || it.isMemberClass && !it.isStatic }
-            .map { it.id }
-    }
+
+    private val springInjectedClasses: Set<ClassId>
+        get() {
+            if (springInjectedClassesStorage.isEmpty()) {
```
Damtev, line 1275: Also, there is a possible performance problem - for example, there are a lot of `beanQualifiedNames` **but** all of them are filtered out, so this condition is always true, and we will never use the `springInjectedClassesStorage` field. I'm not sure whether it can be fixed or not, need to think a bit.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -1244,13 +1269,20 @@ class SpringApplicationContext(
...
+                springInjectedClassesStorage = beanQualifiedNames
+                    .map { fqn -> utContext.classLoader.loadClass(fqn) }
+                    .filterNot { it.isAbstract || it.isInterface || it.isLocalClass || it.isMemberClass && !it.isStatic }
+                    .mapTo(mutableSetOf()) { it.id }
+            }
+
+            return springInjectedClassesStorage
+        }
+
+    private var springInjectedClassesStorage = mutableSetOf<ClassId>()
```
Damtev, line 1285: Why is it a `var`? Also, this field requires some docs

utbot-framework/src/main/kotlin/org/utbot/engine/Traverser.kt
```diff
@@ -2322,35 +2322,26 @@ class Traverser(
...
-     */
-    @Suppress("KDocUnresolvedReference")
-    private fun SootField.speculativelyCannotProduceNullPointerException(): Boolean = isFinal || !isPublic
-
     private fun createArray(pName: String, type: ArrayType): ArrayValue {
         val addr = UtAddrExpression(mkBVConst(pName, UtIntSort))
         return createArray(addr, type, useConcreteType = false)
     }
 
+
```
Damtev, line 2344: Why?

Damtev (changes requested):

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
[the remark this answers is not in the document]
EgorkaKulikov: Actually no, I think. In case of inverted condition we are to use two return statements instead of one.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -1196,6 +1200,7 @@ enum class TypeReplacementMode {
  * @param mockFrameworkInstalled shows if we have installed framework dependencies
  * @param staticsMockingIsConfigured shows if we have installed static mocking tools
  */
+@Suppress("KDocUnresolvedReference")
```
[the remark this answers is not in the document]
EgorkaKulikov: I have just copied it from Dima and this suppressing is really required. Or we are to bad in Kotlin to create these docs without warnings...

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
EgorkaKulikov: I see the problem, but don't see a good solution. Going to think about it tomorrow a little more.

EgorkaKulikov: Actually, it seems a little bit impossible for me not to have at least one not filtered type here in real application context. However, I have applied a potential fix.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -1244,13 +1269,28 @@ class SpringApplicationContext(
...
+            if (springInjectedClassesStorage.isEmpty()) {
+                springInjectedClassesStorage += beanQualifiedNames
+                    .map { fqn -> utContext.classLoader.loadClass(fqn) }
+                    .filterNot { it.isAbstract || it.isInterface || it.isLocalClass || it.isMemberClass && !it.isStatic }
+                    .mapTo(mutableSetOf()) { it.id }
+
+                // This is done to be sure that this storage is not empty after the first class loading iteration.
+                // So, even if all loaded classes were filtered out, we will not try to load them again.
+                // Having `Object` in a list of injected classes is harmless from the point of abstract types replacements.
+                springInjectedClassesStorage += objectClassId
```
Damtev, line 1285: It seems a bit dangerous and very hard to debug. Consider using an additional flag to check whether `springInjectedClassesStorage` was initialized or not

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -1244,13 +1269,28 @@ class SpringApplicationContext(
...
+            }
+
+            return springInjectedClassesStorage
+        }
+
+    // This is a service field to model the lazy behavior of [springInjectedClasses].
+    // Do not call it outside the getter.
+    private val springInjectedClassesStorage = mutableSetOf<ClassId>()
```
Damtev, line 1293: Please, add some comments that we have an unresolved problem with classloaders when using `lazy`

Damtev (changes requested):

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
EgorkaKulikov: Replaced with flags

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -1244,16 +1269,35 @@ class SpringApplicationContext(
...
-    }
 
-    override val typeReplacementMode: TypeReplacementMode
-        get() = if (shouldUseImplementors) KnownImplementor else NoImplementors
+    private var areInjectedClassesInitialized : Boolean = false
+
+    // Classes representing concrete types that are actually used in Spring application
+    private val springInjectedClasses: Set<ClassId>
+        get() {
+            if (!areInjectedClassesInitialized && springInjectedClassesStorage.isEmpty()) {
```
Damtev, line 1278: I don't really think that we need the second condition anymore

Damtev (approved):

PR #2064, opened 2023-03-28
Rider plugin 223

Domonion:
1. Removed unnecessary SolutionComponents
2. ProjectPublisher now cancels correctly
3. ProjectPublisher correctly spawn publishing process
4. VSharp type loading fixes, this allows generating tests for Rider SDK classes
5. Assemblies now resolve from Rider-provided set of assemblies
6. Removed VSharp log directories after test generation
7. Correctly checks for overloads in method
8. Bumped minimum compatible version to .NET 7
9.  Bumped rd version to 2022.3.3
10. Rider plugin now shows consoles for publish and vsharp processes
11.  Rider version is targeting 223
12. Introduced UtBotRiderModel, which available when includeRiderInBuild=true. It copies rider-model.jar to generate necessary files


utbot-rider/src/dotnet/UtBot/UtBot/Utils/ProjectPublisher.cs
```diff
@@ -2,32 +2,28 @@
...
 namespace UtBot.Utils;
 
-[SolutionComponent]
-public class ProjectPublisher
+public static class ProjectPublisher
 {
-    private readonly LifetimeDefinition _lifetimeDef;
-
-    public ProjectPublisher(Lifetime lifetime)
+    public static bool PublishSync(ILogger logger, IBackgroundProgressIndicator indicator, IProject project, IProjectConfiguration config, VirtualFileSystemPath outputDir, UtBotRiderModel model)
```
mxprshn, line 20: Is it conceptually correct to bind publisher with indicator? Maybe it's cleaner to pass here a publishLifetimeDef object itself, and do `indicator.Cancel.Advise(Lifetime.Eternal, canceled => {if (canceled) publishLifetimeDef.Terminate();});` in the code which calls PublishSync

utbot-rider/src/dotnet/UtBot/UtBot/UnitTestBuilder.cs
```diff
@@ -202,16 +198,12 @@ private TypeDescriptor ToTypeDescriptor(IType typ, IMethod declMethod, ITypeElem
...
-        catch (Exception e)
-        {
-            var title = $"Cannot build project {project.Name}";
+
+        if (!ProjectPublisher.PublishSync(_logger, progressIndicator, project, config, outputDir, _riderModel)) {
+            var title = $"Cannot publish project {project.Name}";
             var openExceptionMessageCommand = new UserNotificationCommand(
                 "Show error info",
-                () => MessageBox.ShowError(e.Message, title));
+                () => MessageBox.ShowError(title, title));
```
mxprshn, line 206: This will just show a popup window with title "Cannot publish project X" and with the same content inside, which seems to be strange for user. If we don't pass any error info here now, it't better to remove the command

utbot-rider/src/dotnet/UtBot/UtBot/UnitTestBuilder.cs
```diff
@@ -250,16 +242,27 @@ void ChangeMethodName()
...
-            var proc = new ProcessWithRdServer(name, workingDir, port, runnerPath, project.Locks, _lifetime, _logger);
+            var proc = new ProcessWithRdServer(name, workingDir, port, runnerPath, project.Locks, _riderModel, _lifetime, _logger);
             var projectCsprojPath = project.ProjectFileLocation.FullPath;
+            List<MapEntry> allAssemblies;
+            using (_shellLocks.UsingReadLock())
+            {
+                allAssemblies = solution.GetAllAssemblies()
+                    .Where(it => it.Location.AssemblyPhysicalPath is not null)
+                    .Select(it => new MapEntry(it.FullAssemblyName, it.Location.AssemblyPhysicalPath!.FullPath))
+                    .DistinctBy(it => it.Key)
```
mxprshn, line 253: Why can we have different paths for the same assembly name? If this is a case, isn't it dangerous to pick a random one with distinct?

utbot-rider/src/dotnet/UtBot/UtBot/ProcessWithRdServer.cs
```diff
@@ -59,22 +60,25 @@ public ProcessWithRdServer(string name, string workingDir, int port, string exeP
                             blockingCollection.TryAdd(s);
                         }
                     });
-                    VSharpModel.Log.Advise(Lifetime, s => logger.Info($"V#: {s}"));
+                    VSharpModel.Log.Advise(Lifetime, s =>
+                    {
+                        logger.Info($"V#: {s}");
+                        riderModel.LogVSharp.Fire(s);
```
mxprshn, line 66: V# process logs are not intended to be read by the user if they don't explicitly want to read them. It's better to remove this at all or do this only in debug builds

mxprshn (changes requested):

utbot-rider/src/dotnet/UtBot/UtBot/ProcessWithRdServer.cs
mxprshn: Or this model is not used at all in release builds of plugin?

utbot-rider/src/dotnet/UtBot/UtBot/Utils/ProjectPublisher.cs
Domonion: Rider does not allow access to progress indicator lifetimeDefinition, so we cannot manually terminate it

mxprshn (approved):

Issue #2100, opened 2023-04-05
Update python coverage collection


tamarinvs19:
**Description**
__Problem__: now we can't get covered lines if the tested function was stopped by timeout.

__Possible solutions__:
* Use custom tracer and send covered lines to special socket channel.
* Use custom tracer, multithread in python executor for timeout tracking and save trace inside python and send it after timeout 

**Expected behavior**

There are opportunity to get coverage if the tested function was cancelled.

PR #2101, opened 2023-04-05
Python codegen improvement: timeouted tests and exceptions

### LINKED ISSUE #1939: Python: generate tests for timeouted functions
**Description**
Currently when function under test reaches `Hanging test timeout` limit then Python UTBot just skips such case. Instead we suggest to generate test with expectation that it fails within specified timeout or at least create _disabled_ test with comment about potential timeout.

**Expected behavior**

In case function timeouts during concrete execution Python UTBot may generate test which would handle that situation by adding some timeout constraint the same as UTBot Java does that with JUnit5 `@Timeout` annotation.

**Potential alternatives**

Also we could generate test which is disabled by default and add comment that test is disabled due timeout.

**Context**
Sample from Java with usage of JUnit5, similar expected to be implemented in Python.

User code:
```java
    int timeout(int a, int b) throws InterruptedException {
        Thread.sleep(2000);
        return a + b;
    }
```

Generated test, with `@Timeout` annotation which would prevent test from hanging more than 1sec.
```java
    @Test
    @DisplayName("timeout: return a + b -> TimeoutExceeded")
    @org.junit.jupiter.api.Timeout(value = 1000L, unit = TimeUnit.MILLISECONDS)
    public void testTimeout_ThreadSleep() throws InterruptedException {
        Sets sets = new Sets();
        
        /* This execution may take longer than the 1000 ms timeout
         and therefore fail due to exceeding the timeout. */
        assertTimeoutPreemptively(Duration.ofMillis(1000L), () -> sets.timeout(-255, -255));
    }
```

### LINKED ISSUE #2098: Use setting `Tests with exception` for Python
**Description**

Use option `Tests with exception` from UnitTestBot settings for Python same as for Java. 

**Expected behavior**

* with option `Fail`
Снимок экрана (8)

* with option `Pass`
Снимок экрана (9)

tamarinvs19:
## Description

* (Fixes #1939) Added tests skipping for timeouted tests. New problem: now we can't minimise these tests because we don't have coverage information. Current fix: maximal number of timeouted tests have been fixed. New issue about it #2100.
* (Fixes #2098) Assertions for executions are rendering according to global setting `Tests with exception`

#### Timeouted tests
1. Generate tests for the long function, e.g.
```python
import time
def f(x):
    time.sleep(2000)
```
2. Expected: test with skipping decorator
   - for `unittest`: `@unittest.skip(reason='Disabled due to the fact that the execution is longer then 1000 ms')`
   - for `pytest`: `@pytest.skip(reason='Disabled due to the fact that the execution is longer then 1000 ms')`

#### Exception asserts
See #2098 

## Self-check list


utbot-python/src/main/kotlin/org/utbot/python/framework/codegen/model/constructor/tree/PythonTestFrameworkManager.kt
```diff
@@ -48,6 +59,18 @@ internal class PytestManager(context: CgContext) : TestFrameworkManager(context)
     override fun addTestDescription(description: String) = Unit
 
     override fun disableTestMethod(reason: String) {
+        require(testFramework is Pytest) { "According to settings, JUnit4 was expected, but got: $testFramework" }
```
tochilinak, line 62: Probably should be Pytest in error message, not JUnit4

tamarinvs19: Fixed

Issue #2104, opened 2023-04-06
Incorrect cast to String in mock definition for a Controller


alisevych:
**Description**

Controller is cast to String in the mock definition for spring-petclinic project.

**To Reproduce**

1. Open as Gradle the [spring-petclinic project ](https://github.com/spring-projects/spring-petclinic) in IntelliJ Idea 2022.3.3
2. Install [the latest plugin built from main](https://github.com/UnitTestBot/UTBotJava/actions/runs/4618990661)
3. Generate tests for OwnerController
4. Check compilation errors in the generated test class

**Expected behavior**

There should not be casts of Controller classes to String.
Also OwnerController is added as a field with @InjectMocks annotation.

**Actual behavior**

There are two tests in ERROR SUITE where Controller is cast to String in the mock definition.
1 - in SYMBOLIC EXECUTION: ERROR SUITE for method processFindForm
2 - in OTHER: ERROR SUITE for method processFindForm

**Visual proofs (screenshots, logs, images)**

~~~java
public final class OwnerControllerTest {

	@InjectMocks
	public OwnerController ownerController;

	/**
	 * @utbot.classUnderTest {@link OwnerController}
	 * @utbot.methodUnderTest {@link OwnerController#processFindForm(int, Owner, BindingResult, Model)}
	 * @utbot.executesCondition {@code (owner.getLastName() == null): False}
	 * @utbot.throwsException {@link NullPointerException} in: ownersResults.isEmpty()
	 */
	@Test
	@DisplayName("processFindForm: ownersResults.isEmpty() -> ThrowNullPointerException")
	public void testProcessFindForm_OwnerGetLastNameNotEqualsNull() {
		org.mockito.MockedStatic mockedStatic = null;
		try {
			mockedStatic = mockStatic(PageRequest.class);
			(mockedStatic.when(() -> PageRequest.of(anyInt(), anyInt()))).thenReturn(((PageRequest) null));
            /* WARNING!!! Automatically used "Mockito static mocking" framework for mocking statics
            because execution encountered flaky methods
            To change this behaviour edit [Settings -> UtBot -> Force static mocking] */
			(when(ownerRepositoryMock.findByLastName(any(), any()))).thenReturn(((Page) null));
			Owner ownerMock = mock(Owner.class);
			(when(ownerMock.getLastName())).thenReturn(((String) ownerController), ((String) null));

[...]

PR #2191, opened 2023-04-28
Fix unmatched variables in Spring test methods

### LINKED ISSUE #2107: Incorrect mock for a method returning List of entities
**Description**

Incorrect mock is generated for the method of a Repository that returns List<PetType>
PetType is the entity.

**To Reproduce**

1. Open as Gradle the [spring-petclinic project ](https://github.com/spring-projects/spring-petclinic) in IntelliJ Idea 2022.3.3
2. Install [the latest plugin built from main](https://github.com/UnitTestBot/UTBotJava/actions/runs/4618990661)
3. Generate tests for PetTypeFormatter$parse
4. Check compilation errors in the generated test class

**Expected behavior**

Some valid List of PetType should be instantiated and used in the mock.
Some valid String should be passed into the method call.

**Actual behavior**

There is `arrayList` word returned from mock. It is not defined anywhere in the test. It is used in several tests.
There is `string` word as the input parameter of  the method `parse` call. It is not defined anywhere in the test either.

image

The following test is generated:

~~~java
	///region Test suites for executable org.springframework.samples.petclinic.owner.PetTypeFormatter.parse

	///region SYMBOLIC EXECUTION: SUCCESSFUL EXECUTIONS for method parse(java.lang.String, java.util.Locale)

	/**
	 * @utbot.classUnderTest {@link PetTypeFormatter}
	 * @utbot.methodUnderTest {@link PetTypeFormatter#parse(String, Locale)}
	 * @utbot.invokes {@link OwnerRepository#findPetTypes()}
	 * @utbot.invokes {@link java.util.Collection#iterator()}
	 * @utbot.iterates iterate the loop {@code for(PetType type: findPetTypes)} once
	 */
	@Test
	@DisplayName("parse: OwnerRepositoryFindPetTypes -> return type")
	public void testParse_TypeGetNameEquals() throws ParseException, ClassNotFoundException, IllegalAccessException, NoSuchFieldException, InvocationTargetException, NoSuchMethodException {
		(when(ownerRepositoryMock.findPetTypes())).thenReturn(arrayList);

		PetType actual = petTypeFormatter.parse(string, ((Locale) null));

		String petTypeFormatterName = ((String) getFieldValue(petTypeFormatter, "org.springframework.samples.petclinic.model.NamedEntity", "name"));
		String actualName = ((String) getFieldValue(actual, "org.springframework.samples.petclinic.model.NamedEntity", "name"));
		assertEquals(petTypeFormatterName, actualName);

		Integer actualId = ((Integer) getFieldValue(actual, "org.springframework.samples.petclinic.model.BaseEntity", "id"));
		assertNull(actualId);

	}
	///endregion
~~~

**Environment**

Windows 10 Pro
IntelliJ IDEA 2022.3.3 Ultimate
Project is opened as Gradle

sofurihafe:
## Description

For some reason UtAssembleModel *equal method* did not have comparison by ClassId, while one in UtCompositeModel has it.
Some observations proved that the former should also have it due to problems with Maps where keys are UtModels...

image

The necessity of ClassId comparisons in UtAssembleModels was under consideration some time ago, but it was ignored for an unknown reason. From a cursory glance, it does seem missing. If by any chance it's true, it may require some refactoring.

Fixes #2107, #2104

See the comment below⬇️.

### Manual tests

Tested several times on OwnerController and VetController classes.

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -62,7 +62,7 @@ import org.utbot.framework.plugin.api.TypeReplacementMode.*
 import org.utbot.framework.plugin.api.util.allDeclaredFieldIds
 import org.utbot.framework.plugin.api.util.fieldId
 import org.utbot.framework.plugin.api.util.isSubtypeOf
-import org.utbot.framework.plugin.api.util.objectClassId
+import org.utbot.framework.plugin.api.util.isNotSubtypeOf
 import org.utbot.framework.plugin.api.util.utContext
 import org.utbot.framework.process.OpenModulesContainer
 import soot.SootField
@@ -546,7 +546,9 @@ data class UtAssembleModel private constructor(
         other as UtAssembleModel
 
         if (id != other.id) return false
-        if (classId != other.classId) return false
+        if (classId != other.classId) {
+            if (classId isNotSubtypeOf other.classId && other.classId isNotSubtypeOf classId) return false
+        }
 
         return true
     }
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -544,7 +544,11 @@ data class UtAssembleModel private constructor(
         if (javaClass != other?.javaClass) return false
 
         other as UtAssembleModel
-        return id == other.id
+
+        if (id != other.id) return false
+        if (classId != other.classId) return false
+
+        return true
     }
 
     override fun hashCode(): Int = id ?: 0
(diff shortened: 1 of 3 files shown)


sofurihafe:
The first attempt was partially successful.
"SetIteratorsTest" failed (see this [run](https://github.com/UnitTestBot/UTBotJava/actions/runs/4832486210)).

Generated tests before and after look like this. There were some extra initializations instead of using already created linkedHashSet.
image

So, it seems like two different UtAssembleModels have the same Id (to be honest, I don't get it 😞) in one execution and in one testSet, that's why it was not being recreated before (see screenshot). And it was proved in a clean simple maven project without Spring.
image

It works fine for non-Spring tests. But we still have the problem with clashing ids for Spring with no way of fixing it without actually making some changes in UtAssembleModel comparison logic (that's how I see it by the moment 😿).

💡 But what if we **also** check for ClassIds being a subtype of each other? And, when at least one of them is a subtype of another, we consider them as "equal", but it's not truly true, though. This tweak solves our problem with VetController and ArrayList models being equal and does not influence the way UtAssembleModel are used and replaced when they have the same id.

EgorkaKulikov (changes requested):
Discussed in internal chats

PR #2213, opened 2023-05-19
Better Bean class name extraction

### LINKED ISSUE #2161: Try to avoid getBean call in UtBotBeanFactoryPostProcessor
**Description**

`UtBotBeanFactoryPostProcessor` should not instantiate beans due to potential side effects of this process.
However, currently we do it sometimes to obtain relevant bean definitions.

We should investigate how to avoid calling `getBean`.
One of possible solutions is to remember the name of the method, instantiating a bean, and find actual type in utbot from PsiMethod.
However, this solution is going to work from plugin only, it will have problems in utbot-cli.

**Expected behavior**

Let's firstly suggest an approach, may be it will be implemented later.
May be it should be discussed with Ilya.

sofurihafe:
## Description

This PR removes usage of *getBean()* method that may lead to unwanted side effects.
Now, we extract bean class name by analyzing corresponding PsiMethod.

Fixes #2161

### Automated tests

No automated tests.

### Manual tests


utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/util/PsiClassHelper.kt
```diff
@@ -21,6 +24,11 @@ val PsiMember.isAbstract: Boolean
 val PsiMember.isStatic: Boolean
     get() = modifierList?.hasModifierProperty(PsiModifier.STATIC)?: false
 
+fun findClass(name: String, project: Project): PsiClass? =
+    JavaPsiFacade
+        .getInstance(project)
+        .findClass(name, GlobalSearchScope.projectScope(project))
+
```
IlyaMuravjov, line 31: Don't put such a generic name in a global scope, put it inside object

utbot-intellij/src/main/kotlin/org/utbot/intellij/plugin/generator/UtTestsDialogProcessor.kt
```diff
@@ -445,6 +448,60 @@ object UtTestsDialogProcessor {
...
+                        // Here we try to take a single element
+                        // because we expect no or one method matching previous conditions only.
+                        // If there were two or more similar methods in one class, it would be a weird case.
+                        .singleOrNull()
+                        ?: return@runReadAction null
+                            .also {
+                                logger.warn(
+                                    "Two similar methods named ${bean.beanName} " +
+                                            "were found in ${additionalData.configClassFqn} configuration class."
+                                )
```
IlyaMuravjov, line 487: Can also be triggered when 0 or 3 methods are found

IlyaMuravjov (approved):

Issue #2217, opened 2023-05-22
Unnecessary variables in Spring integration tests


EgorkaKulikov:
**Description**

Consider generating Spring integration tests for project `spring-boot-testing`, method `OrderService.save`

Generated tests look like as follows:

```java
@Test(description = "createOrder: order = Order()")
    public void testCreateOrder() {
       <...>

        Order actual = orderService.createOrder(order1);

        Long long1 = 1L;
        String string = "abc";
        Double double1 = Double.POSITIVE_INFINITY;
        Order expected = new Order(long1, string, double1, 0);
        // com.rest.order.models.Order has overridden equals method
        assertEquals(expected, actual);
    }
```

It seems that there are many unnecessary variables. The code may be much more simple:

```java
Order expected = new Order(1L, "abc", Double.POSITIVE_INFINITY, 0);
```

EgorkaKulikov:
It seems that this behavior can not be easily corrected. 
More details can be found in the mentioned PR

PR #2234, opened 2023-05-24
Create literals for assemble models representing primitive wrappers

EgorkaKulikov:
## Description

Fixes # ([2217](https://github.com/UnitTestBot/UTBotJava/issues/2217))
Also corrects `classId` in methods of some assemble models obtained from concrete execution.

### Automated tests

`utbot-samples` pipeline as a set of regression checks

### Manual tests

Scenario from the issue

## Self-check list

- [] **Manual tests** have been provided optionally.


EgorkaKulikov:
Unfortunately, one two tests for current method fails after these changes:

```java
public class ObjectWithFinalStatic {
    public final static Integer keyValue = 420;

    public int parameterEqualsFinalStatic(Integer key, int value) {
        if (key == keyValue) {
            return value;
        }
        return -420;
    }
}
```

The reason is that we need the value 420 two times.
When the variable is created, it is ok. When literals are used, standard boxing leads to different wrappers for one constant that are not equal.

PR #2256, opened 2023-06-05
Improve fuzzing for domain specific API

Markoutte:
## Description

Several enhancements and bug fixes are done:

1. Fix `ClassCastException` when collecting constants from Soot.
2. Enable fuzzing for static methods with no parameters.
3. Add an ability to generate very specific values in some specific cases.

## Self-check list


utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
```diff
@@ -358,18 +379,33 @@ private fun <TYPE, RESULT, DESCRIPTION : Description<TYPE>, FEEDBACK : Feedback<
...
+        error("Forked description and origin description must be same type, but fork has ${fork::class} type")
+    }
+
+    val seeds = when {
+        fork !== description -> {
+            fuzzing.generate(fork, type).toList()
+        }
+        else -> state.cache.computeIfAbsent(type) {
+            fuzzing.generate(fork, it).toList()
+        }
```
IlyaMuravjov, line 395: It seems unintended that we use `state.cache` here when the `description` is already a fork and is not the original description that was used to populate `state.cache`, but we just happened to not fork `description` one more time during this execution of the `produce` function, so `fork === description`

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/JavaLanguage.kt
```diff
@@ -19,18 +17,38 @@ private val logger = KotlinLogging.logger {}
...
     description.parameters.mapIndexed { index, classId ->
         description.fuzzerType(index) ?: FuzzedType(classId)
     }
-) {
+), ScopeSensitive<FuzzedType, FuzzedDescription> {
     val constants: Sequence<FuzzedConcreteValue>
         get() = description.concreteValues.asSequence()
+
+    override fun fork(type: FuzzedType, scope: ScopeParams): FuzzedDescription? {
+        if (checkParamIsThis() == true) {
```
IlyaMuravjov, line 35: Only forking when `checkParamIsThis(scope) == true` leads to only being able to transition to `scope` such that `checkParamIsThis(scope) == true`, but not backwards. Since `description` is propagated from `thisInstanceParam` to parameters of constructor used to instantiate a value for `thisInstanceParam` and parameters of methods used to mutate it, this causes all these parameters to also be treated as if they were `thisInstance` themselves, that in turn leads to not covering an exception branch of `checkInnerIsNotNull` method in the following class:

```java
public class Outer {
    public static class Inner { }

    private final Inner inner;

    public Outer(Inner inner) {
        this.inner = inner;
    }

    public void checkInnerIsNotNull() {
        if (inner == null)
            throw new RuntimeException("Inner is null");
    }
}
```

But if we fork description every time, how is adding some dynamic flags/properties to `FuzzedType` is any worse? Flag examples: `NullValueProvider.IgnoredTypeFlag`, `ObjectValueProvider.IgnoredTypeFlag`.

Similar (even more complex) extension of Java type system with flags is used by the [Checker Framework](https://checkerframework.org/manual/#writing-annotations) just fine, so adding flags should not break Java fuzzing type system.

IlyaMuravjov (changes requested):

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/type/factories/FuzzedTypeFactory.kt
```diff
@@ -1,8 +0,0 @@
-package org.utbot.fuzzing.type.factories
-
-import org.utbot.fuzzer.FuzzedType
-import java.lang.reflect.Type
-
-interface FuzzedTypeFactory {
-    fun createFuzzedType(type: Type, isThisInstance: Boolean): FuzzedType
-}
```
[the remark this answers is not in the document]
Markoutte: As it was discussed I'll try to introduce another way for extending types in java fuzzer

IlyaMuravjov (changes requested):
`org.utbot.fuzzing.ValueProvider$withFallback$1` doesn't override `enrich()` method and hence `SpringBeanValueProvider.enrich()` is never called

Markoutte:
> `org.utbot.fuzzing.ValueProvider$withFallback$1` doesn't override `enrich()` method and hence `SpringBeanValueProvider.enrich()` is never called

@IlyaMuravjov, could you please, review and mark it resolved if the problem doesn't reproduce anymore and Spring's implementation works as it does before.

Markoutte:
> The final design of scope sensitive fuzzing looks quite cumbersome. How about we just make fuzzer work with `ScopeSensetiveValueProvider`s instead of simple `ValueProvider`s, while also keeping simple `ValueProvider`s for backward compatibility?

Let's create a new issue to further discussion of this particular problem. Currently, I think that:
1. A user shouldn't know about scopes until the scope is really needed (by design, scope is not needed)
2. ValueProvider have to be stateless. In this example:

```
    override fun generate(description: D, type: T): ((scope: FuzzingScope) -> List<Seed<T, R>>) {
        val cache = mutableMapOf<K, List<Seed<T, R>>>()
        return { scope -> 
            val key = chooseKey(description, type, scope)
            cache.getOrPut(key) { generate(description, type, key).toList() }
        }
    }
```
cache is the state, that can be held indefinitely.

IlyaMuravjov:
> > The final design of scope sensitive fuzzing looks quite cumbersome. How about we just make fuzzer work with `ScopeSensetiveValueProvider`s instead of simple `ValueProvider`s, while also keeping simple `ValueProvider`s for backward compatibility?
> 
> Let's create a new issue to further discussion of this particular problem. Currently, I think that:
> 
> 1. A user shouldn't know about scopes until the scope is really needed (by design, scope is not needed)
> 2. ValueProvider has to be stateless. In this example:
> 
> ```
>     override fun generate(description: D, type: T): ((scope: FuzzingScope) -> List<Seed<T, R>>) {
>         val cache = mutableMapOf<K, List<Seed<T, R>>>()
>         return { scope -> 
>             val key = chooseKey(description, type, scope)
>             cache.getOrPut(key) { generate(description, type, key).toList() }
>         }
>     }
> ```
> 
> cache is the state, that can be held indefinitely.

1. A user that doesn't need scopes can implement `ValueProvider` instead of `ScopeSensetiveValueProvider`
2. In that example, it's important to clarify that the `ValueProvider` doesn't directly hold a reference to the `cache`. Instead, the `cache` is captured by the function that is returned by the `generate` method. This means that the `cache` will only be retained as long as the returned function is stored by the caller of the `ValueProvider`.

IlyaMuravjov (approved):
Now Spring related fuzzing works as it did before

Vassiliy-Kudryashov (approved):

PR #2257, opened 2023-06-05
Support direct field accesses in call models

EgorkaKulikov:
## Description

Enlarges a way to construct assemble models.

Previously only executables could be used to modify intilial state. Actually, direct field acessors may be useful too (e.g. for Spring projects).

## How to test

Difficult to test.
Needs to take in mind all cycle of assemble models processing.
May be we should rely on pipeline...

## Self-check list

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -495,14 +495,14 @@ data class UtArrayModel(
  *
  * The default constructor is made private to enforce using a safe constructor.
  *
- * @param instantiationCall is an [UtExecutableCallModel] to instantiate represented object. It **must** not return `null`.
+ * @param instantiationCall is an [UtStatementCallModel] to instantiate represented object. It **must** not return `null`.
  * @param modificationsChain is a chain of [UtStatementModel] to construct object state.
  */
 data class UtAssembleModel private constructor(
     override val id: Int?,
     override val classId: ClassId,
     override val modelName: String,
-    val instantiationCall: UtExecutableCallModel,
+    val instantiationCall: UtStatementCallModel,
     val modificationsChain: List<UtStatementModel>,
     val origin: UtCompositeModel?
 ) : UtReferenceModel(id, classId, modelName) {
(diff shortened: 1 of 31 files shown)


utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -691,23 +691,23 @@ sealed class UtStatementModel(
...
+ * Step of assemble instruction that calls executable or accesses a field.
  *
+ * Contains statement to call, call parameters and an instance model before call.
  * @param [instance] **must be** `null` for static methods and constructors
  */
-data class UtExecutableCallModel(
+data class UtStatementCallModel(
     override val instance: UtReferenceModel?,
-    val executable: ExecutableId,
+    val statement: StatementId,
```
Damtev, line 701: What if we pass a few parameters with a `DirectFieldAccessId`? As I see from the usages, we expect only one argument. Shouldn't we check it during initialization?

Moreover, perhaps it's better to extract an abstract class that stores only the `instance` and the abstract `statement` that will have 2 inheritors - for executable calls with list of parameters and for setting field with only one parameter as a value.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/util/constructor/ValueConstructor.kt
```diff
@@ -435,11 +437,19 @@ class ValueConstructor {
             method.invokeCatching(obj = instance, args = args).getOrThrow()
         }
 
+    private fun DirectFieldAccessId.call(args: List<Any?>, instance: Any?): Any? {
```
Damtev, line 440: This name looks a bit confusing - perhaps, use `set` instead?

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/execution/constructors/MockValueConstructor.kt
```diff
@@ -435,6 +437,13 @@ class MockValueConstructor(
             invokeCatching(obj = instance, args = args).getOrThrow()
         }
 
+    private fun DirectFieldAccessId.call(args: List<Any?>, instance: Any?): Any? {
```
Damtev, line 440: Same about the name

Damtev (changes requested):

Damtev (changes requested):
I see some problems that may be solved by extending a hierarchy - with 2 different classes for executable calls and filed setters with one common abstract ancestor

PR #2261, opened 2023-06-06
Support direct field accesses in call models

EgorkaKulikov:
## Description

Enlarges a way to construct assemble models.

Previously only executables could be used to modify intilial state. Actually, direct field acessors may be useful too (e.g. for Spring projects).

## How to test

Difficult to test.
Needs to take in mind all cycle of assemble models processing.
May be we should rely on pipeline...

## Self-check list

### DIFF
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -740,7 +740,7 @@ data class UtExecutableCallModel(
  *
  * Contains parameter value to set and an instance model before call.
  */
-data class UtDirectFieldAccessModel(
+data class UtDirectGetFieldModel(
     override val instance: UtReferenceModel,
     val fieldAccess: DirectFieldAccessId,
 ) : UtStatementCallModel(instance, fieldAccess, emptyList()) {
--- a/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
+++ b/utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
@@ -502,7 +502,7 @@ data class UtAssembleModel private constructor(
     override val id: Int?,
     override val classId: ClassId,
     override val modelName: String,
-    val instantiationCall: UtExecutableCallModel,
+    val instantiationCall: UtStatementCallModel,
     val modificationsChain: List<UtStatementModel>,
     val origin: UtCompositeModel?
 ) : UtReferenceModel(id, classId, modelName) {
@@ -527,7 +527,7 @@ data class UtAssembleModel private constructor(
         id: Int?,
         classId: ClassId,
         modelName: String,
-        instantiationCall: UtExecutableCallModel,
+        instantiationCall: UtStatementCallModel,
         origin: UtCompositeModel? = null,
         modificationsChainProvider: UtAssembleModel.() -> List<UtStatementModel> = { emptyList() }
     ) : this(id, classId, modelName, instantiationCall, mutableListOf(), origin) {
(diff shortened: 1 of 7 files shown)


Damtev (approved):
LGTM

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
```diff
@@ -119,7 +120,7 @@ class SpringTestClassModelBuilder(val context: CgContext): TestClassModelBuilder
                 currentModel.modificationsChain.forEach { stmt ->
                     stmt.instance?.let { collectRecursively(it, allModels) }
                     when (stmt) {
-                        is UtExecutableCallModel -> stmt.params.forEach { collectRecursively(it, allModels) }
+                        is UtStatementCallModel -> stmt.params.forEach { collectRecursively(it, allModels) }
```
IlyaMuravjov, line 123: We probably want to call `collectRecursively` on `params.instance` as well

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgVariableConstructor.kt
```diff
@@ -231,21 +234,22 @@ open class CgVariableConstructor(val context: CgContext) :
...
         baseName: String?
     ) {
-        val executable = executableCall.executable
-        val params = executableCall.params
+        val executable = statementCall.statement
+        val params = statementCall.params
 
         val type = when (executable) {
             is MethodId -> executable.returnType
+            is DirectFieldAccessId -> executable.fieldId.declaringClass
```
IlyaMuravjov, line 245:
```suggestion
            is DirectFieldAccessId -> executable.fieldId.type
```

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
```diff
@@ -242,7 +243,7 @@ private fun UtModel.calculateSize(used: MutableSet<UtModel> = mutableSetOf()): I
 private fun UtStatementModel.calculateSize(used: MutableSet<UtModel> = mutableSetOf()): Int =
     when (this) {
         is UtDirectSetFieldModel -> 1 + fieldModel.calculateSize(used)
-        is UtExecutableCallModel -> 1 + params.sumOf { it.calculateSize(used) }
+        is UtStatementCallModel -> 1 + params.sumOf { it.calculateSize(used) }
```
IlyaMuravjov, line 246: I think, here we should also add `instance?.calculateSize(used) ?: 0` in all cases. It won't get accounted multiple times because `calculateSize` returns `0` for models that has already been `used`.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
```diff
@@ -726,6 +721,28 @@ data class UtExecutableCallModel(
...
+
+/**
+ * Step of assemble instruction that directly accesses a field.
+ *
+ * Contains parameter value to set and an instance model before call.
+ */
+data class UtDirectFieldAccessModel(
+    override val instance: UtReferenceModel,
+    val fieldAccess: DirectFieldAccessId,
+) : UtStatementCallModel(instance, fieldAccess, emptyList())
```
IlyaMuravjov, line 744: `UtReflectiveSetFieldModel` seems a better name, not to be confused with already existing `UtDirectSetFieldModel` that is used, for example, like this:
```kotlin
is UtDirectSetFieldModel -> {
    val instance = declareOrGet(statementModel.instance)
    // fields here are supposed to be accessible, so we assign them directly without any checks
    instance[statementModel.fieldId] `=` declareOrGet(statementModel.fieldModel)
}
```

IlyaMuravjov (changes requested):

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/domain/models/builders/SpringTestClassModelBuilder.kt
EgorkaKulikov: Do you mean `stmt.instance`? 
But this has already been done in line 121.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
EgorkaKulikov: Your names seems strange. This is getter, not setter model.
To avoid confusion with `UtDirectSetFieldModel, I named it `UtDirectGetFieldModel`

Actually, it is not very reflective as uses `DirectFieldAccessId`, even though it allows the rendering in reflective way.

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
EgorkaKulikov: We should better keep this place "as is" in current PR. If we really want to change smth in minimization, let's do it in Andrey's PR, for example.

utbot-framework-api/src/main/kotlin/org/utbot/framework/plugin/api/Api.kt
IlyaMuravjov: Why is it using `DirectFieldAccessId` instead of just `FieldId` if it allows rendering in a reflective way?

IlyaMuravjov (approved):
Checkout new comments before merging

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
EgorkaKulikov: Asked Andrey do not forget about it.

PR #2274, opened 2023-06-09
Automatically detect the names of tables and repositories

EgorkaKulikov:
## Description

Avoid hardcoding the names of tables and repositories: get this data via reflection and bean definitions.

### Automated tests

Please specify the _automated tests_ for your code changes: you should either mention the existing tests or add the new ones.

For example:

> `utbot-fuzzing/src/test/kotlin/org/utbot/fuzzing/FuzzerSmokeTest.kt`

### Manual tests

If it is impossible to provide the automated tests, please reason why. Usually, it is relevant only for UI- or documentation-related PRs.
If this is your case, share the detailed _manual scenarios_ that help to verify your changes.

## Self-check list

- [[x] I've set the proper **labels** for my PR (at least, for category and component).


utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/util/ReflectionUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+        if (alreadyAnalyzedTypes.contains(clazz)) {
+            return emptySet()
+        }
+
+        collectedFieldTypes += collectFieldTypes(clazz)
+
+        val fieldTypes = clazz.fields.map { it.type }
+        for (fieldType in fieldTypes) {
+            collectedFieldTypes += collectAccessibleFieldTypes(fieldType)
+        }
```
IlyaMuravjov, line 25: Consider the following example, where we want to collect dependent beans for `OrderController`.
```kotlin
@Controller
class OrderController {
    @Autowired
    private val orderService: OrderService
}

@Service
class SimpleOrderService : OrderService {
    @Autowired
    private val orderRepository: OrderRepository
}

interface OrderService

@Repository
interface OrderRepository
```

The suggested implementation will collect field types for `OrderController` and then for `OrderService` and will miss `OrderRepository` that is referenced not by `OrderService` itself, but rather by `SimpleOrderService`.

To avoid that I suggest instead of using field types of the class itself use actual runtime types of field values stored in the bean object. Such approach will also simplify getting bean names from bean types. Once reachable runtime field types are collected, reachable beans can be determined by taking beans that have a type that is equal to one of reachable runtime field types (including beans that have a type that is a subtype of one of collected field types is no longer necessary).

utbot-spring-commons/src/main/kotlin/org/utbot/spring/context/SpringContextWrapper.kt
```diff
@@ -1,23 +1,30 @@
...
-        beanDefinitionRegistry.removeBeanDefinition(beanName)
+    override fun getBean(beanName: String): Any = context.getBean(beanName)
+    override fun getRelatedBeanTypes(clazz: Class<*>): Set<SimpleBeanDefinition> {
+        val cutName = clazz.canonicalName
+        return context.beanDefinitionNames
+            .map { name -> name to context.getBean(name) }
+            .filter {
+                val bean = it.second
+                val typeName = bean.javaClass.canonicalName
+                typeName != null && cutName.commonPrefixWith(typeName).isNotEmpty()
```
IlyaMuravjov, line 17: [`Class#isAssignableFrom(Class)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isAssignableFrom-java.lang.Class-) should be used here, but it's no longer a relevant problem, given the previous comment is addressed.

IlyaMuravjov (changes requested):

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/util/ReflectionUtils.kt
IlyaMuravjov: It's worth noting that this approach may be able to find `dataSource: DataSource` bean that may be time consuming to reset, so we may want to filter it out since we are planning to "reset" `dataSource` by doing everything in a transaction and then rollbacking it.

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/execution/SpringUtExecutionInstrumentation.kt
```diff
@@ -79,22 +81,37 @@ class SpringUtExecutionInstrumentation(
...
+            .filter {
+                val implementedInterfaces = it.beanType.genericInterfaces.map { i -> i.javaClass.canonicalName }
+                "org.springframework.data.repository.Repository" in  implementedInterfaces
+            }
+            .map { (it.beanType.genericInterfaces[0] as ParameterizedType).actualTypeArguments[0] }
+            .map { it.typeName }
+
+        //TODO: table name may be different from entity name (set with annotation on entity)
+        //TODO: plural form is not always obtained with "+s", it can be even Person->People
+        val tableNames = entityNames.map { it + "s" }
```
IlyaMuravjov, line 108: Don't think too much about it now, truncating tables is planned to be replaced with doing everything in a transaction and then rollbacking it.

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/util/AccessibleTypesAnalyzer.kt
```diff
@@ -28,14 +31,14 @@ class ReflectionUtils {
     }
 
     private fun collectFieldTypes(clazz: Class<*>): Set<Class<*>> {
-        val types = mutableSetOf<Class<*>>()
+        val subTypes = Reflections().getSubTypesOf(clazz)
```
IlyaMuravjov, line 34: I suspect that for that to work you may need to configure `Reflections` to use `utContext.classLoader`. Also, this operation may turn out to be very expensive, taking a few seconds to scan the entire classpath every time it's performed.

utbot-spring-commons-api/src/main/kotlin/org/utbot/spring/api/context/ContextWrapper.kt
```diff
@@ -3,11 +3,22 @@ package org.utbot.spring.api.context
...
+data class SimpleBeanDefinition(
+    val beanName: String,
+    val bean: Any,
+)
+data class RepositoryDescription(
+    val beanName: String,
+    val repositoryName: String,
+    val entityName: String,
+    val tableName: String,
+)
```
IlyaMuravjov, line 24: This class seems to be similar to `SpringRepositoryId` from `utbot-framework-api`, can we use `SpringRepositoryId` instead of it. Although, on second thought adding dependency on `utbot-framework-api` can noticeably increase `spring-analyzer-shadow.jar` size, since `utbot-framework-api` has, for example, a dependency on `soot`.

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/execution/SpringUtExecutionInstrumentation.kt
```diff
@@ -43,39 +59,46 @@ class SpringUtExecutionInstrumentation(
...
+        for (repositoryDefinition in repositoryDefinitions) {
+            val truncateTableCommand = "TRUNCATE TABLE ${repositoryDefinition.tableName}"
+            jdbcTemplate::class.java
+                .getMethod("execute", truncateTableCommand::class.java)
+                .invoke(jdbcTemplate, truncateTableCommand)
+
+            val restartIdCommand = "ALTER TABLE ${repositoryDefinition.tableName} ALTER COLUMN id RESTART WITH 1"
+            jdbcTemplate::class.java
+                .getMethod("execute", restartIdCommand::class.java)
+                .invoke(jdbcTemplate, restartIdCommand)
```
IlyaMuravjov, line 78: Here the name of the column ("id") and the way it's reset are still hardcoded. By saying that the way of resetting is hardcoded I mean that, for example, if `id` is defined as shown below we need to reset sequence instead of the column:
```java
@GeneratedValue(strategy = GenerationType.SEQUENCE)
private Integer id;
```
To avoid dealing with that, it's planned to do everything in a transaction and rollback the transaction after each run, to avoid dealing with all the possible ways to reset something. For now, I suggest to just omit resetting columns.

utbot-rd/src/main/rdgen/org/utbot/rd/models/InstrumentedProcessModel.kt
```diff
@@ -49,6 +49,14 @@ object InstrumentedProcessModel : Ext(InstrumentedProcessRoot) {
         field("beanModel", array(PredefinedType.byte))
     }
 
+    val GetSpringRepositoriesParams = structdef {
+        field("classId", array(PredefinedType.byte))
+    }
+
+    val GetSpringRepositoriesResult = structdef {
+        field("repositoryDescriptions", array(PredefinedType.byte))
```
IlyaMuravjov, line 57:
```suggestion
        field("springRepositoryIds", array(PredefinedType.byte))
```

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/execution/SpringUtExecutionInstrumentation.kt
```diff
@@ -43,39 +59,46 @@ class SpringUtExecutionInstrumentation(
...
+        }
 
         return delegateInstrumentation.invoke(clazz, methodSignature, arguments, parameters)
     }
 
+    private fun getRelevantBeanNames(clazz: Class<*>): Set<String> =
```
IlyaMuravjov, line 84: Consider using `relatedBeansCache` inside of this method, right now `getRepositoryDescriptions` calls this method ignoring the cache.

utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/ConcreteExecutor.kt
```diff
@@ -285,6 +282,16 @@ fun ConcreteExecutor<*,*>.warmup() = runBlocking {
     }
 }
 
+fun ConcreteExecutor<*,*>.getSpringRepositories(classId: ClassId): Set<SpringRepositoryId> = runBlocking {
```
IlyaMuravjov, line 285:
```suggestion
fun ConcreteExecutor<*,*>.getRelevantSpringRepositories(classId: ClassId): Set<SpringRepositoryId> = runBlocking {
```

IlyaMuravjov (approved):

PR #2298, opened 2023-06-20
Support fuzzing for regular expressions in Python

tamarinvs19:
## Description

Supported fuzzing generation strings by regular expressions from python testing function.

### Manual tests

See sample `utbot-python/samples/easy_samples/regex.py`:
```python
def check_regex(string: str) -> bool:
    pattern = r"'(''|\\\\|\\'|[^'])*'"
    if re.match(pattern, string):
        return True
    return False
```

1. Try to generate tests
2. __Expected__: test with string for regex

## Self-check list

### DIFF
--- a/utbot-python/src/main/kotlin/org/utbot/python/fuzzing/provider/StrValueProvider.kt
+++ b/utbot-python/src/main/kotlin/org/utbot/python/fuzzing/provider/StrValueProvider.kt
@@ -3,6 +3,7 @@ package org.utbot.python.fuzzing.provider
 import org.utbot.fuzzing.Seed
 import org.utbot.fuzzing.ValueProvider
 import org.utbot.fuzzing.seeds.KnownValue
+import org.utbot.fuzzing.seeds.RegexValue
 import org.utbot.fuzzing.seeds.StringValue
 import org.utbot.python.framework.api.python.PythonTree
 import org.utbot.python.framework.api.python.util.pythonStrClassId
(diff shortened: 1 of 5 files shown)


utbot-python/src/main/kotlin/org/utbot/python/fuzzing/provider/utils/StringUtils.kt
```diff
@@ -12,9 +15,25 @@ fun String.transformQuotationMarks(): String {
...
         return this.drop(1).dropLast(1)
     }
 
     return this
-}
\ No newline at end of file
+}
+
+fun String.transformRawString(): String {
+    return if (Pattern.matches("r\".*\"", this) || Pattern.matches("r'.*'", this)) {
```
Markoutte, line 26: I'd recommend to user simpler checks like `startsWith`, `endWith` due to the complexity of regex checks.

utbot-python/src/main/kotlin/org/utbot/python/fuzzing/provider/utils/StringUtils.kt
```diff
@@ -12,9 +15,25 @@ fun String.transformQuotationMarks(): String {
...
+        this.drop(2).dropLast(1)
```
Markoutte, line 27: You can use `String.substring(int, int)` method

utbot-python/src/main/kotlin/org/utbot/python/fuzzing/provider/StrValueProvider.kt
```diff
@@ -3,38 +3,53 @@ package org.utbot.python.fuzzing.provider
...
 
     override fun generate(description: PythonMethodDescription, type: Type) = sequence {
         val strConstants = getStrConstants(description.concreteValues) + listOf(
-            StringValue("pythön"),
-            StringValue("abcdefghijklmnopqrst"),
-            StringValue("foo"),
-            StringValue("€"),
+            "pythön",
+            "foo",
+            "\t\n\r",
```
Markoutte, line 42: I'm not sure, but I did see the error with bad input because of this string and its escaping. I'll try to reproduce this.

utbot-python/src/main/kotlin/org/utbot/python/fuzzing/provider/StrValueProvider.kt
```diff
@@ -3,38 +3,53 @@ package org.utbot.python.fuzzing.provider
...
         )
-        strConstants.forEach { yieldStrings(it) { value } }
+        strConstants.forEach { yieldStrings(StringValue(it)) { value } }
+
+        strConstants
+            .filter {
+                it.isPattern()
+            }
+            .forEach {
+                yieldStrings(RegexValue(it, Random(0)), StringValue::value)
```
Markoutte, line 51: Java Fuzzing has the same logic, but it will be fixed, using `description.random`. I think, here `PythonMethodDescription` can have the random instance to pass anywhere it is need

Markoutte (approved):

PR #2371, opened 2023-07-06
Prioritize executions with identical trace to minimize `stateBefore`

### LINKED ISSUE #2219: Improve Spring integration tests minimization to minimize database content
**Description**

Spring integration tests store some information to database before running method under tests.
However, if there are two potential test methods for one branch, we should prefer the method without database communication or with minimal database communication, because it is faster and more safe.

We should improve minimization strategy to select method with minimal database interaction if all other minimization conditions are similar for them.

IlyaMuravjov:
## Description

Fixes #2219

See [this comment](https://github.com/UnitTestBot/UTBotJava/issues/2219#issuecomment-1617790482) for more details.

## How to test

Standalone testing is not required, just as a part of full Spring integration test generation testing.

## Self-check list


utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
```diff
@@ -423,6 +424,7 @@ class UtBotSymbolicEngine(
                     .with(ValueProvider.of(relevantRepositories.map { SavedEntityValueProvider(defaultIdGenerator, it) }))
                     .with(ValueProvider.of(generatedValueFieldIds.map { FieldValueProvider(defaultIdGenerator, it) }))
             }.let(transform)
+        val coverageToMinStateBeforeSize = mutableMapOf<Coverage, Int>()
```
Markoutte, line 427: It looks like this map is inefficient, because it requires at least O(n) to calculate hash and it stores as much coverages as program can generate. In bad cases it generates too many coverages because of loops and recursions. The `Trie` was created exactly for solving this problem. Still has O(n) to add value but has lower memory consumption.

Please, consider using `Trie` or another more efficient collection to store coverages (traces).

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
```diff
@@ -53,16 +41,17 @@ fun <T : Any> minimizeTestCase(
 
 fun minimizeExecutions(executions: List<UtExecution>): List<UtExecution> {
     val unknownCoverageExecutions =
-        executions.indices.filter { executions[it].coverage?.coveredInstructions?.isEmpty() ?: true }.toSet()
+        executions.filter { it.coverage?.coveredInstructions.isNullOrEmpty() }.toSet()
     // ^^^ here we add executions with empty or null coverage, because it happens only if a concrete execution failed,
     //     so we don't know the actual coverage for such executions
 
-    val filteredExecutions = executions.filterIndexed { idx, _ -> idx !in unknownCoverageExecutions }
+    val filteredExecutions = filterOutDuplicateCoverages(executions - unknownCoverageExecutions)
```
Markoutte, line 48: I'm not sure that this works as expected, because it compared indices (which are integers) then but now it compares UtExecutions for subtracting. But there's no `hashCode` and `equals` for UtExecution. Also, this implementation is less efficient when comparing values.

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
```diff
@@ -71,6 +60,18 @@ fun minimizeExecutions(executions: List<UtExecution>): List<UtExecution> {
     }
 }
 
+private fun filterOutDuplicateCoverages(executions: List<UtExecution>): List<UtExecution> {
+    val (executionIdxToCoveredEdgesMap, _) = buildMapping(executions)
```
Markoutte, line 64: This operation can be heavy when number of executions is huge. Could it be verified that it works OK for a huge number of executions?

Markoutte (changes requested):

utbot-framework/src/main/kotlin/org/utbot/engine/UtBotSymbolicEngine.kt
IlyaMuravjov: Replaced key type with `Trie.Node<Instruction>`, but as far as I can see the use of the `Trie` shouldn't reduce memory consumption, since reference to `Coverage` is stored in every emitted `UtExecution` anyway, that was my original motivation for not using the `Trie` in the first place.

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
IlyaMuravjov: It should work as expected, because reference equality is used when `hashCode` and `equals` are not overridden.

For some context, I changed this part, because there was a bug in minimization. To be more exact, in the following code fragment indices referring to positions of `UtExecution`s in `filteredExecutions` list were saved into `usedExecutionIndexes`:
```kotlin
val (mapping, executionToPriorityMapping) = buildMapping(filteredExecutions)
val usedExecutionIndexes = (GreedyEssential.minimize(mapping, executionToPriorityMapping) +  /* ... */).toSet()
```
And later on these `usedExecutionIndexes` were used as if they were indices that refer to positions of `UtExecution`s in the whole `executions` list:
```kotlin
val usedMinimizedExecutions = executions.filterIndexed { idx, _ -> idx in usedExecutionIndexes }
```

utbot-framework/src/main/kotlin/org/utbot/framework/minimization/Minimization.kt
IlyaMuravjov: If it doesn't _work OK for a huge number of executions_, then it should have already been noticed on functions with loops and recursions, beside that inspecting `buildMapping` function body we can see that it only iterates over instructions once.

Although, it worth noting that `buildMapping` may have some performance related issues, since it uses `allCoveredEdges`, that can get quite large and cause a lot of cache misses, but I still think it's premature to optimize it now without any evidence (nor even known incidents) of it being a bottleneck.
```kotlin
// (inst1, instr2) -> edge id --- edge represents as a pair of instructions, which are connected by this edge
val allCoveredEdges = mutableMapOf<Pair<Long, Long?>, Int>()
```

By the way, on the topic of filtering out duplicate coverages, is that intended that fuzzer and minimization do it differently? Right now, when engine produces multiple executions with identical coverage, but throwing different exceptions, all these executions are preserved by minimization, however fuzzer itself only keeps one execuction.

Executions can throw different exceptions, while having identical coverage, if exceptions are thrown from within the JRE itself or from some large libraries (e.g. Spring) that are not transformed for trace collection.

Markoutte (approved):

PR #2440, opened 2023-07-24
Make fuzzer use static methods to create arbitrary objects

### LINKED ISSUE #2437: Java fuzzer can't create any values for types with no public constructor (e.g. `LocalDateTime`)
**Description**

When creating instances of arbitrary classes Java fuzzer only considers their accessible constructors. However, some classes (e.g. `LocalDateTime`) only have private constructors while their instances are meant to be created using `static` methods.

**Real world (concrete) example**

Same as in #2428 

**To Reproduce**

Run the following unit test.
```kotlin
@Test
fun `fuzzer can create instances of classes without public constructors but with static factory method in their class`() {
    var seenLocalDateTime = false
    runBlockingWithContext {
        try {
            runJavaFuzzing(
                TestIdentityPreservingIdGenerator,
                methodUnderTest = LocalDateTime::getMinute.javaMethod!!.executableId,
                constants = emptyList(),
                names = emptyList(),
            ) { thisInstance, _, _ ->
                val control = runCatching {
                    ValueConstructor()
                        .construct(listOfNotNull(thisInstance?.model))
                        .singleOrNull()?.value
                }.getOrNull()?.let { constructedThisInstance ->
                    assertInstanceOf(LocalDateTime::class.java, constructedThisInstance)
                    seenLocalDateTime = true
                    Control.STOP
                } ?: Control.PASS
                BaseFeedback(Trie.emptyNode(), control)
            }
        } catch (e: Throwable) {
            e.printStackTrace()
        }
    }
    assertTrue(seenLocalDateTime) { "No value was generated for type LocalDateTime" }
}
```

**Expected behavior**

Test passes.

**Actual behavior**

Test fails.

~~~
No value was generated for type LocalDateTime ==> expected: <true> but was: <false>
~~~

IlyaMuravjov:
## Description

Now fuzzer can use both static methods and constructors to create instances of arbitrary types.

Fixes #2437 

### Automated tests

> utbot-java-fuzzing/src/test/kotlin/org/utbot/fuzzing/JavaFuzzingTest.kt

## Self-check list


utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -186,6 +186,12 @@ class AbstractsObjectValueProvider(
     }
 }
 
+private fun findAccessibleCreators(description: FuzzedDescription, classId: ClassId): Sequence<ExecutableId> =
+    (classId.allConstructors + (classId.jClass.methods
+                .filter { it.isStatic && it.returnType.id.isSubtypeOf(classId) }
```
Markoutte, line 191: Could you, please, check what will happen if generics are used? Something like:

```java
class A<T> {
    public static A<String> getStringTyped() { return new A<>("Test") }
    private A(T value) {};
}

...

public void testedMethod(A<Int> expected) { ... }
```

IlyaMuravjov: It will erroneously use `getStringTyped()` to create `A<Integer>`. I was assuming that java fuzzer doesn't attempt to fully respect generic types here, since similar issue could have been reproduced even before this pull request:
```java
interface A<T> {
    T get();
}

class AImpl implements A<String> {
    @Override
    public String get() {
        return "Test";
    } 
}

...

public void testedMethod(A<Integer> expected) { ... }
```

Markoutte: Fair enough, I tried to fix it with #2451. Please, review it when it is possible. I think same can be done for this PR when 2451 is merged.

Markoutte (approved):

PR #2446, opened 2023-07-25
Add bytecode transformation for if statement with call of String methods

egiptipavel:
## Description

Bytecode transformation of the if statement with call of the `String.equals`, `String.startsWith` or `String.endsWith` method with a constant string into a sequence of comparisons of each char of the string with each char of the constant string.

For example, this code
```Java
if (a.equals("abc")) {
    // block of code
}
```
at the bytecode level will be transformed to this
```Java
if (a.length() == 3) {
    if (a.charAt(0) == 'a') {
        if (a.charAt(1) == 'b') {
            if (a.charAt(2) == 'c') {
                // block of code
            }
        }
    }
}
```

### Automated tests

Run `org.utbot.examples.TestBytecodeTransformation` in `utbot-instrumentation-tests`  module

### Manual tests

```Java
public class Transformation {
    public boolean test1(String a) {
        if (a.equals(null)) {
            return true;
        }
        return false;
    }

    public boolean test2(String a) {
        if (a.equals("")) {
            return true;
        }
        return false;
    }

    public boolean test3(String a) {
        if (a.equals("abc")) {
            return true;
        }
        return false;
    }

    public boolean test4(String a) {
        if (a.startsWith("abc")) {
            return true;
        }
        return false;
    }

    public boolean test5(String a) {
        if (a.endsWith("def")) {
            return true;
        }
        return false;
    }

    public boolean test6(String a) {
        if (a.startsWith("")) {
            return true;
        }
        return false;
    }

    public boolean test7(String a) {
        if (a.endsWith("")) {
            return true;
        }
        return false;
    }
}
```

## Self-check list


utbot-instrumentation/src/main/kotlin/org/utbot/instrumentation/instrumentation/execution/SimpleUtExecutionInstrumentation.kt
```diff
@@ -130,15 +129,11 @@ class SimpleUtExecutionInstrumentation(
...
-        instrumenter.visitClass(object : ClassVisitorBuilder<BytecodeTransformer> {
-            override val writerFlags: Int
-                get() = 0
-
-            override val readerParsingOptions: Int
-                get() = 0
-
-            override fun build(writer: ClassWriter): BytecodeTransformer = BytecodeTransformer(writer)
-        })
+        if (UtSettings.useBytecodeTransformation) {
```
Markoutte, line 132: Instrumentation has no access to UtSettings. Because of this it fails to create any test

Markoutte (approved):

PR #2496, opened 2023-08-10
Add --fuzzing-ratio to CLI

Markoutte:
## Description

Now, CLI can be set with specific fuzzing ration instead of default 0.1. If no option is specified, then the default one is used.

## Self-check list

### DIFF
--- a/utbot-cli/src/main/kotlin/org/utbot/cli/GenerateTestsAbstractCommand.kt
+++ b/utbot-cli/src/main/kotlin/org/utbot/cli/GenerateTestsAbstractCommand.kt
@@ -7,6 +7,7 @@ import com.github.ajalt.clikt.parameters.options.multiple
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.unique
 import com.github.ajalt.clikt.parameters.types.choice
+import com.github.ajalt.clikt.parameters.types.double
 import com.github.ajalt.clikt.parameters.types.long
 import mu.KotlinLogging
 import org.utbot.common.PathUtil.classFqnToPath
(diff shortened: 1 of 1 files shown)


utbot-cli/src/main/kotlin/org/utbot/cli/GenerateTestsCommand.kt
```diff
@@ -149,9 +148,12 @@ class GenerateTestsCommand :
             projectRootPath == null -> {
                 println("The path to the project root is required to generate a report. Please, specify \"--project-root\" option.")
             }
+            sourceCodeFile == null -> {
```
denis-fokin, line 151: I am not sure if we need curly braces here )

utbot-cli/src/main/kotlin/org/utbot/cli/GenerateTestsCommand.kt
```diff
@@ -149,9 +148,12 @@ class GenerateTestsCommand :
             projectRootPath == null -> {
                 println("The path to the project root is required to generate a report. Please, specify \"--project-root\" option.")
```
denis-fokin, line 149: May be we should not ask specifying but suggest specifying the source code. Otherwise the user may think that it is mandatory for proper work of utility.

Markoutte: The `--project-root` is mandatory only when `--sarif` is set. In others cases it is not mandatory

utbot-cli/src/main/kotlin/org/utbot/cli/GenerateTestsAbstractCommand.kt
```diff
@@ -131,14 +125,20 @@ abstract class GenerateTestsAbstractCommand(name: String, help: String) :
         .long()
         .default(LONG_GENERATION_TIMEOUT)
 
+    private val fuzzingRation by option(
```
liblit-at-amazon, line 128: Nit: "ration" seems misused here. Consider calling this property `fuzzingRatio` instead of `fuzzingRation`.

utbot-cli/src/main/kotlin/org/utbot/cli/GenerateTestsAbstractCommand.kt
```diff
@@ -131,14 +125,20 @@ abstract class GenerateTestsAbstractCommand(name: String, help: String) :
...
+        "--fuzzing-ratio",
+        help = "Specify the ratio between symbolic engine and fuzzing"
```
liblit-at-amazon, line 130: Consider documenting (in the help text) what the default ratio is if not overridden. According to [this comment from @tyuldashev](https://github.com/orgs/UnitTestBot/discussions/2735#discussioncomment-8589010), the default is 0.1 fuzzing, 0.9 symbolic.

liblit-at-amazon, line 130: If we take the help text literally, then the default "ratio between symbolic engine and fuzzing" would be 0.9 ÷ 0.1 = 9. But is 9 a sensible value here, and would that be equivalent to the default? Or should the argument's value be 0.1 to mimic the default behavior? Or maybe 0.9?

I think that the option's name and help text need to be reconsidered. It's not clear that they correspond to how the provided value is actually used.

PR #2561, opened 2023-08-28
Support mock strategies and type replacement in Spring unit test fuzzing

IlyaMuravjov:
## Description

Makes fuzzing in Spring unit tests respect mock strategy and type replacement:
  * All types are replaced according to selected Spring configuration before they are passed to actual value providers.
  * `MockValueProvider` is used if mock strategy allows it and type isn't replaced, otherwise `ObjectValueProvider` is used (`MockValueProvider` is also used when no other way to create non-null value is found)
  * `InjectMockValueProvider` (one that outputs composite models) is only used to create `thisInstance`.

### Manual tests

Generate unit tests with every mocking strategy (including Spring-specific mocking strategy that uses Spring configuration) for `spring-boot-testing` project.

## Self-check list

### DIFF
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
@@ -70,9 +70,16 @@ fun anyObjectValueProvider(idGenerator: IdentityPreservingIdGenerator<Int>) =
         ovp.withFallback(AbstractsObjectValueProvider(idGenerator))
     }
 
+/**
+ * Marker interface that shows that this [JavaValueProvider] can potentially provide values of
+ * arbitrary types, unlike type-specific value providers that were designed to provide values of
+ * few specific popular types (e.g. `List`, `String`, etc.).
+ */
+interface AnyObjectValueProvider : JavaValueProvider
+
 class ObjectValueProvider(
     val idGenerator: IdGenerator<Int>,
-) : JavaValueProvider {
+) : AnyObjectValueProvider {
 
     override fun accept(type: FuzzedType) = !isIgnored(type.classId)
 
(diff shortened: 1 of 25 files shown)


Damtev:
@Markoutte, could you take a look, please?

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -70,9 +70,16 @@ fun anyObjectValueProvider(idGenerator: IdentityPreservingIdGenerator<Int>) =
         ovp.withFallback(AbstractsObjectValueProvider(idGenerator))
     }
 
+/**
+ * Marker interface that shows that this [JavaValueProvider] can potentially provide values of
+ * arbitrary types, unlike type-specific value providers that were designed to provide values of
+ * few specific popular types (e.g. `List`, `String`, etc.).
+ */
+interface AnyObjectValueProvider : JavaValueProvider
```
Markoutte, line 78: Marker interface usually doesn't extend any other interfaces and has no methods. This is because you could use several marker interfaces for one class. In this case `AnyObjectValueProvider` is used as real type in hierarchy, that maybe useful, but I think it is redundant.

Markoutte (commented):
For me it is hard to understand all type hierarchy in Spring module and all those wrappers, delegators and adapters. Maybe, documentation could help?

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
IlyaMuravjov: It's not redundant, because it's used in `MockingJavaFuzzingContext`, which avoids mocking types that have value providers other than `AnyObjectValueProvider`, because it's a bad practice to mock types like `List`, `Date`, and `Map`.

I can't simply list all implementers of `AnyObjectValueProvider` in `MockingJavaFuzzingContext`, because the set of these implementers is subject to change as #2583 shows.

Made `AnyObjectValueProvider` truly a marker interface that doesn't extend any interfaces.

EgorkaKulikov (approved):

PR #2583, opened 2023-09-08
Introduce object creation using builders for fuzzers

Markoutte:
## Description

We introduce new `BuilderObjectValueProvider` for Java that finds any method, that can create a value of given type. Together with `AbstractObjectValueProvider` it works only when ObjectValueProvider and others fails to build a value.

The provider searches any method from any class that can return object of this type or any type that can be inferenced from the given one. It doesn't matter if this method static or not, but if this method is not static, that we suppose that we have a builder class for it. As a builder class we try to find other methods that returns an instance of this builder. This logic added because of default way to create such builder, where any set or add method returns the instance of builder to support chain calling. This builder should work to find methods for factories and singletones as well.

To support this provider, we introduce new routine in fuzzer's API called `Modify` (the name maybe is not final). It is added for `Seed.Recurisve` and works as follows: it accepts a sequence of routines that changes or creates new object with given this instance and fuzzed values. For example, we firstly create builder's object and update it with builder method, then finally we call the method, that creates an object using this builder. This sequence is always called when creating an object and never changed its ordering.

In addition, you can use this method, apply some method to the instance and return the instance object itself, for example, to always close a `Closable` object.

As a result, `ContestEstimator` covers 53% of guava project code, before it covers only 43%.

### Automated tests

All fuzzer tests should pass.

### Manual tests

You can try to generate tests for this little project:

```java
public interface Muzzi {
    int getA();
}
```

```java
public class RealMuzzi implements Muzzi {

    private final int a;

    private RealMuzzi(int a) {
        this.a = a;
    }

    @Override
    public int getA() {
        return a;
    }

    public static class Builder {

        private int a;

        public Builder setA(int a) {
            this.a = a;
            return this;
        }

        public Muzzi build() {
            return new RealMuzzi(a);
        }
    }
}
```

```java
public class MuzziTest {
    public int size(Muzzi muzzi) {
        if (muzzi.getA() < 0) {
            return -1;
        } else if (muzzi.getA() > 0) {
            return 1;
        }
        return 0;
    }
}
```

## Self-check list

### DIFF
--- a/utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
+++ b/utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
@@ -142,7 +142,7 @@ sealed interface Seed<TYPE, RESULT> {
         val construct: Routine.Create<TYPE, RESULT>,
         val modify: Sequence<Routine.Call<TYPE, RESULT>> = emptySequence(),
         val empty: Routine.Empty<TYPE, RESULT>,
-        val finally: Routine.Modify<TYPE, RESULT>? = null,
+        val transformers: Sequence<Routine.Modify<TYPE, RESULT>> = emptySequence(),
     ) : Seed<TYPE, RESULT>
 
     /**
@@ -550,20 +550,26 @@ private fun <TYPE, RESULT, DESCRIPTION : Description<TYPE>, FEEDBACK : Feedback<
                         }
                     )
                 },
-            finally = task.finally?.let { finally ->
-                fuzz(
-                    finally.types,
-                    fuzzing,
-                    description,
-                    random,
-                    configuration,
-                    finally,
-                    state.copy {
-                        recursionTreeDepth++
-                        iterations = -1
-                        parameterIndex = -1
-                    }
-                )
+            transformers = task.transformers.let { transformer ->
+                if (transformer === emptySequence<Node<TYPE, RESULT>>() || transformer.none()) {
+                    emptyList()
+                } else {
+                    transformer.map { f ->
+                        fuzz(
+                            f.types,
+                            fuzzing,
+                            description,
+                            random,
+                            configuration,
+                            f,
+                            state.copy {
+                                recursionTreeDepth++
+                                iterations = -1
+                                parameterIndex = -1
+                            }
+                        )
+                    }.toList()
+                }
             }
         )
     } catch (nsv: NoSeedValueException) {
@@ -600,12 +606,16 @@ private fun <TYPE, R> create(result: Result<TYPE, R>): R = when(result) {
                 else -> error("Undefined object call method ${func.builder}")
             }
         }
-        finally?.let { func ->
-            when (val builder = func.builder) {
-                is Routine.Modify<TYPE, R> -> builder(obj, func.result.map { create(it) })
-                else -> error("Undefined object call method ${func.builder}")
+        transformers.let { transformers ->
+            var transformed = obj
+            transformers.forEach { transformer ->
+                transformed = when (val builder = transformer.builder) {
+                    is Routine.Modify<TYPE, R> -> builder(obj, transformer.result.map { create(it) })
+                    else -> error("Undefined object call method ${transformer.builder}")
+                }
             }
-        } ?: obj
+            transformed
+        }
     }
     is Result.Collection<TYPE, R> -> with(result) {
         val collection: R = when (val c = construct.builder) {
@@ -687,7 +697,7 @@ sealed interface Result<TYPE, RESULT> {
     class Recursive<TYPE, RESULT>(
         val construct: Node<TYPE, RESULT>,
         val modify: List<Node<TYPE, RESULT>>,
-        val finally: Node<TYPE, RESULT>?,
+        val transformers: List<Node<TYPE, RESULT>>,
     ) : Result<TYPE, RESULT>
 
     /**
--- a/utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
+++ b/utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
@@ -141,7 +141,8 @@ sealed interface Seed<TYPE, RESULT> {
     class Recursive<TYPE, RESULT>(
         val construct: Routine.Create<TYPE, RESULT>,
         val modify: Sequence<Routine.Call<TYPE, RESULT>> = emptySequence(),
-        val empty: Routine.Empty<TYPE, RESULT>
+        val empty: Routine.Empty<TYPE, RESULT>,
+        val finally: Routine.Modify<TYPE, RESULT>? = null,
     ) : Seed<TYPE, RESULT>
 
     /**
@@ -184,6 +185,13 @@ sealed class Routine<T, R>(val types: List<T>) : Iterable<T> by types {
         operator fun invoke(instance: R, arguments: List<R>) {
             callable(instance, arguments)
         }
+    }
+
+    class Modify<T, R>(
+        types: List<T>,
+        val callable: (instance: R, arguments: List<R>) -> R
+    ) : Routine<T, R>(types) {
+        operator fun invoke(instance: R, arguments: List<R>): R = callable(instance, arguments)
     }
 
     /**
@@ -541,7 +549,22 @@ private fun <TYPE, RESULT, DESCRIPTION : Description<TYPE>, FEEDBACK : Feedback<
                             parameterIndex = -1
                         }
                     )
-                }
+                },
+            finally = task.finally?.let { finally ->
+                fuzz(
+                    finally.types,
+                    fuzzing,
+                    description,
+                    random,
+                    configuration,
+                    finally,
+                    state.copy {
+                        recursionTreeDepth++
+                        iterations = -1
+                        parameterIndex = -1
+                    }
+                )
+            }
         )
     } catch (nsv: NoSeedValueException) {
         @Suppress("UNCHECKED_CAST")
@@ -577,7 +600,12 @@ private fun <TYPE, R> create(result: Result<TYPE, R>): R = when(result) {
                 else -> error("Undefined object call method ${func.builder}")
             }
         }
-        obj
+        finally?.let { func ->
+            when (val builder = func.builder) {
+                is Routine.Modify<TYPE, R> -> builder(obj, func.result.map { create(it) })
+                else -> error("Undefined object call method ${func.builder}")
+            }
+        } ?: obj
     }
     is Result.Collection<TYPE, R> -> with(result) {
         val collection: R = when (val c = construct.builder) {
@@ -659,6 +687,7 @@ sealed interface Result<TYPE, RESULT> {
     class Recursive<TYPE, RESULT>(
         val construct: Node<TYPE, RESULT>,
         val modify: List<Node<TYPE, RESULT>>,
+        val finally: Node<TYPE, RESULT>?,
     ) : Result<TYPE, RESULT>
 
     /**
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
@@ -191,6 +191,12 @@ class BuilderObjectValueProvider(
                 sootClass.methods.asSequence()
                     .filter { isAccessible(it, description.description.packageName) }
                     .filter { (implementors.contains((it.returnType as? RefType)?.sootClass)) }
+                    .filter {
+                        // Simple check whether the method is synthetic,
+                        // because Soot has no information about this:
+                        // https://mailman.cs.mcgill.ca/pipermail/soot-list/2012-November/004972.html
+                        !it.name.matches(nameWithDigitsAfterSpecialCharRegex)
+                    }
                     .forEach { method ->
                         val returnType = method.returnType as RefType
                         val returnClassId = returnType.classId
@@ -258,6 +264,8 @@ class BuilderObjectValueProvider(
 
 }
 
+private val nameWithDigitsAfterSpecialCharRegex = """.*\$\d+$""".toRegex()
+
 private fun SootClass.isUseful(): Boolean {
     if (!isConcrete) return false
     val packageName = packageName
@@ -269,7 +277,7 @@ private fun SootClass.isUseful(): Boolean {
         )
             return false
     }
-    val isAnonymousClass = name.matches(""".*\$\d+$""".toRegex())
+    val isAnonymousClass = name.matches(nameWithDigitsAfterSpecialCharRegex)
     if (isAnonymousClass) {
         return false
     }
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
@@ -22,7 +22,6 @@ import org.utbot.framework.util.executableId
 import org.utbot.fuzzer.FuzzedType
 import org.utbot.fuzzer.FuzzedValue
 import org.utbot.fuzzer.IdGenerator
-import org.utbot.fuzzer.IdentityPreservingIdGenerator
 import org.utbot.fuzzer.fuzzed
 import org.utbot.fuzzing.FuzzedDescription
 import org.utbot.fuzzing.JavaValueProvider
@@ -57,13 +56,6 @@ private fun isIgnored(type: ClassId): Boolean {
             || type.isAbstract
             || (type.isInner && !type.isStatic)
 }
-
-fun anyObjectValueProvider(idGenerator: IdentityPreservingIdGenerator<Int>) =
-    ObjectValueProvider(idGenerator).letIf(UtSettings.fuzzingImplementationOfAbstractClasses) { ovp ->
-        ovp.withFallback(
-            AbstractsObjectValueProvider(idGenerator).with(BuilderObjectValueProvider(idGenerator))
-        )
-    }.withFallback(AnyDepthNullValueProvider)
 
 class ObjectValueProvider(
     val idGenerator: IdGenerator<Int>,
@@ -109,7 +101,7 @@ class ObjectValueProvider(
                     when {
                         fd.canBeSetDirectly -> {
                             yield(Routine.Call(listOf(fd.type)) { self, values ->
-                                val model = self.model as UtAssembleModel
+                                val model = self.model as? UtAssembleModel ?: return@Call
                                 model.modificationsChain as MutableList += UtDirectSetFieldModel(
                                     model,
                                     FieldId(classId, fd.name),
@@ -120,7 +112,7 @@ class ObjectValueProvider(
 
                         fd.setter != null && fd.getter != null -> {
                             yield(Routine.Call(listOf(fd.type)) { self, values ->
-                                val model = self.model as UtAssembleModel
+                                val model = self.model as? UtAssembleModel ?: return@Call
                                 model.modificationsChain as MutableList += UtExecutableCallModel(
                                     model,
                                     fd.setter.executableId,
@@ -245,7 +237,7 @@ class BuilderObjectValueProvider(
                                          }
                                     },
                                 empty = nullRoutine(returnClassId),
-                                finally = Routine.Modify(parameters) { instance, values ->
+                                transformers = sequenceOf(Routine.Modify(parameters) { instance, values ->
                                     UtAssembleModel(
                                         idGenerator.createId(),
                                         returnClassId,
@@ -256,7 +248,7 @@ class BuilderObjectValueProvider(
                                             values.map { it.model }
                                         ),
                                     ).fuzzed {  }
-                                }
+                                })
                             ))
                         }
                     }
@@ -272,7 +264,8 @@ private fun SootClass.isUseful(): Boolean {
     if (packageName != null) {
         if (packageName.startsWith("jdk.internal") ||
             packageName.startsWith("org.utbot") ||
-            packageName.startsWith("sun.")
+            packageName.startsWith("sun.") ||
+            packageName.startsWith("com.sun.")
         )
             return false
     }
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
@@ -300,14 +300,13 @@ class AbstractsObjectValueProvider(
                 }
                 val jClass = sc.id.jClass
                 return isAccessible(jClass, description.description.packageName) &&
-                        jClass.declaredConstructors.any { isAccessible(it, description.description.packageName) } &&
-                        jClass.let {
-                            // This won't work in case of implementations with generics like `Impl<T> implements A<T>`.
-                            // Should be reworked with accurate generic matching between all classes.
-                            toFuzzerType(it, description.typeCache).traverseHierarchy(description.typeCache)
-                                .contains(type)
-                        }
+                        jClass.declaredConstructors.any { isAccessible(it, description.description.packageName) }
+                        // This won't work in case of implementations with generics like `Impl<T> implements A<T>`.
+                        // Should be reworked with accurate generic matching between all classes.
+                        && toFuzzerType(jClass, description.typeCache).traverseHierarchy(description.typeCache)
+                            .contains(type)
             } catch (ignore: Throwable) {
+                logger.warn("Cannot resolve class $sc", ignore)
                 return false
             }
         }
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
@@ -2,13 +2,7 @@ package org.utbot.fuzzing.providers
 
 import mu.KotlinLogging
 import org.utbot.framework.UtSettings
-import org.utbot.framework.plugin.api.ClassId
-import org.utbot.framework.plugin.api.ConstructorId
-import org.utbot.framework.plugin.api.FieldId
-import org.utbot.framework.plugin.api.UtAssembleModel
-import org.utbot.framework.plugin.api.UtDirectSetFieldModel
-import org.utbot.framework.plugin.api.UtExecutableCallModel
-import org.utbot.framework.plugin.api.id
+import org.utbot.framework.plugin.api.*
 import org.utbot.framework.plugin.api.util.classClassId
 import org.utbot.framework.plugin.api.util.constructor
 import org.utbot.framework.plugin.api.util.dateClassId
@@ -24,6 +18,7 @@ import org.utbot.framework.plugin.api.util.isStatic
 import org.utbot.framework.plugin.api.util.jClass
 import org.utbot.framework.plugin.api.util.method
 import org.utbot.framework.plugin.api.util.stringClassId
+import org.utbot.framework.util.executableId
 import org.utbot.fuzzer.FuzzedType
 import org.utbot.fuzzer.FuzzedValue
 import org.utbot.fuzzer.IdGenerator
@@ -40,13 +35,11 @@ import org.utbot.fuzzing.utils.hex
 import org.utbot.modifications.AnalysisMode
 import org.utbot.modifications.FieldInvolvementMode
 import org.utbot.modifications.UtBotFieldsModificatorsSearcher
+import soot.RefType
 import soot.Scene
 import soot.SootClass
-import java.lang.reflect.Field
-import java.lang.reflect.Member
-import java.lang.reflect.Method
-import java.lang.reflect.Modifier
-import java.lang.reflect.TypeVariable
+import soot.SootMethod
+import java.lang.reflect.*
 
 private val logger = KotlinLogging.logger {}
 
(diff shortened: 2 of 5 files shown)


utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
```diff
@@ -141,7 +141,8 @@ sealed interface Seed<TYPE, RESULT> {
     class Recursive<TYPE, RESULT>(
         val construct: Routine.Create<TYPE, RESULT>,
         val modify: Sequence<Routine.Call<TYPE, RESULT>> = emptySequence(),
-        val empty: Routine.Empty<TYPE, RESULT>
+        val empty: Routine.Empty<TYPE, RESULT>,
+        val transformers: Sequence<Routine.Modify<TYPE, RESULT>> = emptySequence(),
```
IlyaMuravjov, line 145: If `transformers` are added then wherever a `Recursive` seed is constructed based on another `Recursive` seed `transformers` should be preserved, that includes `ModifyingWithMethodsProviderWrapper` and `PropertyPreservingValueProvider`.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+            Scene.v().activeHierarchy.let { it.getImplementersOf(t) + it.getSubinterfacesOf(t) + t }.toSet()
+        } else {
+            Scene.v().activeHierarchy.let { it.getSubclassesOf(t) + t }.toSet()
+        }
+        Scene.v().classes.forEach { sootClass ->
+            val classId = sootClass.id
+            if (sootClass.isUseful() && isAccessible(sootClass, description.description.packageName)) {
+                sootClass.methods.asSequence()
+                    .filter { isAccessible(it, description.description.packageName) }
+                    .filter { (implementors.contains((it.returnType as? RefType)?.sootClass)) }
```
IlyaMuravjov, line 193: Why not just check that `it.returnType` is a subtype of `t` to avoid collecting `implementors` at all?

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+                    .filter {
+                        // Simple check whether the method is synthetic,
+                        // because Soot has no information about this:
+                        // https://mailman.cs.mcgill.ca/pipermail/soot-list/2012-November/004972.html
+                        !it.name.matches(nameWithDigitsAfterSpecialCharRegex)
```
IlyaMuravjov, line 198: Bridge methods are synthetic, but don't have any suffix. Why not (also) check `!it.executableId.executable.isSynthetic`, while catching exceptions in case method name in Soot and bytecode don't match?

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+                    }
+                    .forEach { method ->
+                        val returnType = method.returnType as RefType
+                        val returnClassId = returnType.classId
+                        val isStaticMethod = method.isStatic
+                        val parameters = method.parameterTypes.map {
+                            toFuzzerType(it.classId.jClass, description.typeCache)
+                        } as MutableList<FuzzedType>
```
IlyaMuravjov, line 206: Here generics get lost. Can you use `resolveParameterTypes(method.executableId, type.jType)` that I added in #2584 so generic types can be inferred based on the return type?

I'll undraft #2584 when tests are added for it.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+                                        idGenerator.createId(),
+                                        returnClassId,
+                                        method.toString(),
+                                        UtExecutableCallModel(
+                                            null,
+                                            method.executableId,
+                                            params.map { it.model }
+                                        ),
+                                        modificationsChainProvider = { mutableListOf() }
+                                    ).fuzzed { }
```
IlyaMuravjov, line 220: Is summery intended to be missing here? I'd suggest something like:
```kotlin
summary = "%var% = ${when (method.executableId) {
    is ConstructorId -> classId.simpleName
    is MethodId -> creatorExecutableId.simpleNameWithClass
}}(${creatorExecutableId.parameters.joinToString { it.simpleName }})"
```

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+                                    UtAssembleModel(
+                                        idGenerator.createId(),
+                                        returnClassId,
+                                        method.toString(),
+                                        UtExecutableCallModel(
+                                            instance.model as? UtAssembleModel ?: return@Modify nullFuzzedValue(returnClassId),
+                                            method.executableId,
+                                            values.map { it.model }
+                                        ),
+                                    ).fuzzed {  }
```
IlyaMuravjov, line 256: Same here

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+                                             val utAssembleModel = instance.model as? UtAssembleModel ?: return@Call
+                                             (utAssembleModel.modificationsChain as MutableList<UtStatementModel>).add(UtExecutableCallModel(
+                                                     utAssembleModel,
+                                                     builderMethod.executableId,
+                                                     values.map { it.model }
+                                                 ))
+                                         }
+                                    },
+                                empty = nullRoutine(returnClassId),
+                                transformers = sequenceOf(Routine.Modify(parameters) { instance, values ->
```
IlyaMuravjov, line 246: This can be rewritten without the use of `transformers`. Although there are two reasons to keep `transformers`:
* if aliasing is added, intermediate steps of `transformers` can be aliased;
* `MutationFactory` can be made to handle `transformers` and multi-arg `construct` differently.

```kotlin
Seed.Recursive(
    construct = Routine.Create(listOf(builderInstance) + parameters) { values ->
        UtAssembleModel(
            idGenerator.createId(),
            returnClassId,
            method.toString(),
            UtExecutableCallModel(
                values.first().model as? UtAssembleModel ?: return@Create nullFuzzedValue(returnClassId),
                method.executableId,
                values.drop(1).map { it.model }
            ),
        ).fuzzed {  }
    },
    modify = sootClass.methods.asSequence()
        .filter { isAccessible(it, description.description.packageName) }
        .filter { (it.returnType as? RefType)?.sootClass == sootClass }
        .map { builderMethod ->
            Routine.Call(builderMethod.parameterTypes.map {
                toFuzzerType(it.classId.jClass, description.typeCache)
            }) { instance, values ->
                val utAssembleModel =
                    ((instance.model as? UtAssembleModel)?.instantiationCall?.instance as? UtAssembleModel)
                        ?: return@Call
                (utAssembleModel.modificationsChain as MutableList<UtStatementModel>).add(UtExecutableCallModel(
                    utAssembleModel,
                    builderMethod.executableId,
                    values.map { it.model }
                ))
            }
        },
    empty = nullRoutine(returnClassId)
)
```

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +167,123 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+    val packageName = packageName
+    if (packageName != null) {
+        if (packageName.startsWith("jdk.internal") ||
+            packageName.startsWith("org.utbot") ||
+            packageName.startsWith("sun.") ||
+            packageName.startsWith("com.sun.")
+        )
+            return false
+    }
+    val isAnonymousClass = name.matches(nameWithDigitsAfterSpecialCharRegex)
```
IlyaMuravjov, line 280: Non-anonymous classes can actually be declared inside of anonymous classes resulting in a name like `Outer$1$Inner`, that is not matched by the regex, even though such classes are probably still "useless" for us.

IlyaMuravjov (changes requested):

utbot-fuzzing/src/main/kotlin/org/utbot/fuzzing/Api.kt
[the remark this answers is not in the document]
Markoutte: I see. I don't know anything about this part of code. Please, fix it in your code and push it into this branch or create a separate PR.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -114,7 +101,7 @@ class ObjectValueProvider(
                     when {
                         fd.canBeSetDirectly -> {
                             yield(Routine.Call(listOf(fd.type)) { self, values ->
-                                val model = self.model as UtAssembleModel
+                                val model = self.model as? UtAssembleModel ?: return@Call
```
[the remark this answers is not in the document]
Markoutte: Removed these checks. I saw the problem with UtAssembleModel that leads to JVM using assembleModel method and these checks helped somehow. Will investigate this later.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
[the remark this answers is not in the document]
Markoutte: It is not required, because SootClass keeps the list of methods and it. This builder is fast only because it uses Soot whenever it is possible and rarely uses other things like reflection.

Markoutte: Here only SootClasses is used. There's a number of different methods to understand if one SootClass implements or extends another class (), and all those methods need some time for hierarchy traversing. It is much easier to collect all SootClass into the set and checks whether or not some another SootClass in this set.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
Markoutte: It loads a method from the classpath that is 80x times longer than this simple check (on my local machine run).  I don't think it is problem right now.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
Markoutte: Let's try it after #2584 merge and do it in separate PR.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
Markoutte: Added

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
Markoutte: Added

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
IlyaMuravjov: To use generics we would have to use reflection anyway. Perhaps, to avoid it being a bottleneck we could limit maximum number of methods considered, similarly to how it's done in `AbstractsObjectValueProvider` with `take(10)`.

I suppose, bridge methods are rare enough to ignore handling them for now, although when present they will degrade performance, making value construction phase fail with `ClassCastException` when they are called directly.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +168,135 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+
+    override fun generate(description: FuzzedDescription, type: FuzzedType) = sequence<Seed<FuzzedType, FuzzedValue>> {
+        val t = try {
+            Scene.v().getRefType(type.classId.name).sootClass
+        } catch (ignore: Throwable) {
+            logger.error(ignore) { "Soot may be not initialized" }
+            return@sequence
+        }
+        val implementors = if (t.isInterface) {
+            Scene.v().activeHierarchy.let { it.getImplementersOf(t) + it.getSubinterfacesOf(t) + t }.toSet()
```
IlyaMuravjov, line 185: Why is `activeHierarchy` used here while `AbstractsObjectValueProvider` uses `fastHierarchy`. By the way, why don't we use `orMakeFastHierarchy` in `AbstractsObjectValueProvider`?

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
```diff
@@ -180,6 +168,135 @@ object AnyDepthNullValueProvider : JavaValueProvider {
...
+                        // because Soot has no information about this:
+                        // https://mailman.cs.mcgill.ca/pipermail/soot-list/2012-November/004972.html
+                        !it.name.matches(nameWithDigitsAfterSpecialCharRegex)
+                    }
+                    .forEach { method ->
+                        val returnType = method.returnType as RefType
+                        val returnClassId = returnType.classId
+                        val isStaticMethod = method.isStatic
+                        val parameters = method.parameterTypes.map {
+                            toFuzzerType(it.classId.jClass, description.typeCache)
```
IlyaMuravjov, line 206: `jClass` here can throw `ClassNotFoundException` if one of parameter classes is not present on the classpath, while the class declaring the method is present on the classpath.

For example, for me that happened while trying to use method `org.springframework.format.number.money.CurrencyUnitFormatter.print(javax.money.CurrencyUnit.CurrencyUnit, Locale): String` to create `CharSequence`, while not having `javax.money.CurrencyUnit.CurrencyUnit` on the classpath.

NOTE: it's normal to have classes on the classpath that reference classes that are not on the classpath, for example, some library `A` may provide adapters that make it possible to use it with another library `B` while leaving it up to the users to install or not to install the library `B`.

IlyaMuravjov (changes requested):

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
Markoutte: Added try-catch to handle such situations. Please, check it

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Objects.kt
Markoutte: Good question, for me there's no difference. Changed fastHierarchy in AxtractsObjectValueProvider to activeHierarchy, because the former is not used anywhere except those place.

IlyaMuravjov:
Now it finds bridge method `org.springframework.format.number.money.CurrencyUnitFormatter.print(Object, Locale): String` and latter crashes with `ClassNotFoundException: javax.money.CurrencyUnit` when trying to create a value for `CurrencyUnitFormatter`.

I think the codebase is not ready to deal with arbitrary classes that can be found on the classpath and the best solution for now is to make sure that we only create `FuzzedType` instances from classes that only mention classes that are on the classpath. 

In other words some `ValueProvider`s expect that using `fuzzedType.classId.allMethods`, `fuzzedType.classId.allConstructors`, and `fuzzedType.classId.allDeclaredFieldIds` won't throw an exception.

The easiest way to achieve that, that I can see is to make `runJavaFuzzing` pass `ClassDependenciesCheckingValueProvider(delegate = ValueProvider.of(providers)` to `runFuzzing` where `ClassDependenciesCheckingValueProvider.generate(description, type)`:
   * returns `sequenceOf(Seed.Simple(nullFuzzedValue(classClassId))` if `type.classId.allMethods`, `type.classId.allConstructors`, or `type.classId.allDeclaredFieldIds` throws any exceptions
   * otherwise it returns `delegate.generate(description, type)`, preferably filtering out routines that accept classes, mentioning classes that are not on the classpath

Markoutte:
@IlyaMuravjov I think that there're a many of other problems that we won't find by this review. Let's create separate issues with reproducers and stack traces for them and fix them. The main goal of this PR is to add this builder for Java, not for Spring. And I'd like to keep Java part as simple as possible without additional wrappers and delegators.

IlyaMuravjov:
@Markoutte 
I've pushed a commit that adds a decorator letting Spring test generation to work with `BuilderObjectValueProvider` without crashing with `ClassNotFoundException`. However, this problem is not Spring-specific, so I've also made an issue #2594 showing how it can be reproduced without Spring.

I merged `main` to resolve conflicts instead of rebasing onto `main`, because I didn't want to force push into your branch.

Also, this PR reveled a problem in instrumnted-process that marks executions as timed out even if the thing that actually timed out is the argument creation prior to method under test invocation, see #2595.

Next, it seems that `Scene.v().classes` doesn't contain all classes on the classpath, rather it only includes:
   * classes that are reachable from classes located in the `.class` files (as opposed to `.jar` files)
   * some classes from the JDK
   * classes that are reachable from `org.utbot.framework.util.classesToLoad`

Meaning that for large projects `BuilderObjectValueProvider` can get significantly slower and the result of `53%` stated in the PR description depends on the project under test being provided as `.jar` file (similarly to how it's done in the `ContestEstimator`) and not `.class` files (similarly to how it's done when UtBot is used via Intellij plugin).

Finally, I tried to check by how much providing `guava` in form of `.class` files instead of `.jar` file affects performance, but when I tried generating tests for the following method, the fuzzer crashed with the following exception.
```java
public static boolean isNull(LocalDateTime localDateTime) {
    if (localDateTime == null) return true;
    return false;
}
``` 
```
23:32:56.036 | INFO  | JavaLanguage              | Fuzzing is stopped because of an error
java.lang.ClassNotFoundException: com.google.common.io.TempFileCreator$JavaNioCreator$lambda_userPermissions_3__261
	at java.net.URLClassLoader.findClass(URLClassLoader.java:445) ~[?:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:587) ~[?:?]
	at org.utbot.common.FallbackClassLoader.loadClass(FallbackClassLoader.kt:41) ~[utbot-core-2023.09-SNAPSHOT.jar:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:520) ~[?:?]
	at org.utbot.framework.plugin.api.util.IdUtilKt.getJClass(IdUtil.kt:105) ~[utbot-framework-api-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.providers.BuilderObjectValueProvider$generate$1.invokeSuspend(Objects.kt:237) ~[utbot-java-fuzzing-2023.09-SNAPSHOT.jar:?]
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.SequenceBuilderIterator.hasNext(SequenceBuilder.kt:129) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at org.utbot.fuzzing.ValueProvider$Combined$generate$1.invokeSuspend(Providers.kt:231) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.SequenceBuilderIterator.hasNext(SequenceBuilder.kt:129) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at org.utbot.fuzzing.ValueProvider$Combined$generate$1.invokeSuspend(Providers.kt:231) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.SequenceBuilderIterator.hasNext(SequenceBuilder.kt:129) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.FlatteningSequence$iterator$1.ensureItemIterator(Sequences.kt:307) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.FlatteningSequence$iterator$1.hasNext(Sequences.kt:303) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.SequencesKt___SequencesKt.toCollection(_Sequences.kt:787) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.SequencesKt___SequencesKt.toMutableList(_Sequences.kt:817) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlin.sequences.SequencesKt___SequencesKt.toList(_Sequences.kt:808) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at org.utbot.fuzzing.FuzzingApi$produce$seeds$1.apply(Api.kt:427) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi$produce$seeds$1.apply(Api.kt:426) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at java.util.HashMap.computeIfAbsent(HashMap.java:1220) ~[?:?]
	at org.utbot.fuzzing.FuzzingApi.produce(Api.kt:426) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz(Api.kt:399) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.reduce(Api.kt:484) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.produce(Api.kt:438) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz(Api.kt:399) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.reduce(Api.kt:519) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.produce(Api.kt:437) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz(Api.kt:399) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.reduce(Api.kt:519) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.produce(Api.kt:437) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz(Api.kt:399) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.reduce(Api.kt:539) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.produce(Api.kt:437) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz(Api.kt:399) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz$fuzzOne(Api.kt:331) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.fuzz(Api.kt:350) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi.access$fuzz(Api.kt:1) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at org.utbot.fuzzing.FuzzingApi$fuzz$2.invokeSuspend(Api.kt) ~[utbot-fuzzing-2023.09-SNAPSHOT.jar:?]
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) ~[kotlin-stdlib-1.8.10.jar:1.8.10-release-430(1.8.10)]
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:284) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:85) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:38) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source) ~[kotlinx-coroutines-core-jvm-1.6.3.jar:?]
	at org.utbot.common.ConcurrencyKt.runBlockingWithCancellationPredicate(Concurrency.kt:38) ~[utbot-core-2023.09-SNAPSHOT.jar:?]
	at org.utbot.framework.plugin.api.TestCaseGenerator$generate$2$3.invoke(TestCaseGenerator.kt:204) ~[utbot-framework-2023.09-SNAPSHOT.jar:?]
	at org.utbot.framework.plugin.api.TestCaseGenerator$generate$2$3.invoke(TestCaseGenerator.kt:203) ~[utbot-framework-2023.09-SNAPSHOT.jar:?]
	at org.utbot.common.ConcurrencyKt.runIgnoringCancellationException(Concurrency.kt:47) ~[utbot-core-2023.09-SNAPSHOT.jar:?]
	at org.utbot.framework.plugin.api.TestCaseGenerator.generate(TestCaseGenerator.kt:203) ~[utbot-framework-2023.09-SNAPSHOT.jar:?]
	at org.utbot.framework.process.EngineProcessMainKt$setup$4.invoke(EngineProcessMain.kt:127) ~[utbot-framework-2023.09-SNAPSHOT.jar:?]
	at org.utbot.framework.process.EngineProcessMainKt$setup$4.invoke(EngineProcessMain.kt:112) ~[utbot-framework-2023.09-SNAPSHOT.jar:?]
	at org.utbot.rd.IdleWatchdog$measureTimeForActiveCall$1$2$1.invoke(ClientProcessUtil.kt:115) ~[utbot-rd-2023.09-SNAPSHOT.jar:?]
	at org.utbot.rd.IdleWatchdog.wrapActive(ClientProcessUtil.kt:88) ~[utbot-rd-2023.09-SNAPSHOT.jar:?]
	at org.utbot.rd.IdleWatchdog$measureTimeForActiveCall$1.invoke(ClientProcessUtil.kt:114) ~[utbot-rd-2023.09-SNAPSHOT.jar:?]
	at com.jetbrains.rd.framework.IRdEndpoint$set$1.invoke(TaskInterfaces.kt:182) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.IRdEndpoint$set$1.invoke(TaskInterfaces.kt:182) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.impl.RdCall.onWireReceived(RdTask.kt:362) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.MessageBroker$invoke$2$2.invoke(MessageBroker.kt:57) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.MessageBroker$invoke$2$2.invoke(MessageBroker.kt:56) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.impl.ProtocolContexts.readMessageContextAndInvoke(ProtocolContexts.kt:148) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.MessageBroker$invoke$2.invoke(MessageBroker.kt:56) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.framework.MessageBroker$invoke$2.invoke(MessageBroker.kt:54) ~[rd-framework-2023.1.2.jar:?]
	at com.jetbrains.rd.util.threading.SingleThreadSchedulerBase.queue$lambda-3(SingleThreadScheduler.kt:41) ~[rd-core-2023.1.2.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]
```

PR #2610, opened 2023-09-20
Add more correct type resolving for collections and maps

### LINKED ISSUE #2571: Fuzzer ignores generics of `class StringMap extends HashMap<String, String>`
**Description**

Fuzzer tries to put arbitrary `Objects` into `StringMap`

**To Reproduce**

* Set fuzzing to `100%`
* Generate tests for the following class

```java
public class StringMap extends HashMap<String, String> {
    public String getAnyString() {
        return keySet().iterator().next();
    }
}
```

**Expected behavior**

There's one passing and one failing test.

**Actual behavior**

No passing tests are generated, one of the tests fails to compile with the following error: "incompatible types: `Object` cannot be converted to `String` `stringMap.put(object, object1);`".

```java
/**
 * @utbot.classUnderTest {@link StringMap}
 * @utbot.methodUnderTest {@link StringMap#getAnyString()}
 */
@Test
@DisplayName("getAnyString:  -> throw ClassCastException")
public void testGetAnyStringThrowsCCE() {
    StringMap stringMap = new StringMap();
    Object object = new Object();
    Object object1 = new Object();
    stringMap.put(object, object1);
    
    /* This test fails because method [org.example.StringMap.getAnyString] produces [java.lang.ClassCastException: class java.lang.Object cannot be cast to class java.lang.String (java.lang.Object and java.lang.String are in module java.base of loader 'bootstrap')]
        org.example.StringMap.getAnyString(StringMap.java:7) */
    stringMap.getAnyString();
}
```

Markoutte:
## Description

Fixes #2571 

The generic retrieving logic for collections and maps was naive. This PR improves that logic using guava type resolving to find out correct types. For every type it tries to resolve a type of returned values when calling such methods as `Map.keySet`, `Map.values` and `Collection.iterator`. A generic used in the returned type of those methods is a target type used for building a collection or a map.

Note, that the fix reveals another problem in the code generator. Because it uses untyped model it fails to generate a compilable test. For example, the result of fuzzing `getAnyString` method from the issue is follow:

```
    @Test
    @DisplayName("getAnyString: ")
    public void testGetAnyString() {
        Issue2571 issue2571 = new Issue2571();
        issue2571.put(((Object) "XZ"), ((Object) "abc"));

        String actual = issue2571.getAnyString();

        String expected = "XZ";

        assertEquals(expected, actual);
    }
```

This line `issue2571.put(((Object) "XZ"), ((Object) "abc"));` has syntax error because the map expects a string, not an object. @EgorkaKulikov , please, check this one.

### Automated tests

New tests added:

> utbot-java-fuzzing/src/test/kotlin/org/utbot/fuzzing/JavaValueProviderTest.kt`

### Manual tests

All manual tests from collections samples should pass.

## Self-check list

### DIFF
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
@@ -7,10 +7,9 @@ import org.utbot.fuzzer.FuzzedType
 import org.utbot.fuzzer.FuzzedValue
 import org.utbot.fuzzer.IdGenerator
 import org.utbot.fuzzer.fuzzed
+import org.utbot.fuzzer.jType
 import org.utbot.fuzzing.*
-import org.utbot.fuzzing.spring.utils.jType
 import org.utbot.fuzzing.utils.hex
-import java.lang.reflect.Method
 import kotlin.reflect.KClass
 
 class EmptyCollectionValueProvider(
--- a/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
+++ b/utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
@@ -1,5 +1,6 @@
 package org.utbot.fuzzing.providers
 
+import com.google.common.reflect.TypeToken
 import org.utbot.framework.plugin.api.*
 import org.utbot.framework.plugin.api.util.*
 import org.utbot.fuzzer.FuzzedType
@@ -7,7 +8,9 @@ import org.utbot.fuzzer.FuzzedValue
 import org.utbot.fuzzer.IdGenerator
 import org.utbot.fuzzer.fuzzed
 import org.utbot.fuzzing.*
+import org.utbot.fuzzing.spring.utils.jType
 import org.utbot.fuzzing.utils.hex
+import java.lang.reflect.Method
 import kotlin.reflect.KClass
 
 class EmptyCollectionValueProvider(
@@ -80,9 +83,25 @@ class EmptyCollectionValueProvider(
 class MapValueProvider(
     idGenerator: IdGenerator<Int>
 ) : CollectionValueProvider(idGenerator, java.util.Map::class.id) {
+
+    private enum class MethodCall { KEYS, VALUES }
+
+    private fun findTypeByMethod(description: FuzzedDescription, type: FuzzedType, method: MethodCall): FuzzedType {
+        val methodName = when (method) {
+            MethodCall.KEYS -> "keySet"
+            MethodCall.VALUES -> "values"
+        }
+        val m = Map::class.java.getMethod(methodName)
+        return resolveTypeByMethod(description, type, m)?.let {
+            assert(it.classId.isSubtypeOf(collectionClassId))
+            assert(it.generics.size == 1)
+            it.generics[0]
+        } ?: FuzzedType(objectClassId)
+    }
+
     override fun resolveType(description: FuzzedDescription, type: FuzzedType) = sequence {
-        val keyGeneric = type.generics.getOrNull(0) ?: FuzzedType(objectClassId)
-        val valueGeneric = type.generics.getOrNull(1) ?: FuzzedType(objectClassId)
+        val keyGeneric = findTypeByMethod(description, type, MethodCall.KEYS)
+        val valueGeneric = findTypeByMethod(description, type, MethodCall.VALUES)
         when (type.classId) {
             java.util.Map::class.id -> {
                 if (keyGeneric.classId isSubtypeOf Comparable::class) {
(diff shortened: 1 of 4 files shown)


utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
```diff
@@ -1,13 +1,16 @@
...
 
+import com.google.common.reflect.TypeToken
 import org.utbot.framework.plugin.api.*
 import org.utbot.framework.plugin.api.util.*
 import org.utbot.fuzzer.FuzzedType
 import org.utbot.fuzzer.FuzzedValue
 import org.utbot.fuzzer.IdGenerator
 import org.utbot.fuzzer.fuzzed
 import org.utbot.fuzzing.*
+import org.utbot.fuzzing.spring.utils.jType
```
IlyaMuravjov, line 11: If `TypeUtils` are used here, then I think they should be moved out of `spring` package, similarly to how it's done in #2584

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
```diff
@@ -80,9 +83,25 @@ class EmptyCollectionValueProvider(
...
+            assert(it.generics.size == 1)
+            it.generics[0]
+        } ?: FuzzedType(objectClassId)
+    }
+
     override fun resolveType(description: FuzzedDescription, type: FuzzedType) = sequence {
-        val keyGeneric = type.generics.getOrNull(0) ?: FuzzedType(objectClassId)
-        val valueGeneric = type.generics.getOrNull(1) ?: FuzzedType(objectClassId)
+        val keyGeneric = findTypeByMethod(description, type, MethodCall.KEYS)
+        val valueGeneric = findTypeByMethod(description, type, MethodCall.VALUES)
```
IlyaMuravjov, line 104: I was expecting that you'd add something like this to `TypeUtils`:
```kotlin
private fun resolveGenericsOfSuperClass(
    type: FuzzedType,
    description: FuzzedDescription,
    superClass: Class<*>,
): List<FuzzedType> {
    return try {
        check(superClass.isAssignableFrom(type.classId.jClass)) { "$superClass isn't super class of $type" }
        @Suppress("UNCHECKED_CAST")
        toFuzzerType(
            type.jType.typeToken.getSupertype(superClass as Class<in Any>).type,
            description.typeCache
        ).generics
    } catch (e: Throwable) {
        logger.error("Failed to resolve $superClass generics for $type, using unresolved type parameters")
        superClass.typeParameters.map { toFuzzerType(it, description.typeCache) }
    }
}
```
And use it like this:
```kotlin
val (keyGeneric, valueGeneric) = resolveGenericsOfSuperClass(type, description, Map::class.java)
```
But your approach should also work.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
```diff
@@ -171,6 +202,20 @@ abstract class CollectionValueProvider(
         }
     }
 
+    /**
+     * Can be used to resolve some types using [type] and some method of this type
+     */
+    protected fun resolveTypeByMethod(description: FuzzedDescription, type: FuzzedType, method: Method): FuzzedType? {
```
IlyaMuravjov, line 208: In your PR this method is only used for `Map`s and `Collection`s, but we also have an `IteratorValueProvider`, that also needs to resolve generics.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
Markoutte: Approved that

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
Markoutte: Thank you, this is much better. Fixed.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
Markoutte: It is not really necessary, because `IteratorValueProvider` accepts only Iterator class itself, so, it has no modifications that can cause a problem. But I added the same call to unify these providers.

utbot-java-fuzzing/src/main/kotlin/org/utbot/fuzzing/providers/Collections.kt
```diff
@@ -306,4 +265,21 @@ class IteratorValueProvider(val idGenerator: IdGenerator<Int>) : JavaValueProvid
...
+): List<FuzzedType> {
+    val superClass = T::class.java
+    return try {
+        check(superClass.isAssignableFrom(type.classId.jClass)) { "$superClass isn't super class of $type" }
+        @Suppress("UNCHECKED_CAST")
+        toFuzzerType(
+            TypeToken.of(type.jType).getSupertype(superClass as Class<in Any>).type,
+            description.typeCache
+        ).generics
+    } catch (e: Throwable) {
```
IlyaMuravjov, line 282: I think `e` should be logged here, since it should never be thrown if everything goes right. Also, without logging `check(superClass.isAssignableFrom(type.classId.jClass))` is pretty much useless.

IlyaMuravjov (approved):
Mostly LGTM

Issue #2639, opened 2023-10-03
Incorrect assertNull is applied to the value returned from a mocked method


alisevych:
**Description**

assertNull is applied to the method returned value, while it's method is mocked.
assertEquals to exact value is expected.

**To Reproduce**

1. Install **-IU-**  [UnitTestBot plugin ](https://github.com/UnitTestBot/UTBotJava/actions/runs/6379925168) in IntelliJ IDEA
2. Open `spring-petclinic`
3. Set Fuzzing 100% mode
4. Generate `Unit tests` for `PetTypeFormatter` with `PetClinicApplication` configuration
5. Run the tests

**Expected behavior**

Method `PetType$getName()` is mocked with exact return value.
`assertEquals` for the `actualName` is expected.

**Actual behavior**

There is failing `assertNull` for `actualName` in successful test.

**Screenshots, logs**

~~~java
	@Test
	@DisplayName("parse: text = 'XZ', locale = Locale(String, String, String)")
	public void testParseWithNonEmptyString() throws ParseException {
		ArrayList arrayList = new ArrayList();
		PetType petTypeMock = mock(PetType.class);
		(when(petTypeMock.getName())).thenReturn("XZ");
		arrayList.add(petTypeMock);
		(when(ownerRepositoryMock.findPetTypes())).thenReturn(arrayList);
		Locale locale = new Locale("\n\t\r", "#$\\\"'", "#$\\\"'");

		PetType actual = petTypeFormatter.parse("XZ", locale);

		String actualName = actual.getName();
		assertNull(actualName);

		Integer actualId = actual.getId();
		assertNull(actualId);
	}
~~~

![image](https://github.com/UnitTestBot/UTBotJava/assets/37301492/13455697-8764-44d6-af41-65d40c49666f)

**Environment**

IntelliJ IDEA version - Community 2023.2
Project - Maven
JDK - 17

PR #2641, opened 2023-10-03
Use assertSame instead of deepEquals if actual is a mock

EgorkaKulikov:
## Description

Fixes # ([2639](https://github.com/UnitTestBot/UTBotJava/issues/2639))

## How to test

Automated tests: utbot-samples pipeline as a set of regression checks
Manual tests: scenario from issue

## Self-check list

### DIFF
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/services/framework/TestFrameworkManager.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/services/framework/TestFrameworkManager.kt
@@ -44,6 +44,8 @@ abstract class TestFrameworkManager(val context: CgContext)
 
     val assertions = context.testFramework.assertionsClass
 
+    val assertSame = context.testFramework.assertSame
+
     val assertEquals = context.testFramework.assertEquals
     val assertFloatEquals = context.testFramework.assertFloatEquals
     val assertDoubleEquals = context.testFramework.assertDoubleEquals
@@ -84,6 +86,10 @@ abstract class TestFrameworkManager(val context: CgContext)
 
     open fun assertEquals(expected: CgValue, actual: CgValue) {
         +assertions[assertEquals](expected, actual)
+    }
+
+    open fun assertSame(expected: CgValue, actual: CgValue) {
+        +assertions[assertSame](expected, actual)
     }
 
     open fun assertFloatEquals(expected: CgExpression, actual: CgExpression, delta: Any) {
--- a/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
+++ b/utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
@@ -1226,7 +1226,12 @@ open class CgMethodConstructor(val context: CgContext) : CgContextOwner by conte
         if (!successfullyConstructedCustomAssert) {
             val expectedVariable = variableConstructor.getOrCreateVariable(expected, expectedVariableName)
             if (emptyLineIfNeeded) emptyLineIfNeeded()
-            assertEquality(expectedVariable, actual)
+
+            if (expected.isMockModel()) {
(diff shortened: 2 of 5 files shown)


utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
```diff
@@ -1302,6 +1303,13 @@ open class CgMethodConstructor(val context: CgContext) : CgContextOwner by conte
                     }
                 }
                 else -> {
+                    currentExecution!!.result.onSuccess { resultModel ->
+                        if (resultModel.isMockModel()) {
```
IlyaMuravjov, line 1307: I think here the condition should be `expected.isMockModel()`, since mocks can be found in the result at any depth.

Here's an example of a method such that generating tests for it will result in non top level mocks in the result:
```java
public static List<Object> methodUnderTest(InterfaceWithoutImplementation interfaceWithoutImplementation) {
    if (interfaceWithoutImplementation == null) throw new NullPointerException("interfaceWithoutImplementation");
    return Collections.singletonList(interfaceWithoutImplementation);
}
```

utbot-js/src/main/kotlin/framework/codegen/model/constructor/tree/JsTestFrameworkManager.kt
```diff
@@ -51,6 +51,10 @@ class MochaManager(context: CgContext) : TestFrameworkManager(context) {
         +assertions[jsAssertEquals](expected, actual)
     }
 
+    override fun assertSame(expected: CgValue, actual: CgValue) {
+        error("assertSame does not exist in Mocha")
```
IlyaMuravjov, line 55: I'd generate an `assertEquals` and log a warning in such case.

EgorkaKulikov: I think this solution will hide a real problem of using a wrong assert not relevant for this framework.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/tree/CgMethodConstructor.kt
```diff
@@ -1226,7 +1226,12 @@ open class CgMethodConstructor(val context: CgContext) : CgContextOwner by conte
         if (!successfullyConstructedCustomAssert) {
             val expectedVariable = variableConstructor.getOrCreateVariable(expected, expectedVariableName)
             if (emptyLineIfNeeded) emptyLineIfNeeded()
-            assertEquality(expectedVariable, actual)
+
+            if (expected.isMockModel()) {
```
IlyaMuravjov, line 1230: Can you check that `assertSame` is properly generated when result contains non top level mocks (see example from earlier comment)?

On the first glance, it seems that this check will work only for top level mocks, since this method doesn't accept `depth` and hence should not be recursive.

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/services/framework/TestFrameworkManager.kt
```diff
@@ -86,6 +88,10 @@ abstract class TestFrameworkManager(val context: CgContext)
         +assertions[assertEquals](expected, actual)
     }
 
+    open fun assertSame(expected: CgValue, actual: CgValue) {
```
IlyaMuravjov, line 91: This new method is never used

IlyaMuravjov (approved):
Mostly LGTM

utbot-framework/src/main/kotlin/org/utbot/framework/codegen/services/framework/TestFrameworkManager.kt
EgorkaKulikov: Yes, but may be useful in future, even though I have already  written it.

Damtev (approved):
LGTM

PR #2642, opened 2023-10-04
Avoid spying collection-like types with @Spy if it may mislead @InjectMocks behaviour

tepa46:
## Description

If we have models of different types implementing Collection or Map, we should not construct fields of these models with `@Spy` annotation but just use `setField` for it. Otherwise it will cause problems with `@InjectMocks` in Mockito because injection by type is used and in generated tests both `List<A>` and `List<B>` are represented as `List listSpy`.

One of the possible scenarios to reproduce the bug we fix here is to generate tests for `Owner` class in `spring-petclinic` project with 100% fuzzing. Before this fix, several different spies were created but injected not as we expect after that.

## Self-check list

### DIFF
--- a/utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
+++ b/utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
@@ -1,5 +1,6 @@
 package org.utbot.framework.codegen.tree.fieldmanager
 
+import org.utbot.framework.codegen.domain.UtModelWrapper
 import org.utbot.framework.codegen.domain.builtin.spyClassId
 import org.utbot.framework.codegen.domain.context.CgContext
 import org.utbot.framework.codegen.domain.models.CgFieldDeclaration
@@ -12,6 +13,7 @@ import org.utbot.framework.plugin.api.UtAssembleModel
 import org.utbot.framework.plugin.api.UtModel
 import org.utbot.framework.plugin.api.canBeSpied
 import org.utbot.framework.plugin.api.isMockModel
+import org.utbot.framework.plugin.api.util.jClass
 
 class CgSpiedFieldsManager(context: CgContext) : CgAbstractClassFieldManager(context) {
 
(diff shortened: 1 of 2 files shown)


utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
```diff
@@ -38,9 +40,45 @@ class CgSpiedFieldsManager(context: CgContext) : CgAbstractClassFieldManager(con
...
+        val suitableDependentSpyModels = getSuitableDependentSpyModels(dependentSpyModels)
+        return constructFieldsWithAnnotation(suitableDependentSpyModels)
     }
 
+    /*
+     * If we have models of different types implementing Collection,
+     * we should not construct fields of these models with @Spy annotation
+     * because in this case, Spring cannot inject fields.
+     *
+     * The situation is similar with Map.
```
IlyaMuravjov, line 52: This comment doesn't make the situation much clear to me, can you add an example?

utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
```diff
@@ -38,9 +40,45 @@ class CgSpiedFieldsManager(context: CgContext) : CgAbstractClassFieldManager(con
...
 
-        return constructFieldsWithAnnotation(dependentSpyModels)
+        val suitableDependentSpyModels = getSuitableDependentSpyModels(dependentSpyModels)
+        return constructFieldsWithAnnotation(suitableDependentSpyModels)
     }
 
+    /*
+     * If we have models of different types implementing Collection,
+     * we should not construct fields of these models with @Spy annotation
+     * because in this case, Spring cannot inject fields.
```
IlyaMuravjov, line 50: It's not Spring, but Mockito that injects fields when it comes to `@InjectMock`. Also, it doesn't fail to inject fields, in fact it seems that it injects both `@Spy LinkedList linkedList` and `@Spy ArrayList arrayList` with the latter replacing the former causing tests that expect `linkedList` to fail.

utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
```diff
@@ -38,9 +40,45 @@ class CgSpiedFieldsManager(context: CgContext) : CgAbstractClassFieldManager(con
...
+     *
+     * The situation is similar with Map.
+     */
+    private fun getSuitableDependentSpyModels(dependentSpyModels: MutableSet<UtModelWrapper>): Set<UtModelWrapper> =
+        getSuitableDependentSpyModelsImplementing(Collection::class.java, dependentSpyModels) +
+                getSuitableDependentSpyModelsImplementing(Map::class.java, dependentSpyModels)
```
IlyaMuravjov, line 56: Somewhere else in Spy-related context you already have set of `Collection::class.java` and Map::class.java` used, can you extract this set into some constant so it's clear that same set is used in both places.

utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
```diff
@@ -38,9 +40,45 @@ class CgSpiedFieldsManager(context: CgContext) : CgAbstractClassFieldManager(con
...
+
+
+    private fun getSuitableDependentSpyModelsImplementing(clazz: Class<*>, dependentSpyModels: MutableSet<UtModelWrapper>): Set<UtModelWrapper> {
+        return when{
+            isSuitableSpyModelsImplementing(clazz, dependentSpyModels) -> dependentSpyModels.filter { clazz.isAssignableFrom(it.model.classId.jClass) }.toSet()
```
IlyaMuravjov, line 61: You can apply filtering `dependentSpyModels.filter { clazz.isAssignableFrom(it.model.classId.jClass) }` before calling the `isSuitableSpyModelsImplementing()` method to avoid implementing filtering logic twice. Furthermore, you can simplify `isSuitableSpyModelsImplementing` method to just `filteredSpyModels.map { it.classId }.toSet().size == 1` and just inline it here.

utbot-spring-framework/src/main/kotlin/org/utbot/framework/codegen/tree/fieldmanager/CgSpiedFieldsManager.kt
```diff
@@ -57,4 +60,29 @@ class CgSpiedFieldsManager(context: CgContext) : CgAbstractClassFieldManager(con
         classId.canBeInjectedByTypeInto(classUnderTest)
 
     override fun constructBaseVarName(model: UtModel): String = super.constructBaseVarName(model) + "Spy"
+
+    private fun getSuitableSpyModels(potentialSpyModels: MutableSet<UtModelWrapper>): Set<UtModelWrapper> =
+        spiedTypes.fold(setOf()) { spyModels, type ->
+            spyModels + getSuitableSpyModelsOfType(type, potentialSpyModels)
+        }
```
IlyaMuravjov, line 67: This `fold` looks like manually implemented `flatMap`.

IlyaMuravjov (approved):
Mostly LGTM

EgorkaKulikov (approved):

PR #2690, opened 2023-11-16
SBFT'24 An attempt to support UTestGlobalMock in converters

EgorkaKulikov:


utbot-junit-contest/src/main/kotlin/org/utbot/contest/usvm/UTestInst2UtModelConverter.kt
```diff
@@ -71,6 +75,8 @@ class UTestInst2UtModelConverter(
         return alreadyCreatedModel
     }
 
+    fun findInstrumentations(): List<UtInstrumentation> = instrumentations
```
IlyaMuravjov, line 78: That introduces yet another method call order dependency: `findInstrumentations()` should only be called after `processUTest`, how about we avoid that and make `UTestInst2UtModelConverter` accept `UTest` as a constructor parameter to be more consistent with `JcToUtExecutionConverter`.

IlyaMuravjov (approved):

PR #2627, opened 2023-09-29
Limit number of unknown coverage executions

IlyaMuravjov:
## Description

Make it so minimizer limits number of executions with unknown coverage and identical result type, result model type, and exception type.

Previously 50 almost identical tests were generated whenever we fail to collect the coverage (see #2622 and #2625)

### Manual tests

In scenarios described #2622 and #2625 there shouldn't be 50 identical tests with unknown coverage.

## Self-check list

- [x] I've set the proper **labels** for my PR (at least, for category and component).
- [x] PR **title** and **description** are clear and intelligible.
- [x] I've added enough **comments** to my code, particularly in hard-to-understand areas.
- [ ] The functionality I've repaired, changed or added is covered with **automated tests**.
- [x] **Manual tests** have been provided optionally.
- [x] The **documentation** for the functionality I've been working on is up-to-date.

PR #2700, opened 2023-11-27
Use symbolic result instead of concrete if it is better

EgorkaKulikov:


utbot-junit-contest/src/main/kotlin/org/utbot/contest/usvm/converter/JcToUtExecutionConverter.kt
```diff
@@ -206,12 +244,16 @@ class JcToUtExecutionConverter(
...
+    private fun getTrace(executionResult: UTestResultWrapper): List<JcInst>? = when (executionResult) {
+        is UTestConcreteExecutionResult -> when (val res = executionResult.uTestExecutionResult) {
+            is UTestExecutionExceptionResult -> res.trace
+            is UTestExecutionInitFailedResult -> res.trace
+            is UTestExecutionSuccessResult -> res.trace
+            is UTestExecutionFailedResult -> emptyList()
+            is UTestExecutionTimedOutResult -> emptyList()
+        }
+        is UTestSymbolicExceptionResult -> emptyList()
+        is UTestSymbolicSuccessResult -> emptyList()
```
IlyaMuravjov, line 256: Executions with empty coverage are not minimized. It's somewhat ok for timeout executions, since there can only be so many of them (due to time it takes register timeout). If it's too difficult to extract any coverage info from the machine, consider merging #2627 into [usvm_competitions_2024](https://github.com/UnitTestBot/UTBotJava/tree/usvm_competitions_2024) with `UtSettings.maxUnknownCoverageExecutionsPerMethodPerResultType` set to something like `10`.

utbot-junit-contest/src/main/kotlin/org/utbot/contest/usvm/converter/JcToUtExecutionConverter.kt
```diff
@@ -65,56 +68,91 @@ class JcToUtExecutionConverter(
...
-            }
+                val resultUtModel = instToModelConverter.findModelByInst(resultWrapper.result)
 
-            is UTestExecutionTimedOutResult -> {
-                logger.warn { "Timeout on ${jcExecution.method.method}" }
                 UtUsvmExecution(
                     stateBefore = constructStateBeforeFromUTest(),
                     stateAfter = MissingState,
-                    result = UtTimeoutException(TimeoutException("Concrete execution timed out")),
+                    result = UtExecutionSuccess(resultUtModel),
```
IlyaMuravjov, line 102: I am not sure if codegen can handle missing state after with `UtExecutionSuccess`. Can you check that?

EgorkaKulikov: Seems yes to me. `UtSymbolicExecution.hasThisInstance` will just return false. Methods `analyzeThisInstance`, `analyzeParameters`, `analyzeStatics` can process it. We will just not have any field state assertions. Or do I miss something?

utbot-junit-contest/src/main/kotlin/org/utbot/contest/usvm/converter/JcToUtExecutionConverter.kt
EgorkaKulikov: Merged a copy of that PR into competitions, added a comment here.

